PR #10, opened 2018-06-28
Documentation cleanup.

grigi:
Changed to Cloud Sphinx theme.

Added tortoise drawing as logo from http://www.freestockphotos.biz/stockphoto/16830 (public domain) as interm logo. It has a bit of an O'Reily feel to it. Please feed back re logo.

Split index into advertisement and a 'Getting Started' page.
Separate root ToC page to keep index cleaner.
Changed the syntax highlighting to use the `python3` variant, which also treats `await` and `async` as keywords.
Updated docs to pull version number the same as `setup.py`
Cleaned up `setup.py` to be a bit simpler.
Applied same cleanups to README.rst

Test out by doing a `make docs` and viewing them locally.

**Things I didn't get to**
Make docs use autodoc to generate more documentation. → I will need lots of time to get this up and running, and I should get more familiar with the codebase before I can do this well.


grigi:
@Zeliboba5 Sorry, I failed to commit a template file. Please try again.

grigi:
There are some theme options in Cloud: red/blue/green
redcloud
bluecloud
greencloud

What is your preference?

Zeliboba5:
I think green feels the best for the name tortoise :) 

I also noticed that there is bug in markup in fields section (http://tortoise-orm.readthedocs.io/en/latest/models_and_fields.html#fields) - options are not in list and just thrown in the heap
image
Could you fix it within this PR?

PR #12, opened 2018-06-28
PyLint plugin

grigi:
My first PyLint plugin, it was easier than I expected (but also quite fiddly, I used `pylint-django` as a reference, as the `Astroid` docs was sorely lacking)

The `tortoise.pylint` module now contains two AST transforms for:
* Models → `_meta`, `id` and `related_name` of relationships on remote model.
* Fields on Models → Transforms the type to ALSO be the instantiated python typ.

This fixes nearly every case of PyLint not finding an attribute, or the attribute not following through.
+
Made `Aggregate.aggregate_function` point to `pypika.terms.AggregateFunction` instead of None, as the astract class will never be called, and this resolves issues on Aggregates.
A few other small/safe changes. (not error related)

I re-enabled `pylint -E` as part of `make check` as all "errors" are now fixed.

I also added the PyLint plugin and usage of it to the docs.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -15,14 +15,14 @@ ignore-patterns=
 
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
-#init-hook=
+init-hook='import sys; sys.path.append(".")'
 
 # Use multiple processes to speed up Pylint.
 jobs=0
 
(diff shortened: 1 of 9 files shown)


Zeliboba5:
👍

Issue #11, opened 2018-06-28
Nested support for values and values_list


Zeliboba5:
`values` and `values_list` should also support getting fields from related models. Important feature for `values` would be overriding name of field in resulting dict.
Like this:

```
await Event.all().values_list('id', 'name', 'tournament__name')
await Event.all().values(
    'id', 
    'name', 
    tournament='tournament__name'
)
```

grigi:
Same here. Close this please or give me permission to close this?

PR #15, opened 2018-06-29
Nested support for values and values_list

Zeliboba5:
Resolving #11
I will leave it here for some time in case of someone wanting to make a review and comments, than I will merge it and distribute as 0.9.0

### DIFF
--- a/examples/aggregation.py
+++ b/examples/aggregation.py
@@ -21,7 +21,9 @@ class Event(Model):
     name = fields.TextField()
     tournament = fields.ForeignKeyField('models.Tournament', related_name='events')
     participants = fields.ManyToManyField(
-        'models.Team', related_name='events', through='event_team'
+        'models.Team',
+        related_name='events',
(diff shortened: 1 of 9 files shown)


examples/aggregation.py
```diff
@@ -57,20 +59,23 @@ def __str__(self):
...
-                                                          ).order_by('events_count')
+    ordered_tournaments = await Tournament.all().annotate(
+        events_count=Count('events'),
+    ).order_by('events_count')
     assert len(ordered_tournaments) == 2 and ordered_tournaments[1].id == tournament.id
-    event_with_annotation = await Event.all().annotate(tournament_test_id=Sum('tournament__id')
-                                                       ).first()
+    event_with_annotation = await Event.all().annotate(
+        tournament_test_id=Sum('tournament__id'),
+    ).first()
```
grigi: Yeah, this was some of the formatting that yapf just didn't want to get right for me.

PR #16, opened 2018-06-30
Started on tests

grigi:
I set up a test runner, and wrote some inital tests for field errors.
Also reports test coverage. (at 82% combined at this stage)
I also refactored some of the asserts to `raise ConfigurationError`.

Tox is working locally, Travis has been tested on my side.
(Can you please enable travis for your root branch so we can see the checks here)
(Also, can we use coveralls or coverage.io?)

I have done a very basic test isolation that creates a unique DB instance for each test. This works on both sqlite and postgres, but is not performant, and likely a surprise for anyone writing tests using this testrunner. (That is what the `{}` is for, a random value) Reasons for doing it like this is:
* Simpler to implement
* Testrunner wants to run tests in parallel
* This behaviour is OK for now, as we don't have clients using our test utils yet.
We should consider doing what Django does, where it as a `TestCase` and `TransactionTestCase` where we can get away with partial rollback or full rollback.

Another restriction right now, is because `tortoise.Tortoise` is a singleton, all tests HAS to have one set of models to test against. This makes testing how the models get built quite difficult.

Items:
* [x] Set up test runner
* [x] Record coverage
* [x] Tests for some stateless code (Field config errors)
* [x] get it to work in tox locally
* [x] set up local travis for now
* [x] use your examples as reference for stateful tests
* [x] Test isloation for SQLite
* [x] Test isolation for PostgreSQL
* [x] Rewrote tests to use unittest assert functions for better error reporting
* [x] Updated docs about test runner

Items I need your assistance on:
* [x] Many, many failing PostgreSQL tests
* [x] Can you please enable travis for your root branch so we can see the checks here
* [x] Can you please set up coveralls.io or codecov.io?

Items outside of the scope of this PR:
* Handling skipping of tests for db-specific features
* Implementing PostgreSQL specific JSONField tests
* Trying to get 100% coverage for anything


grigi:
@Zeliboba5 I have reservations on keeping JSONDict exclusive to PostgreSQL like what Django is doing. I often opt to give up the HSTORE features, and use the `django_jsondict` package instead, as more often than not, I only need to use it as non-indexable storage.

Could we promote JSONField, to give us limited fallback functionality on other databases?
I'm not against DB-specific fields, just against no JSONField being implemented for anything else.

Zeliboba5:
Are you talking about this [package](https://github.com/dmkoch/django-jsonfield)? I think it is good thing to have generic json field, especially considering the fact, that current JSONField doesn't have any special functionality and can be used only to store json.
I created #18 and will look into it soon

grigi:
I'm having what appears to me to be wrong behavior using postgresql in the tests:
https://travis-ci.com/grigi/tortoise-orm/builds/77902362

an example is `TORTOISE_TEST_DB="postgres://postgres:@127.0.0.1:5432/test_{}" green tortoise.tests.test_basic.TestBasic.test_basic`:
This:
```py
        self.assertEquals(
            await Tournament.all().values('id', 'name'),
            [{'id': 1, 'name': 'Updated name'}, {'id': 2, 'name': 'Test 2'}]
        )
```
Gives this difference:
```diff
- [<Record id=1 name='Updated name'>, <Record id=2 name='Test 2'>]
+ [{'id': 1, 'name': 'Updated name'}, {'id': 2, 'name': 'Test 2'}]
```
Which looks wrong to me, as I asked for a `.values()`, but I get objects?

@Zeliboba5 Any ideas? `.values_list()` seems to work.

There is another related issue, that a simple `.values()` doesn't default to `.values(<every possible column>)`

tortoise/backends/asyncpg/client.py
```diff
@@ -39,6 +39,26 @@ def __init__(self, user, password, database, host, port, *args, **kwargs):
...
+            try:
+                await self.execute_script('DROP DATABASE {}'.format(self.database))
+            except asyncpg.InvalidCatalogNameError:
+                pass
+            await self.execute_script(
+                'CREATE DATABASE {} OWNER {}'.format(self.database, self.user)
+            )
+            await self._connection.close()
+            self.single_connection = single_connection
+
```
Zeliboba5: I really don't like idea of possible wiping out current db just by passing a parameter to connection method, it looks really unsafe to me. I think creating schema must be done in some other place, may be you can expand `schema_generator`, something like `client.schema_generator.create_db()`

tortoise/backends/base/db_url.py
```diff
@@ -0,0 +1,73 @@
...
+            'username': 'user',
+            'password': 'password',
+        },
+        'test_defaults': {
+            'single_connection': True
+        },
+    },
+    'sqlite': {
+        'client': SqliteClient,
+        'skipfirstchar': False,
```
Zeliboba5: I don't see why you would want to write this param without splitting `_` - `skip_first_char` seems to be clearer

tortoise/backends/base/db_url.py
```diff
@@ -0,0 +1,73 @@
...
+        'vars': {
+            'path': 'filename',
+        },
+    },
+}
+
+
+def expand_db_url(dburl: str, testing: bool = False) -> dict:
```
Zeliboba5: Same here for `db_url`

tortoise/contrib/pylint/__init__.py
```diff
@@ -33,7 +34,15 @@ def transform_model(cls):
     Also keep track of relationships and make them in the related model class.
     '''
```
Zeliboba5: I missed it last time, but I think we should use double quotes in docstrings. Not like this is clear in PEP8, but even in it's [PEP](https://www.python.org/dev/peps/pep-0257/) all examples are with double quotes

tortoise/tests/test_transactions.py
```diff
@@ -0,0 +1,23 @@
...
+            async with self.db.in_transaction() as connection:
+                event = Tournament(name='Test')
+                await event.save(using_db=connection)
+                await Tournament.filter(id=event.id).using_db(
+                    connection).update(name='Updated name')
+                saved_event = await Tournament.filter(
+                    name='Updated name').using_db(connection).first()
+                self.assertEquals(saved_event.id, event.id)
+                await connection.execute_query('SELECT * FROM non_existent_table')
+        except sqlite3.OperationalError:
```
Zeliboba5: This will work only on sqlite runner, need to go with more broad exception, so this test will work with postgres

tortoise/tests/test_relations.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertEquals(participants[0].events[0], event)
+
+        await Team.fetch_for_list(participants, 'events')
+
+        await Team.filter(events__tournament__id=tournament.id)
+
+        await Event.filter(tournament=tournament)
+
+        await Tournament.filter(events__name__in=['Test', 'Prod']
+                                ).order_by('-events__participants__name').distinct()
```
Zeliboba5: This one is tricky, even though it is valid for sqlite, it gives invalid SQL for postgres (`for SELECT DISTINCT, ORDER BY expressions must appear in select list
`) and it's perfectly fine, because query itself is quite ambiguous - by what exactly name of all participants this should be ordered?
I think right answer is to just remove this query

tortoise/tests/test_filtering.py
```diff
@@ -0,0 +1,49 @@
...
+        team_second = Team(name='Second')
+        await team_second.save()
+
+        await team_first.events.add(event_first)
+        await event_second.participants.add(team_second)
+
+        found_events = await Event.filter(
+            Q(id__in=[event_first.id, event_second.id])
+            | Q(name='3')
+        ).filter(participants__not=team_second.id).order_by('tournament__id').distinct()
```
Zeliboba5: Here is another example of postgres `for SELECT DISTINCT, ORDER BY expressions must appear in select list` - but sadly, I don't think we can do much about it here, even though postgres could guess that tournament__id is unique and it isn't ambiguous it still doesn't want to work that way.
I guess we should change ordering here to just `tournament_id`

tortoise/tests/test_filtering.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertEquals(len(found_events), 2)
+        self.assertEquals(found_events[0].id, event_first.id)
+        self.assertEquals(found_events[1].id, event_third.id)
+        await Team.filter(events__tournament_id=tournament.id).order_by('-events__name')
+        await Tournament.filter(
+            events__name__in=['1', '3'],
+        ).order_by('-events__participants__name').distinct()
```
Zeliboba5: Same here as in `test_relations`

tortoise/tests/test_basic.py
```diff
@@ -0,0 +1,19 @@
...
+        self.assertEquals(saved_event.id, event.id)
+        await Tournament(name='Test 2').save()
+        self.assertEquals(
+            await Tournament.all().values_list('id', flat=True),
+            [1, 2]
+        )
+        self.assertEquals(
+            await Tournament.all().values('id', 'name'),
+            [{'id': 1, 'name': 'Updated name'}, {'id': 2, 'name': 'Test 2'}]
+        )
```
Zeliboba5: Regarding `Record` issue - currently `values` returns records for asyncpg and I don't want to change this behaviour because it will be adding overhead to parse highly efficient asyncpg `Record` to dict just for the matter of convenience  for display. I think you could parse them into dicts here in test.

.green
```diff
@@ -0,0 +1,4 @@
+verbose       = 2
+omit-patterns = tortoise/contrib/pylint/*,tortoise/tests/*
+clear-omit    = true
+run-coverage  = true
```
Zeliboba5: Tests seems to fall because of concurrently trying to create many databases with same names and concurrently trying to destroying it by everybody.
It could be avoided by making `processes = 1` in this config, but some problems will still persist.
You should add some unique token to db name so each test works on it's own database - or you could make some kind of global creating database and then running tests in transactions, something like I am doing in my [web template](https://github.com/Zeliboba5/aiohttp-tortoise-boilerplate/blob/master/conftest.py). But I think for tortoise is better to make db for each test independently, just need to think of some good way to clean them up after failed tests

tortoise/tests/test_two_databases.py
```diff
@@ -0,0 +1,60 @@
...
+from tortoise.contrib.test import TestCase
+from tortoise.tests.testmodels import EventTwo, TeamTwo, Tournament
+from tortoise.utils import generate_schema
+
+
+class TestTwoDatabases(TestCase):
+    async def setUp(self):
+        await self._tearDownDB()
+        self.db = await self.getDB()
+        self.second_db = await self.getDB()
```
Zeliboba5: Here is problem for postgres - you asking for creating of two dbs, but you will get just one - because currently postgres doesn't randomize any parameters and so test fails on dbdelete because it is trying to delete db, while `second_db` still using it

grigi:
Ok, so I'll enable coveralls for the test runner as well.
Looking at your comments a bit later.

tortoise/backends/asyncpg/client.py
grigi: Hmm, The reason I added both create and delete parameters is for the case where you Don't want the test runner to destroy the self made db.

This is really just behaviour for the test runner, but my initial attempt to put it outside the db drivers was super super hacky.
We can at worst make it so that a DB can only be deleted if it was created by the same connection?

Please propose a better way of doing this?

tortoise/tests/test_transactions.py
grigi: Doh, obviously.
Yes, this seems straightforward to fix.

tortoise/tests/test_basic.py
grigi: I get your point.
The point of an ORM to me is to isolate much of the differences between Database implementation, and to provide convenience.
I think we need to be careful and have a good reason to change the behaviour from standard.

The test is testing what is expected, so I don't think it should change.
Possibly have a `.records()` method, that gives `Record` (or Record-like) objects in response, and call it an optimisation?

.green
grigi: No, each db it creates has a randomised name, So it should work fine.
It is an extrapolation of the `{}` in  `TORTOISE_TEST_DB=postgres://postgres:@127.0.0.1:5432/test_{}`.

I was considering making `Tortoise.init()` take a db_url but for testing it was simpler to use `await self.getDB()` which would create a new db that will auto-tear down. (which is subject to change)

tortoise/tests/test_basic.py
grigi: Ok, I read up on `Record` and I see, it is a dict-like tuple object.
So, almost like a namedtuple. It is in general more memory efficient as it is a tuple at heart.

grigi: It makes me think, for all RDBMS you know the order of the returned fields, so this is an opportunity for global optimisation, as we can get rid of dicts internally.
This could noticeably reduce memory pressure, and hence performance under load.
But I think is out of scope of this.

For now I'll make the ValuesQuery do the change itself, so the rest of asyncpg can keep on using the more optimal format.

grigi:
Thanks for travis and coveralls, they are working:
https://travis-ci.com/Zeliboba5/tortoise-orm
https://coveralls.io/github/Zeliboba5/tortoise-orm

I got all the PostgreSQL tests working, although `tortoise.tests.test_filtering.TestFiltering.test_filtering` is failing inconsistently for PostgreSQL. even if I run it on its own with nothing else going on.
So I will have to figure out what is happening there.

So the "big" changes are:
* Added `db.execute_query_dict` to enforce returning dicts. (This allows us to define execute_query as "internal" and it can do the optimal thing)
* Changed some tests that were generating invalid SQL, as per your recommendation.
* Refactored db operational errors (syntax or access) to report as `tortoise.exceptions.OperationalError`.

That first item may or may not be contentious, but I decided to do it that way as we want to focus on simplicity over performance at this stage. Whilst I was investigating how this was working for SQLite I noticed several performance concerns there too. I think we should do a full benchmark system that we can profile as a post-1.0 feature. But sooner rather than later, so that we can keep performance in check.

tortoise/backends/asyncpg/client.py
Zeliboba5: I think you can just make explicit methods for `db_create` and `db_delete` on DbClient and use them. In that case using one of this methods will be fully user own responsibility and it will be hard to misuse such methods. It also seems quite easy to implement code-wise

grigi:
@Zeliboba5 Slightly off topic:
I noticed that all the fields have the following pattern:
```py
def __init__(self, source_field=None, pk=False, *args, **kwargs):
```
The `*args` is AFTER the explicit keywords. I feel this is ambiguous. I almost think we should get rid of `*args` as in the absence of that, it will associate the positional args in order of keword args definition, which is at least clear.
Can I remove the seemingly extraenous `*args`?

Zeliboba5:
Yea, better to remove them. I can't come up with any usecase where user will intentionally will pass everything as args knowing what he is doing

Zeliboba5:
I assume we can merge it?

grigi:
I'd say do another once-over, and if you are happy, you can merge it :-)

Zeliboba5:
Look all good to me.
Thank you for great work 👍

PR #23, opened 2018-07-09
Fix pylint plugin to work on astroid 2 and 1.6

grigi:


grigi:
@Zeliboba5 Please review this.

Zeliboba5:
Seems good to me. But I think we shouldn't add pylint2 as goal for release - it just that we can't affect it anyhow by ourselfs

PR #24, opened 2018-07-09
Replaced all asserts

grigi:
I replaced all `assert` and `AssertionError` with either a `tortoise.exceptions.OperationalError` or `tortoise.exceptions.ConfigurationErrer` as I saw the best match.

I also added https://github.com/PyCQA/bandit as a static analysis tool, to try limit the stupid security issues.

I had to change the examples that now don't have asserts to print instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,12 +25,14 @@ check: deps
 	flake8 $(checkfiles)
 	mypy $(mypy_flags) $(checkfiles)
 	pylint -E $(checkfiles)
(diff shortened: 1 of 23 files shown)


grigi:
@Zeliboba5 Please review this. The coverage drop is expected, as I made many of the `assert` statements to be an `if`→`raise`, which means the branches are now just more explicit (and more representative)

Zeliboba5:
👍

PR #28, opened 2018-07-15
Feature/python3.7

grigi:
pylint2 gone stable, turn CI on to see.


grigi:
@Zeliboba5 This PR is ready
I decided to mark py37+sqlite as failable, and updated the Travis env to use Xenial as a build base.

It may actually cause lots of conflicts with #29 so if you want, I can re-base it on that instead.

grigi:
Seems the SQLite transaction fixes I did in #33 resolved the py3.7-sqlite test-case.
Once I get #33 working, I'll re-enable the py3.7-sqlite CI.

grigi:
@Zeliboba5 This PR is once again ready, this time without caveats. This PR depends on #33, so please only merge it after that, else you will merge that in by accident.

Issue #14, opened 2018-06-29
Discussion: tortoise.init() & TortoiseManager


grigi:
I'm proposing start with an init handler like so:
```py
def init(...) -> TortoiseManager
    ...

manager = tortoise.init(
    modules=['app', 'app.sub'],
    databases={
        'default': <some db handler>,
        'notes': <some db handler>
    },
    
)

# Models are discoverable (and introspectable)
manager.models →
[<app.Projects>, <app.sub.Notes>, <app.sub.Tasks>]

# Actions are discoverable
manager.actions →
{
    'tortoise': {
        'syncdb': { describe itself and parameters here },
        'makemigrations': ...,
        'migrate': ....,
        'info': ...
    },
   'app.sub': {
        'import_tasks': ...
    }
}

# Actions are executed
await manager.action('syncdb', param1=value1, ...)

# And then functionality useful for testing:
await manager.snapshot(...)
await manager.rollback(...)
await manager.teardown()
```

The `TortoiseManager` object would be a class that contains all the discovered models, and access to some management commands. (a bit like Django, but simpler at first).

Management action functions get the `TortoiseManager` object as an explicit parameter. The idea is to be able to do all introspection through a standard interface, to try and minimise incentive to access private functions.

We could do a `tortoise` shell command, which would be a wrapper around `tortoise.init` and `TortoiseManager`. We do it like that to allow any system it is embedded in to easily help manage it in their own tools.

For testing I would need the ability to tear-down, the tortoise environment, and handle both application-wide models, and test-local models. We can do something like:
```py
from tortoise.testing import TortoiseTest

class TestThis(TortoiseTest):
    <some config here>
    ...
```
Where it would automatically build a test db if models are test-local (and tear it down after the tests complete), or use the globally configured test database.
For our own testing we will probably use test-local models, and real application would probably use global models. So the use case should be optimised (ito convenience) for the global model use case, and test-local models can be a bit rough.

Automatic configuration will be quite important, so I propose a set of "sane" defaults, such as ability to use environment variables/config file to do init automatically.
For that to work we also need to have the ability to initialise databases off a single string.
eg. using URI schema like: https://github.com/kennethreitz/dj-database-url

Any comments/suggestions?

Zeliboba5:
Overall I think this is a good direction and this is where we should head. But I think it should incapsulate even more from user, leaving him just with `Tortoise.init()`, or may be even without it some day.
[...]

PR #29, opened 2018-07-17
Tortoise.init() refactoring and transactions module

Zeliboba5:
#14 
This seems to be biggest change that occurred to tortoise and first one to bring so much backward incompatible changes.

- `Tortoise.init()` is now coroutine that handles all connections and models discovery.
- New transactions that finally work without explicit connection passing and also can be used for easily changing db when working with models.

This is still work in progress, I want to make few more changes, update all documentation and may be some more tests.


examples/aggregation.py
```diff
@@ -38,12 +36,9 @@ def __str__(self):
 
 
 async def run():
-    client = SqliteClient('example_aggregation.sqlite3')
-    await client.create_connection()
-    Tortoise.init(client)
-    await generate_schema(client)
-
-    tournament = Tournament(name='New Tournament')
+    await Tortoise.init(config_file='config.json')
```
grigi: Could we please have this split up?
I mean `Tortoise.init()` should instead recieve configuration in a dict, and for cases where we want to have support of config files or enviromment variables should be done as a layer above that.

This will make it easier to wrap the ORM in another framework.

Zeliboba5: Well - it already receives config as dict as main method of init - look for postgres example. 
`config_file` is just alternative option for those who use tortoise directly

examples/postgres.py
```diff
@@ -20,41 +15,35 @@ def __str__(self):
...
+    await Tortoise.init({
+        'connections': {
+            'default': {
+                'engine': 'tortoise.backends.asyncpg',
+                'credentials': {
+                    'host': 'localhost',
+                    'port': '54325',
+                    'user': 'tortoise',
+                    'password': 'qwerty123',
+                    'database': 'test',
```
grigi: Is it possible to use the db_url stuff I did?
It is a lot more concise, and since we changed over to DB-URLS at work it really made configuration much simpler, as e.g. with Docker you can easily inject environment variables, so we use that for configuration.

examples/aggregation.py
grigi: :+1:

examples/transactions.py
```diff
@@ -26,19 +21,32 @@ def __str__(self):
...
             await Event.filter(id=event.id).using_db(connection).update(name='Updated name')
             saved_event = await Event.filter(name='Updated name').using_db(connection).first()
             await connection.execute_query('SELECT * FROM non_existent_table')
-    except sqlite3.OperationalError:
+    except OperationalError:
+        pass
+    saved_event = await Event.filter(name='Updated name').first()
+    print(saved_event)
+
+    @atomic()
```
grigi: YAY!
I had lots of reservations about how transactions worked, as they were very non-intuive.

Is it possible to do `with atomic:`?

Zeliboba5: You can do it with `async with in_transaction():` without calling it from connection. I separated methods because I am not sure it's intuitive to have function overloaded with two separate effects.

examples/postgres.py
Zeliboba5: I refactored `db_url` to `config_generator` - you can see how config generated for it in `test_db_url`

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{35,36}-{sqlite,postgres}
+envlist = py{36,37}-{sqlite,postgres}
```
grigi: You can leave it as `py{35,36,37}...` and then just set:
`skip_missing_interpreters = True`
To skip that which you don't have installed.

examples/postgres.py
grigi: I feel we should make the examples use that, as it is more concise.
Or just document both, a "shorthand" that is fine for most use cases, and an "advanced"?

tortoise/__init__.py
```diff
@@ -69,15 +45,15 @@ def _init_relations(cls):
...
+                    setattr(related_model, backward_relation_name, relation)
                     related_model._meta.filters.update(
-                        get_backward_fk_filters(backward_relation_name, fk_relation)
+                        get_backward_fk_filters(backward_relation_name, relation)
                     )
 
                     related_model._meta.backward_fk_fields.add(backward_relation_name)
                     related_model._meta.fetch_fields.add(backward_relation_name)
-                    related_model._meta.fields_map[backward_relation_name] = fk_relation
+                    related_model._meta.fields_map[backward_relation_name] = relation
```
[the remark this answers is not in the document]
Zeliboba5: Oh, it's just my bad, I merged it badly today, will revert back

grigi:
I have been playing with the test runner, there is a lot of things not quite right for the test runner.

I have added initialiser and finaliser so that `Testcase` cleans up after itself, and does the migrations exactly once per test process.
There is issues with transactions on SQLite, and also the _drop_databases() command doesn't seem to actually drop anything?

Anyways I'm just stating that I'm refactoring the test suite again.

grigi:
@Zeliboba5 I did PR #33 that should fix the test runner related issues.
Test runs are MUCH faster now, which is a bonus, and I'm much more confident with the handling of transactions.

In the process SQLite is treated more similar to PostgreSQL, but as a consequence always runs in single-connection mode. For now I think that is fine, as SQLite isn't really meant for high-concurrency work.

Next up I'll rebase #28 again to be built on #33, as it appears to fix the Py3.7 + Sqlite issues. (I suspect it was related to the SQLite driver leaking file handles.)

grigi:
I have been building tests for init, and actually now have feedback to give.
There is some inconsistencies in how it appears to work.
I'll get to it the weekend, and it may be me missing some minor point.

grigi:
@Zeliboba5 At this stage I'm happy with this PR, and I'm giving the :+1: to merge it to master.
Then we can look at #37 a bit more leisurely.

grigi:
And great working with you :bowing_man:

Zeliboba5:
@grigi Do you think we should do anything special about this PR, considering it will be backward incompatible? I don't want to publish it as 1.0 already, so may be we should just make some special mark in changelog?

grigi:
@Zeliboba5: v0.10.0 + a mention that it has breaking changes re init should be enough, I think.
Maybe a short little blurb with examples in the CHANGELOG.rst file?

Zeliboba5:
I am really glad that we managed to make this release.
@grigi @etzelwu Thanks for great work 👍

PR #30, opened 2018-07-17
Fix ReadTheDocs to build using Python3.6 and install dependancies for autodoc

grigi:
It seems that RTD uses Python2.7 by default, so autodoc failed with syntax errors.
This tells RTD to use Py3.6, and to also install our standard runtime dependencies so that we don't get import errors.

This is a best effort, so I hope it solves all the problems.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,6 @@
+build:
+  image: latest
+
+python:
+  version: 3.6
+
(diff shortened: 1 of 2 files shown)


grigi:
Hmm, it made it worse.  https://readthedocs.org/projects/tortoise-orm/builds/7500692/
It just failed, but no reason?

grigi:
Ah, seems there is a RTD issue since yesterday:
https://github.com/rtfd/readthedocs.org/issues/4380

PR #33, opened 2018-07-23
New TestCase fixes

grigi:
Fixed `test.TestCase` to isolate the DB so that tests can now be wrapped in transactions.
PostgreSQL all working.

* [x] SQLite needed to have its connection infrastructure refactored so that transactions will actually run in a transaction.
* [x] Added more tests and a special exception type for transactions.
* [x] Handle exception when nested transaction not available due to build-time option.
* [x] Removed implicit commit at excute_query, which caused some transaction weirdness in SQLite
* [x] Test runner requires an `initializer()` and `finalizer()` to set up db and tear down db cleanly.

### DIFF
--- a/tortoise/contrib/test/__init__.py
+++ b/tortoise/contrib/test/__init__.py
@@ -12,7 +12,7 @@ from tortoise.exceptions import DBConnectionError
 from tortoise.transactions import start_transaction
 
 __all__ = ('SimpleTestCase', 'IsolatedTestCase', 'TestCase', 'SkipTest', 'expectedFailure',
-           'skip', 'skipIf', 'skipUnless')
+           'skip', 'skipIf', 'skipUnless', 'initializer', 'finalizer')
 _TORTOISE_TEST_DB = _os.environ.get('TORTOISE_TEST_DB', 'sqlite:///tmp/test-{}.sqlite')
 
 expectedFailure.__doc__ = """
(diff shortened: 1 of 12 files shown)


tortoise/contrib/test/__init__.py
```diff
@@ -21,6 +21,56 @@
...
+    global _APPS
+    global _CONNECTIONS
+    _CONFIG = getDBConfig(
+        app_label='models',
+        model_modules=['tortoise.tests.testmodels'],
+    )
+
+    loop = _asyncio.get_event_loop()
+    loop.run_until_complete(_init_db(_CONFIG))
+    _APPS = Tortoise.apps.copy()
```
Zeliboba5: May be deepcopy should be used here? `apps` defined with nested structure, so deepcopy should be safer in case of some test mutating apps structure

grigi: Good point.

grigi:
@Zeliboba5 I made the changes you suggested.
Can't deepcopy ` Tortoise._connections` as it contains unpickleable objects like threads or network sockets.

Zeliboba5:
It should be okay, while connections is flat dict.
Overall great PR, test are now running blazingly fast 👍 
Lately I've been overrun with work so sorry for the slow response with PRs, but I will try to sort it out.
Thank you for your work

grigi:
:bowing_man: Thanks. All good :-)
I have some time now, but doubt I'll have much in August.
I did some performance testing on SQLite and reported here: https://github.com/jreese/aiosqlite/issues/8

PR #34, opened 2018-07-23
Default model pk: id = fields.IntField(pk=True)

grigi:
If a pk isn't defined, we create a pk called id for that model.
Also added Model method tests.

Based on #28 (which is based on #33). So please only merge this after those.

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -69,7 +69,9 @@ Further, let's take a look at created fields for models
 
     id = fields.IntField(pk=True)
 
-This code defines integer primary key for table. Sadly, currently only simple integer pk is supported.
+This code defines integer primary key for table.
+If you don't define a primary key, we will create a Integer primary key with name of ``id`` for you.
+Sadly, currently only simple integer primary key is supported.
 
 .. code-block:: python3
 
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -70,19 +70,25 @@ Further, let's take a look at created fields for models
     id = fields.IntField(pk=True)
 
 This code defines integer primary key for table.
-If you don't define a primary key, we will create a Integer primary key with name of ``id`` for you.
-Sadly, currently only simple integer primary key is supported.
+
+Currently we **only** support the primary key to be called ``id``, and to be an ``IntField``.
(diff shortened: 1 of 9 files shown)


grigi:
@Zeliboba5 Please review (commit https://github.com/Zeliboba5/tortoise-orm/pull/34/commits/e472e10a1dda9448551a527a82ce11c363ba5907 only)
Based on #28 (which is based on #33). So please only merge this after those. 

We need a better way of doing this. Any suggestions?

docs/models.rst
```diff
@@ -69,7 +69,9 @@ Further, let's take a look at created fields for models
 
     id = fields.IntField(pk=True)
 
-This code defines integer primary key for table. Sadly, currently only simple integer pk is supported.
+This code defines integer primary key for table.
+If you don't define a primary key, we will create a Integer primary key with name of ``id`` for you.
+Sadly, currently only simple integer primary key is supported.
```
Zeliboba5: May be should force using `'id` field at the moment? Current inner implementations using `id` all over the place and setting another name for pk will fall everything apart.
It made me realise why django needed separate `pk` field that stores reference to pk field

grigi: I was having similar thoughts when I did this.
The reason I forced a default `id` fields was because of all the internal code using `.id` everywhere.

grigi: Or should we copy Django, and have a `pk` field, and whatever has `pk=True` is just an alias to `pk`?
Or should we just stick to `id`?

Zeliboba5: Then we should fix this description for now, because it's confusing and gives hope that you can use different name for pk. I created issue regarding this https://github.com/Zeliboba5/tortoise-orm/issues/36 and hope we will fix it before 1.0 because it's quite important in my opinion.
Could you also add some kind of validation that none other fields were defined with `pk=True` while we are forcing it ourselves

grigi:
@Zeliboba5 I updated the docs as you requested.
I also noticed that there was a few more doc changes that needed to be done. So, I did them as well.

Zeliboba5:
👍

PR #35, opened 2018-07-24
Added more Tortoise.init() tests

grigi:
Added more Tortoise.init() tests, documentation, and small refactoring in how it works.

Update 1:
* Fixed resource leak in init_tests.py

Update 2:
* Restored bad merge for fk_relation and m2m_relation in Tortoise.init()
* More typing annotations
* Tortoise.init() directly supports generate_config parameters
* Break up generate_config to use generate_db_url again
* Fix Model.get_or_create to return actual instance, not awaitable
* get_backward_fk_filters was never called
* Changed generate_config to take a dict of app: [model modules]
* Changed generate_config to call DB config "default" by default.

Update 3:
* Fix inconsistency in shorthand format


grigi:
@Zeliboba5 Please review this. It changed the `Tortoise.init()` a bit, so please comment liberally.

Zeliboba5:
Thank you! Great PR 👍

PR #37, opened 2018-07-31
MySQL support

grigi:
Based off the work of @etzelwu
Just changed to work with the new refactoring work.
And some mentions in the documentation/changelog/contrib

Thank for the great work @etzelwu :+1:


tortoise/backends/mysql/client.py
```diff
@@ -0,0 +1,205 @@
...
+    def _in_transaction(self):
+        if self.single_connection:
+            return self._transaction_class(connection=self._connection)
+        else:
+            return self._transaction_class(pool=self._db_pool)
+
+    async def execute_query(self, query):
+        # Temporarily use this method to make mysql work:
+        #  replace VARCHAR to CHAR
+        query = re.sub(r'CAST\((.*?) AS VARCHAR\)', r'CAST(\1 AS CHAR)', query)
```
Zeliboba5: What is the story here? Why it has to be casted in such way? 
Although, if it's inevitable to leave it like this - I think regex should be precompiled

tortoise/backends/mysql/client.py
```diff
@@ -0,0 +1,205 @@
...
+        try:
+            async with self.acquire_connection() as connection:
+                async with connection.cursor(aiomysql.DictCursor) as cursor:
+                    self.log.debug(query)
+
+                    await cursor.execute(query)
+                    if "SELECT" in query or "select" in query:
+                        result = await cursor.fetchall()
+                        return result
+                    return cursor.lastrowid  # return auto-generated id
```
Zeliboba5: I think this if statement should be reversed, for insert lookup, or even better return lastrowid only if some kwarg flag was passed to function itself. Doing two string lookups on each query isn't really necessary here

tortoise/backends/sqlite/client.py
```diff
@@ -64,6 +64,7 @@ def _in_transaction(self):
                     return inserted_id
                 return [dict(row) for row in results]
         except sqlite3.OperationalError as exc:
+            print("pass")
```
Zeliboba5: Should be removed?

tortoise/backends/mysql/client.py
grigi: I had the same question, then disabled it and ran the tests.
Turns out MySQL can't do case-insensitive compares on a VARCHAR, but only on a CHAR.
Which is downright bizzare.

So I left the workaround in. @etzelwu Do you have more to add?

tortoise/backends/mysql/client.py
grigi: Ah, like with SQLlite. Yes.

tortoise/backends/sqlite/client.py
grigi: I didn't even see that. yes, I think it should be removed.

tortoise/backends/mysql/client.py
Zeliboba5: I don't really like that workaround, I'll try to come up with something better to change casting on stage of making query

etzelwu:
https://github.com/Zeliboba5/tortoise-orm/blob/411066455ebc5392cae9261eb08583ea441fbc28/tortoise/models.py#L40-L67
MySQL doesn't support casting a value to VARCHAR , these functions will make MySQL syntax error. 

If we can let query cast as proper type which depends on what database is used, then we can remove that.

tortoise/backends/sqlite/client.py
etzelwu: Yes, this should be removed and be kept unchanged as before, I forgot to restore this file. Sorry.

Zeliboba5:
I pushed some changes that resolve problems with customising filters for different clients, now you only need to add something like this to executor class (adding more overrides for all failing filters):

```
from tortoise.models import insensitive_contains

def mysql_insensitive_contains(field, value):
    return functions.Upper(functions.Cast(field, SqlTypes.CHAR)).like(
        functions.Upper('%{}%'.format(value))
    )


FILTER_FUNC_OVERRIDE = {
    insensitive_contains: mysql_insensitive_contains,
}


class MySQLExecutor(BaseExecutor):
    ...

    @staticmethod
    def get_overridden_filter_func(filter_func):
        return FILTER_FUNC_OVERRIDE.get(filter_func)
```
@grigi Could you add such overrides to PR, please?

grigi:
@Zeliboba5 I implemented the changes as per your suggestion

tortoise/backends/sqlite/client.py
grigi: No worries, we are all human here :smile:

PR #43, opened 2018-08-22
Community profile

grigi:
In this PR I added a standard code of conduct, updated the contribution guide, added PR/Issue templates, and updated the README to reference these.

@Zeliboba5 Please go over the proposed changes in detail.

This is based off the recommendations at https://opensource.guide/starting-a-project/


grigi:
@Zeliboba5 Further recommendations there is to:
* move the project into its own organization
* Create a basic web-page (we could use github pages, and do a simple static one)
* Explain WHY the project (the whole `asyncio` ORM black hole, but in concise wording)
* Consider the branding. e.g. `Tortoise` does not tell me anything about ORM or `asyncio`, a literal name such as `aioorm`,`ormaio',`asyncorm` will also no work. Django is a good example of an unrelated name that succeeded. Tortoises are known to be slow, are we saying this ORM is slow? (alas, we haven't really started focusing on performance yet, still on the correctness part). In some places it is referred to as `Tortoise ORM`, is that better? At minimum we should be consistent, and have a reason for giving it a name.
* Consider the License. Why did you settle on Apache License? What other licenses did you consider? Why?

Zeliboba5:
PR looks all good to me, thank you! 👍 

Regarding questions:

1) I am planning to do it after release, I think this organisation will also be home for some extensions for tortoise (one I have already in mind is some serialization library, which could be used in frameworks, that doesn't have their own serialization engine)

2-3) Yeah, currently I don't know when I will be able to attend to that because recently I have been overrun with work, but such page is something we should have with 1.0 release

4) I think we should stick to `Tortoise ORM`, considering we have `Tortoise` singletone :) 
Regarding name - this name is mostly me being self-ironical about the fact that I disregarded speed that pure sql gave me in favour for easy extensible interface. I still think that such natural unrelated names sounds good and it much more catchy than artificial names and loads off some tension overall.

5) The licenses I checked were  GNU (and it's descendants), Apache and MIT. I didn't liked GNU because it puts limitations on users of Tortoise, regarding distribution of their products, which is something I want to avoid. Apache and MIT are probably quite same, but Apache gives some rights to project owner to defend his trademark and code, which is not something I need right now, but may be, if it turns out in something bigger, those rights could be handy

PR #44, opened 2018-08-31
Set single_connection to True by default, release v0.10.2

grigi:
Since we are both low on time thesedays, I'm pushing a change that disables connection by default pooling for now, until we get round to fixing it.
As per our Correctness focus.

I also included the tests done so far from #38 

This should be v0.10.2

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+0.10.2
+------
+- Set single_connection to True by default, as there is known issues with conection pooling
+- Updated documentation
+
 0.10.1
 ------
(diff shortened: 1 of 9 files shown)


grigi:
Ok, I had permissions to do it myself, travis is smart enough to keep history, whereas coveralls lost history. ah well.

Zeliboba5:
It's a little bit sad to merge something that downgrades system, but I think that should be right decision for now.
Thanks for PR

grigi:
Yes, I wanted to resolve it, but I just didn't have much time with starting a new job, and other family responsibilities.
At least it means that if we start announcing this now, I think we at least give the people whom test it out something we know works, we can always focus on performance after we are feature complete (which is soon)

PR #45, opened 2018-09-07
Separate transactions for separate connections

Zeliboba5:
I stumbled across flaw in design, that forced all connections to be switched to transaction, even if transaction is made on another database. This blocks development for several database connections

I changed it to map of current transactions, so each connection could simultaneously run it's own transaction

### DIFF
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -1,8 +1,9 @@
 import importlib
 import json
 import os
+from copy import deepcopy
 from inspect import isclass
-from typing import Any, Dict, List, Optional, Type  # noqa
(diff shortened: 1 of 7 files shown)


grigi:
I ported the pooled test runner changes from #41 and re-enabled thread pooling.
:disappointed: 
postgres + pool ⇒ hanging
mysql + pool ⇒ errors about IOLoop, fails & hanging

Zeliboba5:
Well, that PR meant to address only issue with transactions in several databases config, so it was predictable that it won't fix things up just like that :)

Zeliboba5:
I reverted merge, take a look at PR code, if it seems alright to you - press approve button :)

PR #47, opened 2018-09-18
Papercuts

grigi:
## Description
Fixing (and/or recording) some papercuts as I came across them:
* [x] Relative sqlite db file now works
* [x] Removed confusing error message for missing db driver dependency
* [x] Added `aiosqlite` as requirement (as every other ORM I looked at has sqlite support natively)
* [x] `execute_script()` now annotates errors just like `execute_query()`, also to reduce confusion
* [ ] There is **NO DOCUMENTATION** about transactions!!!
* [ ] Sample app does not persist changes unless inside an transaction (bug?)
* [ ] Need a run_async helper function to run async code and clean up connections (else app just hangs there)
* [ ] Document cleanup procedure and importance of it

## Motivation and Context
I was interested in determining a baseline ito performance, and did a very naïve benchmark suite (it will get better)
https://github.com/tortoise/orm-benchmarks
I used the documentation to build each of the samples, and found using Tortoise quite painful.
This PR is to address the worst silly issues.


grigi:
@Zeliboba5 This PR is ready for review. :-)

tortoise/__init__.py
```diff
@@ -345,4 +348,32 @@ def _get_config_from_config_file(cls, config_file):
...
+
+            ...
+
+        run_async(do_stuff())
+    """
+    loop = asyncio.get_event_loop()
+    try:
+        loop.run_until_complete(coro)
+    except Exception as exc:
+        raise
```
Zeliboba5: Why is it needed to catch exception just to reraise it?

grigi: can you do a try → finally without an except in Python?

grigi: Answer is "yes". ok, I'll fix that,

PR #53, opened 2018-09-28
Changed DateTimeField auto_now to generate value on to_db_value

grigi:
I extended to_fb_value to have the capability of doing custom logic like auto_now and auto_now_add from DateTimeField.

No more special case in insert/update executors.

### DIFF
--- a/tortoise/backends/asyncpg/executor.py
+++ b/tortoise/backends/asyncpg/executor.py
@@ -6,11 +6,10 @@ from tortoise.backends.base.executor import BaseExecutor
 class AsyncpgExecutor(BaseExecutor):
     async def execute_insert(self, instance):
         self.connection = await self.db.get_single_connection()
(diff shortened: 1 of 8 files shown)


Zeliboba5:
I am not quite getting it, where calling `default` takes place on insert?

Zeliboba5:
The problem is that with such flow it creates value for date field in the moment of instancing object, not on the db write which is why I implemented it in that way.

I think creating datetime on model object instancing is misleading and is not desired behaviour.

Also, as I see right now - there is flaw with current insert, cause it doesn't create value for `auto_now` on insert at all and just skips it.

grigi:
@Zeliboba5 I rewrote it to solve it in a more flexible, whilst still correct way. it will now handle `auto_now` just in the DateTimeField as well.
Makes it easier to do Field-custom logic. or fields that track something else in the model.
e.g. an auto-slugger, etc...

PR #54, opened 2018-10-02
Insert cleanup

grigi:
This branch started with the goal of using prepared statements for inserts, but ended up being a general purpose insert/Client related cleanup.

Work done:
* [x] `_prepare_insert_columns` and `_prepare_insert_values` now do what they are called
* [x] Added type annotations to the DB Client classes (It is partial, due to some very confusing type usage, especially for "SingleConnectionsWrapper" and "TransactionWrapper" classes)
* [x] Split out specialised `execute_insert()` method
* [x] Decorators to translate exceptions, so we repeat less code.
* [x] Use macro executors from https://github.com/jreese/aiosqlite/pull/13 for less synchronization overhead.
* [x] Generate prepared statements, and use `values` as a DB driver parameter (https://github.com/kayak/pypika/issues/163 is stalled, do something slightly hacky in the interm)(We need to try and get this into pypika, as `asyncpg` needs prepared statements to perform well)
* [x] Simple, but safe query cache for inserts

Not related to inserts:
* [x] Pre-generate initial pypika query object per model, instead of in queryset instantiation, as it is immutable (gives 25-50% speedup for small fetch operations)

This should be the last performance-oriented PR from me for now, performance is up over double (and much more for sqlite)

### DIFF
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -1,4 +1,4 @@
-from typing import Callable, Dict, Type  # noqa
+from typing import Callable, Dict, List, Type  # noqa
 
 from pypika import Table
 
@@ -8,6 +8,7 @@ from tortoise.exceptions import OperationalError
 
 class BaseExecutor:
     TO_DB_OVERRIDE = {}  # type: Dict[Type[fields.Field], Callable]
+    FILTER_FUNC_OVERRIDE = {}  # type: Dict[Callable, Callable]
 
     def __init__(self, model, db=None, prefetch_map=None, prefetch_queries=None):
         self.model = model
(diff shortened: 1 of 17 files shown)


tortoise/backends/base/executor.py
```diff
@@ -37,28 +40,43 @@ def _prepare_insert_columns(self):
...
-    async def execute_insert(self, instance):
+    def _prepare_insert_statement(self, columns: List[str]) -> str:
         # Insert should implement returning new id to saved object
         # Each db has it's own methods for it, so each implementation should
         # go to descendant executors
         raise NotImplementedError()  # pragma: nocoverage
 
+    async def execute_insert(self, instance):
+        self.connection = await self.db.get_single_connection()
+        key = '%s:%s' % (self.db.connection_name, self.model._meta.table)
```
Zeliboba5: Does this kind of formatting gives any significant speedup?
If not, I think following `.format()` would be better, cause it feels really much more pythonic way

grigi: Hmm, no idea.
I tend to use the % formatting over the .format() out of habit.
I just did a quick search, and, yep. that is the only such code in the codebase. I'll change it.

grigi:
Yes, people get excited about performance :grin: 
I have a pypi account. I'll forward you my credentials when I find them. Thanks for the offer :smile:

PR #56, opened 2018-10-13
Remove known broken code & cleanup

grigi:
Since pooled connections are known broken, This PR removes it, and cleans up the code, so that it will be easier to re-add it plainly.

### DIFF
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -163,7 +163,7 @@ class Tortoise:
             connection = client_class(**db_params)
             if create_db:
                 await connection.db_create()
-            await connection.create_connection()
+            await connection.create_connection(True)
             cls._connections[name] = connection
             current_transaction_map[name] = ContextVar(name, default=None)
 
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -163,7 +163,7 @@ class Tortoise:
             connection = client_class(**db_params)
             if create_db:
                 await connection.db_create()
-            await connection.create_connection(True)
+            await connection.create_connection(with_db=True)
             cls._connections[name] = connection
             current_transaction_map[name] = ContextVar(name, default=None)
 
--- a/tortoise/backends/asyncpg/client.py
+++ b/tortoise/backends/asyncpg/client.py
@@ -42,13 +42,6 @@ class AsyncpgDBClient(BaseDBAsyncClient):
         self.host = host
         self.port = int(port)  # make sure port is int type
 
-        self.dsn = self.DSN_TEMPLATE.format(
-            user=self.user,
-            password=self.password,
-            host=self.host,
-            port=self.port,
-            database=self.database
-        )
         # self._db_pool = None  # Type: Optional[asyncpg.pool.Pool]
         self._connection = None  # Type: Optional[asyncpg.Connection]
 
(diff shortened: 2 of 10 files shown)


grigi:
rebased off master

grigi:
This PR is getting a bit bit and confused. I removed a lot of code, then added a lot of code... so it just looks like churn
There is 4 things here:
1) Remove connection pools and related broken code
2) Change the config parser to be more generic and pass extra info in.
3) Add lots of type annotations
4) Add some tests
I feel I should break this in at least two PRs.

grigi:
@Zeliboba5 Please review.

tortoise/__init__.py
```diff
@@ -163,7 +163,7 @@ def _discover_models(cls, models_path, app_label) -> List[Type[Model]]:
             connection = client_class(**db_params)
             if create_db:
                 await connection.db_create()
-            await connection.create_connection()
+            await connection.create_connection(True)
```
Zeliboba5: I think it's better to pass it as a kwarg, because right now it's not obvious what this `True` does and it forces reader to jump into function to see why we are passing `True` here

tortoise/backends/asyncpg/client.py
```diff
@@ -42,85 +42,61 @@ def __init__(self, user: str, password: str, database: str, host: str, port: Sup
...
 
-        self.dsn = self.DSN_TEMPLATE.format(
-            user=self.user,
-            password=self.password,
-            host=self.host,
-            port=self.port,
-            database=self.database
-        )
-        self._db_pool = None  # Type: Optional[asyncpg.pool.Pool]
+        # self._db_pool = None  # Type: Optional[asyncpg.pool.Pool]
```
Zeliboba5: I think we don't need commented code in distributed lib, it would just confuse people if they see it

tortoise/backends/asyncpg/client.py
```diff
@@ -42,85 +42,61 @@ def __init__(self, user: str, password: str, database: str, host: str, port: Sup
...
-            if not self.single_connection:
-                self._db_pool = await asyncpg.create_pool(self.dsn)
-            else:
-                self._connection = await asyncpg.connect(self.dsn)
+            self._connection = await asyncpg.connect(dsn)
             self.log.debug(
-                'Created connection with params: user=%s database=%s host=%s port=%s',
-                self.user, self.database, self.host, self.port
+                'Created connection %s with params: user=%s database=%s host=%s port=%s',
+                self._connection, self.user, self.database, self.host, self.port
```
Zeliboba5: May be we should change it to `.format` here for uniformity of string formatting?

grigi: That is a logging message. For performance reasons we should use %s and commas. It then allows the log to become a no-op when you don't have debug level logging configured.
Using string interpolation in the logging functions is considered a code smell.

tortoise/__init__.py
grigi: :+1:

tortoise/backends/asyncpg/client.py
grigi: :man_facepalming:

tortoise/backends/asyncpg/client.py
Zeliboba5: Ah, yeah, wrong on my part 👍

Issue #57, opened 2018-10-19
Allow loop argument for initializer() and finalizer() function


jrmi:
While testing my code using tortoise-orm, I created a specific `event_loop` and dropped the default thread event loop. As consequence the method `get_event_loop()` is not working as excepted. 
The problem is that `initializer` and `finalizer` methods use the asyncio method to get the event loop and there is no other way.

I would like to be able to give the event loop as an optional argument with default value to what we get with the `get_event_loop()` method.

grigi:
What test runner are you using?
I have very recently revamped the test utils to mess around less with the event loop. Might be useful to take that + fixes for you together.
Could you post a small example of how you did it so I can use that as a test case?

jrmi:
I'm trying to use pytest which seems not to be the recommended tool with tortoise ;-)  but seems working. I can't use the contrib TestCase so i tried to simulate the same feature with pytest but the loop argument was missing.

Here the code to reproduce: 

```python
import asyncio
from tortoise.contrib.test import initializer

asyncio.set_event_loop(None)
loop = asyncio.new_event_loop()

initializer() # Should be initializer(loop=loop)
```

grigi:
We have a policy to try and minimise Tortoise dictate project structure, and I think dictating test runner is part of it. So happy to make the unittest runner be easy to integrate with pytest, and nose2.
(I actually had this thought to add it to a project timeline.)

So, I'll take the existing test runner refactor, make it work with a custom event loop as a first step.

Zeliboba5:
@jrmi While I hope we will make tortoise pytest-friendly soon enough - here is example of fixtures for tortoise tests, that I am using in my project, it creates db on test session start and runs each test in separate transaction (there is also pytest-aiohttp plugin used for `loop` fixture and async tests)

```
async def init_tortoise(config):
    try:
        await Tortoise.init(config)
        rules_connection = Tortoise.get_connection('default')
        await rules_connection.close()
        await rules_connection.db_delete()
    except DBConnectionError as exc:
        pass

    rules_connection = AsyncpgDBClient(connection_name='default', **settings.RULES_DB_CONFIG)
    await rules_connection.db_create()
    await Tortoise.init(config)
    await Tortoise._reset_connections()

@pytest.fixture(scope='session')
def tortoise_config(request):
    loop = get_event_loop()
    config = <your actual test config looking at separate test db that can be deleted>
    loop.run_until_complete(init_tortoise(config))
[...]

Issue #58, opened 2018-10-19
Bad error message for CharField about missing max_lenght argument


jrmi:
When `max_length` argument is missing the error message is 
```
tortoise.exceptions.ConfigurationError: max_digits must be >= 1
```
and doesn't show the true problem.

To reproduce it execute this code:

```python
from tortoise.models import Model
from tortoise import fields

class AModel(Model):
    id = fields.IntField(pk=True)
    some_attribute = fields.CharField()
```

I think the error message should be instead:
```
tortoise.exceptions.ConfigurationError: missing 'max_length' parameter
```

grigi:
Good point.
Confusing error mesages waste time.

grigi:
I fixed missing/bad parameters for `CharField`, `DatetimeField` and `DecimalField` with error messages in the form of:
```
missing 'max_length' parameter
'max_length' must be >= 1
missing 'decimal_places' parameter
'decimal_places' must be >= 0
missing 'max_digits' parameter
'max_digits' must be >= 1
You can choose only 'auto_now' or 'auto_now_add'
```
Which is currently the set of field param validators.

PR #59, opened 2018-10-22
Refactor ``Tortoise.init()`` and test runner

grigi:
* [x] Does not re-create connections per test, so now tests pass when using an SQLite in-memory database
* [x] Can pass event loop to test initializer function: `initializer(loop=loop)`
* [x] `nose2` plugin & docs
* [x] `py.test` usage docs
* [x] Fixed Test runner autodoc ignoring the test classes as it had `asynctest` blacklisted.
* [x] Minor doc cleanup

### DIFF
--- a/tortoise/contrib/test/__init__.py
+++ b/tortoise/contrib/test/__init__.py
@@ -14,8 +14,9 @@ from tortoise.exceptions import DBConnectionError
 from tortoise.transactions import current_transaction_map, start_transaction
 
 __all__ = ('SimpleTestCase', 'IsolatedTestCase', 'TestCase', 'SkipTest', 'expectedFailure',
-           'skip', 'skipIf', 'skipUnless', 'initializer', 'finalizer')
-_TORTOISE_TEST_DB = _os.environ.get('TORTOISE_TEST_DB', 'sqlite://:memory:')
+           'skip', 'skipIf', 'skipUnless', 'env_initializer', 'initializer', 'finalizer',
+           'getDBConfig')
+_TORTOISE_TEST_DB = 'sqlite://:memory:'
 
 expectedFailure.__doc__ = """
 Mark test as expecting failiure.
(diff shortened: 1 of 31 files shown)


grigi:
@Zeliboba5 Could you please review this, and test it. I would like some certainty that the testrunner changes I have here is compatible with all your test code.
It also includes changes for #57 & #60 
(Still busy with simple test plugins for ``nose`` and ``py.test``

tortoise/contrib/test/__init__.py
```diff
@@ -92,6 +107,21 @@ class SimpleTestCase(_TestCase):
...
+            loop = None
+        else:
+            loop = self.loop = asyncio.new_event_loop()
+
+        policy = _Policy(asyncio.get_event_loop_policy(),
+                         loop, self.forbid_get_event_loop)
+
+        asyncio.set_event_loop_policy(policy)
+
+        self.loop = self._patch_loop(self.loop)
```
Zeliboba5: I can't really get a hold of this function. Where does it get called? And why is has to be with with double underscore ?

grigi: I'm overriding the function in `asynctest/case.py` (that is where the name comes from).
The only change is to prefer the `_LOOP` that I saved in initalizer(), so that if a custom loop was passed in, it would be used instead of `asyncio.get_event_loop()`.

grigi:
Yeah, I felt a bit bitten by the slightly misleading naming. If you want me to, I can try and refactor to reduce changes a bit?

Zeliboba5:
No, I think it should be fine this way, separating resetting apps and closing connections is actually right decision, so let's keep it that way

tortoise/contrib/test/__init__.py
Zeliboba5: https://github.com/Martiusweb/asynctest/blob/master/asynctest/case.py Are talking about this file?
Shouldn't it then be just one underscore (like `_init_loop`, not `__init_loop`)?

grigi: :man_facepalming:
You are right.
How did that happen...

Zeliboba5:
Could you also, please, change examples to use in-memory sqlite base, so it will be easier to run examples for interested users?

grigi:
@Zeliboba5 I think this is ready for review and release.
Fixes #57 #58 #60 #63

grigi:
I'm merging this in, as already got approval on the core changes, and wanting to get a bugfix release out. I hope this is fine @Zeliboba5

Issue #62, opened 2018-10-23
BooleanField not working as expected with filter


jrmi:
When I try to filter a model by a boolean property with an sqlite DB i get this error:
```python
Traceback (most recent call last):
  File "buggy.py", line 31, in <module>
    loop.run_until_complete(init())
  File "/usr/lib/python3.6/asyncio/base_events.py", line 468, in run_until_complete
    return future.result()
  File "buggy.py", line 23, in init
    result = await Tournament.filter(bool_field=True)
  File "…/Code/tortoise-orm/tortoise/queryset.py", line 479, in _execute
    self.query, custom_fields=list(self._annotations.keys())
  File "…/Code/tortoise-orm/tortoise/backends/base/executor.py", line 24, in execute_select
    raw_results = await self.connection.execute_query(str(query))
  File "…/Code/tortoise-orm/tortoise/backends/sqlite/client.py", line 23, in wrapped
    raise OperationalError(exc)
tortoise.exceptions.OperationalError: no such column: true
```

You can reproduce this bug by executing this code:

```python
import asyncio
from tortoise import Tortoise
from tortoise.models import Model
from tortoise import fields

class Tournament(Model):
    id = fields.IntField(pk=True)
    bool_field = fields.BooleanField()

    def __str__(self):
        return self.name

async def init():
    await Tortoise.init(
        db_url='sqlite://:memory:',
        modules={'models': ['__main__']}
    )
    await Tortoise.generate_schemas()
    result = await Tournament.filter(bool_field=True)
    await Tortoise.close_connections()

loop = asyncio.get_event_loop()
loop.run_until_complete(init())
```
This is the same problem on related models filters.

I was excepting a filtered queryset that contains only Tournament with `bool_field` value to `True`.

jrmi:
As workaround I use `.filter(bool_field=1)` for those who experience this issue.

Zeliboba5:
Seems to be sqlite specific problem, cause they don't have `FALSE` and `TRUE` literals.
Can't figure out elegant way to fix it right away, may be we should implement custom sqlite query builder for pypika so it would cast all boolean values to int. 
Will have a chance to look into it further tomorrow

grigi:
I did cast the bool values on load/save to int as that is how SQLite seems to be storing them. So we did this half-way already.

PR #64, opened 2018-11-05
Make use of pypika objects being immutable, and pre-build as much of the query as possible.

grigi:
Make use of pypika objects being immutable, and pre-build as much of the query as possible.
Also pre-build DB-specific filter override list. (avoids a transaction lookup which is quite expensive)
Also optimised queryset cloning.,

This gives us a further 21-48% speed up for small queries. :grin:

This actually slowed down tests marginally, as in the tests we re-build all the query caches for each test. :man_shrugging:

### DIFF
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -219,7 +219,7 @@ class MetaInfo:
     __slots__ = ('abstract', 'table', 'app', 'fields', 'db_fields', 'm2m_fields', 'fk_fields',
                  'backward_fk_fields', 'fetch_fields', 'fields_db_projection', '_inited',
                  'fields_db_projection_reverse', 'filters', 'fields_map', 'default_connection',
-                 'basequery')
+                 'basequery', 'basequery_all_fields', '_filters')
 
     def __init__(self, meta):
         self.abstract = getattr(meta, 'abstract', False)  # type: bool
@@ -233,11 +233,13 @@ class MetaInfo:
         self.fetch_fields = set()  # type: Set[str]
         self.fields_db_projection = {}  # type: Dict[str,str]
         self.fields_db_projection_reverse = {}  # type: Dict[str,str]
+        self._filters = {}  # type: Dict[str, Dict[str, dict]]
         self.filters = {}  # type: Dict[str, Dict[str, dict]]
         self.fields_map = {}  # type: Dict[str, fields.Field]
         self._inited = False
         self.default_connection = None
         self.basequery = None
+        self.basequery_all_fields = None
 
     @property
     def db(self):
@@ -250,14 +252,18 @@ class MetaInfo:
         )
 
     def get_filter(self, key):
-        filter_info = self.filters[key]
-        overridden_operator = self.db.executor_class.get_overridden_filter_func(
-            filter_func=filter_info['operator'],
-        )
-        if overridden_operator:
-            filter_info = deepcopy(filter_info)
-            filter_info['operator'] = overridden_operator
-        return filter_info
+        return self.filters[key]
+
+    def override_filters(self):
+        get_overridden_filter_func = self.db.executor_class.get_overridden_filter_func
+        for key, filter_info in self._filters.items():
+            overridden_operator = get_overridden_filter_func(
+                filter_func=filter_info['operator'],
+            )
+            if overridden_operator:
+                filter_info = deepcopy(filter_info)
+                filter_info['operator'] = overridden_operator
+            self.filters[key] = filter_info
 
 
 class ModelMeta(type):
@@ -314,7 +320,7 @@ class ModelMeta(type):
         }
         meta.fields = set(fields_map.keys())
         meta.db_fields = set(fields_db_projection.values())
-        meta.filters = filters
+        meta._filters = filters
         meta.fk_fields = fk_fields
         meta.backward_fk_fields = set()
         meta.m2m_fields = m2m_fields
@@ -325,6 +331,7 @@ class ModelMeta(type):
             meta.abstract = True
 
         new_class = super().__new__(mcs, name, bases, attrs)
+
         return new_class
 
 
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -254,7 +254,7 @@ class MetaInfo:
     def get_filter(self, key):
         return self.filters[key]
 
-    def override_filters(self):
+    def generate_filters(self):
         get_overridden_filter_func = self.db.executor_class.get_overridden_filter_func
         for key, filter_info in self._filters.items():
             overridden_operator = get_overridden_filter_func(
@@ -267,6 +267,8 @@ class MetaInfo:
 
 
 class ModelMeta(type):
+    __slots__ = ()
+
     def __new__(mcs, name, bases, attrs, *args, **kwargs):
         fields_db_projection = {}  # type: Dict[str,str]
         fields_map = {}  # type: Dict[str, fields.Field]
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -1,4 +1,4 @@
-from copy import deepcopy
+from copy import copy
 from typing import Any, Dict, List, Optional, Set, Tuple  # noqa
 
 from pypika import JoinType, Order, Table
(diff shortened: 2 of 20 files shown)


tortoise/models.py
```diff
@@ -250,14 +252,18 @@ def db(self):
...
-        overridden_operator = self.db.executor_class.get_overridden_filter_func(
-            filter_func=filter_info['operator'],
-        )
-        if overridden_operator:
-            filter_info = deepcopy(filter_info)
-            filter_info['operator'] = overridden_operator
-        return filter_info
+        return self.filters[key]
+
+    def override_filters(self):
```
Zeliboba5: May be it should be named something like `generate_filters`? Cause in the end it's not just overrides them, but it basically generates values for `filters`  field. And `_filters` could be renamed to `_generic_filters` or `_default_filters` to point out that it is not the same field with `filters`.
It's just that current names made me confused for a while to figure out why there is two of them and how are they connected

grigi: That is a good point.

Zeliboba5:
Also, regarding README, I think we should add info about mysql and pg drivers availability, because right now that info isn't available in README.
Also we could remove defining of `id` field in examples, because at the moment it is still non-configurable, so people could just use predefined `id` field, without giving them hope that they can change it :)

grigi:
I'm in two minds about removing `id` from all examples...
Do you have another suggestion?

Zeliboba5:
May be remove them but leave a comment saying that `id` field will be automatically generated? 
Both in README and docs. 
I think that way it will be quite clear and won't cause misunderstanding for users

grigi:
So, @Zeliboba5 the new pypika doesn't fix it: https://travis-ci.com/tortoise/tortoise-orm/jobs/157473493

However, on my local system the tests passed, before and after the pypika upgrade.
So I suspect that this is fixed in a later version of SQLite...
But most people don't run the latest SQLite, so we need to support the alias of True=1 and False=0.

Zeliboba5:
Sorry if I was misleading, but new new version of PyPika meant not only just version, but also using SQLiteQuery from this commit https://github.com/kayak/pypika/commit/d14965cd8a481f9f99c82b999dbc64224b6f3682
Could you add it and retest please?

grigi:
Unforutnately the implementation is not complete:
```py
>>> from pypika import Table
>>> from pypika.dialects import SQLLiteQuery
>>> table_abc = Table('abc')
>>> SQLLiteQuery.from_('abc').select('a').where(table_abc.a == True)
SELECT "a" FROM "abc" WHERE "a"=true
```

But we have our own `to_db_value` functions, and I see it is configured for inserts, but not filters. Should we not do that instead?

grigi:
@Zeliboba5 I Fixed #62 by using the overrideable to_db functions in filtering, instead of just insert.
In the end I did not use the new SQLLiteQuery dialect, as it gained me nothing. And we now consistently handle per-DB field type overrides.

Please review.

tortoise/queryset.py
```diff
@@ -13,14 +13,15 @@
...
 
-    def __init__(self):
-        self._joined_tables = []
-        self.query = None
-        self.model = None
+    def __init__(self, model, db: Optional[BaseDBAsyncClient]) -> None:
+        self._joined_tables = []  # type: List[Table]
+        self.model = model
+        self.query = model._meta.basequery  # type: Query
+        self._db = db if db else model._meta.db  # type: BaseDBAsyncClient
```
Zeliboba5: As far as I remember I specifically removed `db` field from instancing in `__init__`, because it brings trouble, where you define your query outside of transaction and then run it inside transaction, which leads to query being executed outside of transaction, cause it's connection before transaction even started.
So that's why connection should be fetched only after query being `await`ed 

However, as I see lower, I didn't remove it in `DeleteQuery` and some other queries, which seems wrong to me now. Is there some case that I am missing?

grigi: I think that was valid before the transaction rewrite, because now the connection is actually stored in a context variable?

Also, I was toying with separating the DB instance from the actual "connection", I was trying it for #56 but then realised the test runner was causing chaos, hence I deferred that until after that was done. (Since we are an async-only ORM, why can't we always manage connections as a pool?) (Currently all that PR does is remove all the scattered connection logic, just to simplify things, so we can try and re-add it, and lazy evaluation makes debugging that super complicated)

And then there is the case of allowing a single Model that would be available on different databases (of potentially different types). I considered that, and the whole way we manage QuerySet assumes it is a 1:1 relationship, and it will require a large rewrite to make it, so decided that it is an OK limitation for now, as I never needed to do that, and it will complicate everything too.

Zeliboba5: > I think that was valid before the transaction rewrite, because now the connection is actually stored in a context variable?

Yeah, but then, after refactoring, I encountered the case I described above and fixed it. But you are actually right, that it looks scattered for now. How about making `db` property for `AwaitableQuery`, so code will still look better, but actual value will be evaluated only on call?

Regarding allowing using of single model on different databases - it's seems to me that implementation that allows it will bring much more complexity that we want to handle at the moment, though it doesn't bring benefits for most of users, cause I really think that in most cases you would want to just create two separate models to avoid confusion.

grigi: Yes, too much complexity, in both interface and implementation. I like the suggestion of two separate models in that case. Allows you to manage migrations separately.

I'll look at moving some of that logic to awaitablequery tomorrow. Just concerned with the todb overrides.
Just for clarity, the issue was specific to pooled connections?

grigi: Is it OK with you after I get this merged, then change #56 to be a PR to remove all the pooled logic that isn't working. Then a new PR to try and add it back in a controlled way. Then the main purpose of #56 would be to simplify the codebase?

Zeliboba5: Transaction issue? No, it was encountered in all cases where you define your query and run it separately. Like this:
```
query = Event.all()
async with in_transaction():
    result = await query
```

Because `Event.all()` was called before transaction started - it fetched `self.model.db` and got not `TransactionWrapper`, but `DBClient`. And then, when `query` is `await`ed - it called `.execute_query()` on DBClient, not on `TransactionWrapper`, which results in running query outside transaction, so that brings errors in most cases.

grigi: Thanks, I'll add that as a test case, and try to get it to fail.

Zeliboba5: > Is it OK with you after I get this merged, then change #56 to be a PR to remove all the pooled logic that isn't working. Then a new PR to try and add it back in a controlled way. Then the main purpose of #56 would be to simplify the codebase?

Yeah, that would be nice

grigi: I added the following tests, expecting them to fail:
```py
async def test_await_across_transaction_fail(self):
    tournament = Tournament(name='Test')
    query = tournament.save()

    try:
        async with in_transaction() as connection:
            result = await query
            raise Exception('moo')
    except Exception:
        pass

    self.assertEqual(await Tournament.all(), [])

async def test_await_across_transaction_success(self):
    tournament = Tournament(name='Test')
    query = tournament.save()

    async with in_transaction() as connection:
        result = await query

    self.assertEqual(await Tournament.all(), [tournament])
```
But they pass. Which means to me that if I prepare OUTSIDE the transaction but await INSIDE a transaction, the transaction isolation then works the same as if everything was inside a transaction.

If that is the case, then the issue you were experiencing is now not valid anymore?
Makes sense to me, because `QuerySet._execute()` only runs when you await it. so it inherits the state of where it gets awaited.

Which seems correct to me, so can we leave it as is?

Zeliboba5: Issue I described doesn't apply to `.save()`, cause when you call `.save()` without awaiting nothing occurs, it's only executed when it's awaited.

Problem should occur when you build queryset, and await it later, like that:
```
    async def test_await_across_transaction(self):
        tournament_query = Tournament.filter(name='Test')

        async with in_transaction() as connection:
            await Tournament.create(name='Test')
            result = await tournament_query
            self.assertEqual(len(result), 1)
```

But after actually writing that test I understood that you were right, that problem occurs only in connection pool context and when we deal with single connection it doesn't matter because everything after transaction start will fall into it transaction. 
Sorry for confusion. If we are cutting out pools for now we can leave `_db` as it is.

grigi: Ok, let me also do tests for `.update()` and then we see.
Only `.save()` and `.update()` modify contents, so those should be the only queries applicable to transactions?

Zeliboba5: Well, transactions are not only about writing, it's also about transaction isolation.
In test that I wrote in my previous message - if there was connection pool and `tournament_query` was executed outside of transaction - it won't see changes brought by `.create()` which leads to errors. 
And if we talk about modification - `.delete()` also modifies content.

So I think all operations are equally important in terms of  their interactions with transactions

grigi: Ok, I added tests for insert/update/delete/select and they all work as I expected: https://github.com/tortoise/tortoise-orm/pull/64/files#diff-12a2be3f481afdc9a647b690a7a299f9
So, for now I think this is OK.

And it becomes the problem of when the connection pooling happens.

Zeliboba5:
So far this PR seems complete to me. 
Are you going to add anything or we are good to merge it?

PR #65, opened 2018-11-13
Feature/pypy

grigi:
Add Travis & Tox testing on PyPy, because why not.

And added supported Py versions with exception of PostgreSQL for pypy.


grigi:
@Zeliboba5 Please review.

Zeliboba5:
Great, haven't thought that it would just work without additional tuning :)

PR #67, opened 2018-11-18
Type annotations & tests

grigi:
Add lots of type annotations
Add some tests


grigi:
@Zeliboba5 I think this is enough churn for now. Please review.
Should only be extra typing info and tests...

PR #69, opened 2018-11-26
New pool (WIP)

grigi:
Now that code base is simpler, and test runner should be more sane, attempt to add connection pooling for the third time.

The plan is to change to connection pooling ONLY, as we currently implement persistent connections, but only one persistent connection.
A connection pool should:
* add robustness (if connection dies, then reconnect)
* Allow multiple DB clients to operate at the same time (up to maxsize)
* Allow more conflicts, so we need to handle rollback/retries explicitly.

Things done:
* [x] Change to a connection pooling system for MySQL.
* [ ] Add tests for concurrency
* [ ] Add tests for robustness (hackery allowed)
* [ ] Add tests for handling conflicts.

Concerns:
* Can SQLite be concurrent at all? If difficult, should we limit it?
* We need to add concurrency to the benchmarks, to manage performance

### DIFF
--- a/tortoise/backends/base/client.py
+++ b/tortoise/backends/base/client.py
@@ -87,8 +87,8 @@ class BaseDBAsyncClient:
     async def execute_script(self, query: str) -> None:
         raise NotImplementedError()  # pragma: nocoverage
 
-    async def execute_explain(self, query: str) -> Sequence[dict]:
-        raise NotImplementedError()  # pragma: nocoverage
+    # async def execute_explain(self, query: str) -> Sequence[dict]:
+    #     raise NotImplementedError()  # pragma: nocoverage
 
 
 class ConnectionWrapper:
--- a/tortoise/backends/base/config_generator.py
+++ b/tortoise/backends/base/config_generator.py
@@ -47,6 +47,8 @@ DB_LOOKUP = {
         },
         'defaults': {
             'port': 3306,
+            'minsize': 1,
+            'maxsize': 5,
         },
         'cast': {
             'minsize': int,
--- a/tortoise/backends/mysql/client.py
+++ b/tortoise/backends/mysql/client.py
@@ -14,6 +14,8 @@ from tortoise.exceptions import (DBConnectionError, IntegrityError, OperationalE
                                  TransactionManagementError)
 from tortoise.transactions import current_transaction_map
 
+logger = logging.getLogger(__name__)
+
 
 def translate_exceptions(func):
     @wraps(func)
@@ -29,6 +31,23 @@ def translate_exceptions(func):
     return wrapped
 
 
+class MysQLConnectionWrapper(ConnectionWrapper):
+    __slots__ = ('pool', )
+
+    def __init__(self, pool) -> None:
+        super().__init__(None)
+        self.pool = pool  # type: aiomysql.Pool
+
+    async def __aenter__(self):
+        self.connection = await self.pool.acquire()
+        logger.debug('Acquired connection %s', self.connection)
+        return self.connection
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        logger.debug('Released connection %s', self.connection)
+        self.pool.release(self.connection)
+
+
 class MySQLClient(BaseDBAsyncClient):
     query_class = MySQLQuery
     executor_class = MySQLExecutor
@@ -36,7 +55,7 @@ class MySQLClient(BaseDBAsyncClient):
     capabilities = Capabilities('mysql', safe_indexes=False, requires_limit=True)
 
     def __init__(self, user: str, password: str, database: str, host: str, port: SupportsInt,
-                 **kwargs) -> None:
+                 minsize: SupportsInt, maxsize: SupportsInt, **kwargs) -> None:
         super().__init__(**kwargs)
 
         self.user = user
@@ -44,8 +63,10 @@ class MySQLClient(BaseDBAsyncClient):
         self.database = database
         self.host = host
         self.port = int(port)  # make sure port is int type
+        self.minsize = int(minsize)
+        self.maxsize = int(maxsize)
 
-        self._connection = None  # Type: Optional[aiomysql.Connection]
+        self._connection = None  # Type: Optional[aiomysql.Pool]
 
         self._transaction_class = type(
             'TransactionWrapper', (TransactionWrapper, self.__class__), {}
@@ -58,29 +79,35 @@ class MySQLClient(BaseDBAsyncClient):
             'user': self.user,
             'password': self.password,
             'db': self.database if with_db else None,
+            'minsize': self.minsize,
+            'maxsize': self.maxsize,
             'autocommit': True,
         }
 
         try:
-            self._connection = await aiomysql.connect(**template)
+            self._connection = await aiomysql.create_pool(**template)
             self.log.debug(
-                'Created connection %s with params: user=%s database=%s host=%s port=%s',
-                self._connection, self.user, self.database, self.host, self.port
+                'Created pool %s with params: user=%s database=%s host=%s port=%s minsize=%s'
+                ' maxsize=%s', self._connection, self.user, self.database, self.host, self.port,
+                self.minsize, self.maxsize
             )
         except pymysql.err.OperationalError:
             raise DBConnectionError(
-                "Can't connect to MySQL server: "
-                'user={user} database={database} host={host} port={port}'.format(
-                    user=self.user, database=self.database, host=self.host, port=self.port
+                'user={user} database={database} host={host} port={port} minsize={minsize}'
+                ' maxsize={maxsize}'.format(
+                    user=self.user, database=self.database, host=self.host, port=self.port,
+                    minsize=self.minsize, maxsize=self.maxsize
                 )
             )
 
     async def close(self) -> None:
         if self._connection:  # pragma: nobranch
             self._connection.close()
+            await self._connection.wait_closed()
             self.log.debug(
-                'Closed connection %s with params: user=%s database=%s host=%s port=%s',
-                self._connection, self.user, self.database, self.host, self.port
+                'Closed pool %s with params: user=%s database=%s host=%s port=%s minsize=%s'
+                ' maxsize=%s', self._connection, self.user, self.database, self.host, self.port,
+                self.minsize, self.maxsize
             )
             self._connection = None
 
@@ -100,7 +127,7 @@ class MySQLClient(BaseDBAsyncClient):
         await self.close()
 
     def acquire_connection(self) -> ConnectionWrapper:
-        return ConnectionWrapper(self._connection)
+        return MysQLConnectionWrapper(self._connection)
 
     def _in_transaction(self):
         return self._transaction_class(self.connection_name, self._connection)
@@ -131,15 +158,20 @@ class MySQLClient(BaseDBAsyncClient):
 
 
 class TransactionWrapper(MySQLClient, BaseTransactionWrapper):
-    def __init__(self, connection_name, connection):
+    def __init__(self, connection_name, pool):
         self.connection_name = connection_name
-        self._connection = connection
+        self._pool = pool
         self.log = logging.getLogger('db_client')
         self._transaction_class = self.__class__
         self._finalized = False
         self._old_context_value = None
 
+    def acquire_connection(self) -> ConnectionWrapper:
+        return ConnectionWrapper(self._connection)
+
     async def start(self):
+        self._connection = await self._pool.acquire()
+        self.log.debug('Acquired connection for transaction %s', self._connection)
         await self._connection.begin()
         current_transaction = current_transaction_map[self.connection_name]
         self._old_context_value = current_transaction.get()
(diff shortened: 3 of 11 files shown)


grigi:
There is several useful changes in this PR that I'm going to pull out in its own PR:
* contextvars fixes
* autocommit for MySQL
* Less confusing logs
* Isolation fixes in test runner
* db_url improvements
* minor test enchancements
* dependency updates

grigi:
@abondar Also whilst debugging the idiotic autocommit issue, I found that we don't have a clear expectation of how transactions should operate.

I'm proposing this:
* Root level: `auto-commit`
* First transaction: `full isolation`
* Nested transaction: `no-op`: These are all handled differently depending on version of DB, connector used, etc... and with the current implementation may just create a separate transaction on a different connection causing even more confusion.

grigi:
@abondar I rebased and simplified this connection pooling PR.
Right now only the `*await_across_transaction*` are failing. You are welcome to have a go yourself.

abondar:
@grigi For some reason python3.7/mysql build failed crazily, but other builds are now fine. I'll look into failed build later if you won't manage to do it before me

grigi:
My first thought was "Probably contextvars" as the backport `aiocontextvars` isn't exactly the same as the py3.7 version.
Looking at the test results, it looks like test isolation completely failed. I get more errors with less processes and less errors with more processes, and if I run a test by itself it always passes...

grigi:
It doesn't use the transactioned connection, but a different one:
```
2018-12-28 08:43:32     DEBUG Acquired connection for transaction <aiomysql.connection.Connection object at 0x7fcbf47fbb00>
2018-12-28 08:43:32     DEBUG Acquired connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG INSERT INTO `tournament` (`name`,`created`) VALUES (%s,%s): ['Test', datetime.datetime(2018, 12, 28, 6, 43, 32, 189123)]
2018-12-28 08:43:32     DEBUG Released connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG Acquired connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG UPDATE `tournament` SET `name`='Updated name' WHERE `id`=1
2018-12-28 08:43:32     DEBUG Released connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG Acquired connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG SELECT `created`,`name`,`id` FROM `tournament` WHERE `name`='Updated name' LIMIT 1
2018-12-28 08:43:32     DEBUG Released connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG Acquired connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG INSERT INTO `tournament` (`name`,`created`) VALUES (%s,%s): ['Test 2', datetime.datetime(2018, 12, 28, 6, 43, 32, 195122)]
2018-12-28 08:43:32     DEBUG Released connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG Acquired connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG SELECT `id` `0` FROM `tournament`
2018-12-28 08:43:32     DEBUG Released connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG Acquired connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG SELECT `id` `id`,`name` `name` FROM `tournament`
2018-12-28 08:43:32     DEBUG Released connection <aiomysql.connection.Connection object at 0x7fcbf4b24ac8>
2018-12-28 08:43:32     DEBUG Released connection for rolled back transaction <aiomysql.connection.Connection object at 0x7fcbf47fbb00>
```
It creates a transaction on connection `0x7fcbf47fbb00`, but then runs all SQL on connection `0x7fcbf4b24ac8`.
Which to me does confirm a `contextvars` related issue, as that is where the connection should be stored?
I'm sure I fixed this, but maybe in the last rebase I broke something? 

This should be unrelated to your queryset changes.

grigi:
Yup, it was already broken on my rebase, and was working before then.
Meh, so much changed :-(

grigi:
Ok, I did some more digging, and found that I was mistaken. py37+mysql (pooling) never isolated right.
And if I use contextvars with `reset()` I will always get an error about context having changed.
Meaning somehow the context changes, so the value we set is not seen where we think it should.
So we may have to manage the context manually? Or I may be missing something.

There is some documentation re the differences here: https://pypi.org/project/aiocontextvars/
I feel if we can get the complaint of different Context when we use `reset()` to go away, we will probably have contextvars working properly on 3.7

grigi:
Currently the client looks up a ContextVar that exist globally, outside of its own scope. I think the common issues for globals may actually be relevant here. So I'm going to consider removing the global state tracking and try and simplify the transaction handling in the Client. To make it easier to reason about things.

It is easy to trigger the current issue, and it appears to be a very easy race-condition. Unfortunately the code isn't that easy to reason about, hence my attempt at simplifying it some more.

abondar:
> Currently the client looks up a ContextVar that exist globally, outside of its own scope. I think the common issues for globals may actually be relevant here. 

I don't really understand why there is problem with contextvars, shouldn't it be okay for them to be global, because that what it shows in examples in docs https://docs.python.org/3/library/contextvars.html ?
And what is alternative way to store those variables?

grigi:
Yes, the docs talk of globals, but how I see it being used in Gino is as part of a instance-variable.
There is a behaviour difference between the backport and what is on 3.7 and I don't really know why 3.7 fails for us. 

Also, the way transactions sort-of replace the class just doesn't feel clean, hence me considering refactoring it.
e.g. The Client contains the pool, and the entirety of transaction management could be a ContextVar inside the class.

zhoufeng1989:
Hi, guys. I am using tortoise ORM (PostgreSQL backend) in project. Now I am considering applying database pool (asyncpg pool) in tortoise.

It seems that in 'new_pool' branch, you guys have already implemented pool for MySQL.
I run tests in 'new_pool' branch, four of the test cases fail:
- tortoise.tests.test_capabilities.TestCapabilities.test_actually_runs 
- tortoise.tests.test_capabilities.TestCapabilities.test_connection_name
- tortoise.tests.test_tester.TestTesterASync.test_fail
- tortoise.tests.test_tester.TestTesterASync.TestTesterSync.test_fail

But these fails seem irrelevant to database pool.  I would like to know the progress of this feature. Are problems you talked in this issue still unsolved?  If I add pool for PostgreSQL, is there anything tortoise-specific I need to know about? Can I start with this branch?

Thanks for your excellent work for this library!

@grigi @abondar

grigi:
Hi @zhoufeng1989 
Thank you for your interest! We have done a bit more work on the problem, and understand the issues a bit better.

So, for a short history as to why this have been sitting around so long:
At first we wanted to push for this to get `gather()` operations to work without issues, somewhat naïvely as with pooling is how one solves concurrency in a threaded environment?
But we ran into these odd hard-to-debug issues. (There is another PR where I essentially took all the bits I found helped out as separate PRs, and this is essentially the second attempt).

We then had a few concurrency related bug reports, and I managed to fix all of them quite easily except for one. That was when mutex around a transaction. The fail was identical to this.
I shelved that here: https://github.com/tortoise/tortoise-orm/commit/51b8d208eca645581aa05b6ec6465899a2b5630d And now that I had a small diff I could try and understand why I could get this to work on Py3.5 & Py3.6, but it would fail on Py3.7

The obvious difference is that async contextvars is built-in on Py3.7, and a monkey patch on 3.5/3.6

So it turns out that the context gets applied at higher resolution on py3.7, so would be tied to stack level, instead of to co-routine scheduling. So we need to control the stack a bit better so that we can enable this reliably. I suppose that the "working way" on 3.5/3.6 was probably not entirely reliable.

Then life interfered (Kid got very sick) but that is gradually returning to normal. So for a while most of the dev work was done by contributors, and only recently (the last 2 months or so) I actually did real work on this again.

If you read on #141 Andrey re-attempted it, but ran ultimately into the same issue, and after discussing the core issue as I see it (and I may be wrong) he agreed that we may have to do a large refactor to fix the stacking issue.

So, yes, if you want to contribute in any way to help us reach the milestone reliably, I would be super grateful :grinning: 

I would start with that commit I linked, and see how restructuring the code, could enable the stack-level resolution of the built in contextvars to work.
We now have a relaxed requirement that we just deprecated py3.5 so any work on a large feature can happily break py3.5 compatibility.
So now we can use cleaner, simpler generators: https://docs.python.org/3.7/whatsnew/3.6.html#whatsnew36-pep525 which I think could really help simplify the flow of co-routines, so might allow us to fix this issue without a massive refactor which may or may not fix the issue.

Wow, that was a lot longer than I intended, but I hope it gives enough context.
I want that issue resolved,

grigi:
Hi @zhoufeng1989 
Thank you for your interest! We have done a bit more work on the problem, and understand the issues a bit better.

So, for a short history as to why this have been sitting around so long:
At first we wanted to push for this to get `gather()` operations to work without issues, somewhat naïvely as with pooling is how one solves concurrency in a threaded environment?
But we ran into these odd hard-to-debug issues. (There is another PR where I essentially took all the bits I found helped out as separate PRs, and this is essentially the second attempt).

We then had a few concurrency related bug reports, and I managed to fix all of them quite easily except for one. That was when mutex around a transaction. The fail was identical to this.
I shelved that here: https://github.com/tortoise/tortoise-orm/commit/51b8d208eca645581aa05b6ec6465899a2b5630d And now that I had a small diff I could try and understand why I could get this to work on Py3.5 & Py3.6, but it would fail on Py3.7

The obvious difference is that async contextvars is built-in on Py3.7, and a monkey patch on 3.5/3.6

So it turns out that the context gets applied at higher resolution on py3.7, so would be tied to stack level, instead of to co-routine scheduling. So we need to control the stack a bit better so that we can enable this reliably. I suppose that the "working way" on 3.5/3.6 was probably not entirely reliable.

Then life interfered (Kid got very sick) but that is gradually returning to normal. So for a while most of the dev work was done by contributors, and only recently (the last 2 months or so) I actually did real work on this again.

If you read on #141 Andrey re-attempted it, but ran ultimately into the same issue, and after discussing the core issue as I see it (and I may be wrong) he agreed that we may have to do a large refactor to fix the stacking issue.

So, yes, if you want to contribute in any way to help us reach the milestone reliably, I would be super grateful :grinning: 

I would start with that commit I linked, and see how restructuring the code, could enable the stack-level resolution of the built in contextvars to work.
We now have a relaxed requirement that we just deprecated py3.5 so any work on a large feature can happily break py3.5 compatibility.
So now we can use cleaner, simpler generators: https://docs.python.org/3.7/whatsnew/3.6.html#whatsnew36-pep525 which I think could really help simplify the flow of co-routines, so might allow us to fix this issue without a massive refactor which may or may not fix the issue.

Wow, that was a lot longer than I intended, but I hope it gives enough context.

I want that issue resolved, because once we just get the locking around transactions done, we now have prefect isolation. 
Then adding pools or different behaviour for in-or-out of transaction work (like for bulk operations) would be trivial.
And that would get us very close to where we consider ourselves full production ready YAY!

zhoufeng1989:
Hi @grigi , thanks for your informative reply, but I still feel kind of confused about the problem that fails on Py3.7 but works on Py3.5 & Py3.6.

I am trying to reproduce the problem, but got some other problems.  all my work is based on the latest release 0.13.0.

- Since I only got PostgreSQL now, can I run unit tests only on PostgreSQL? I tried to set environment variable `TORTOISE_TEST_DB` to something like `postgres://postgres:@127.0.0.1:5432/test_tortoise` and also read `db_url` in `conftest.py` from this environment variable, but it seems that some test cases still use sqlite (as I debug, the current_transaction_map is something like `{'default': <ContextVar name='default' default=<tortoise.backends.sqlite.client.SqliteClient object at 0x10b60e9b0> at 10b60e978>}`).

- Would you please show me the test cases that fails on Py3.7 but works on Py3.5 & 3.6? Maybe I can understand better with these test cases.

Thanks!

grigi:
Yes, some tests manually create DB contexts, e.g. for testing db-specific DDL generation. (I think it should not actually create a DB instance, just a client object). It should be limited to `test_connection_params.py` and `test_generate_schema.py`.

I can't remember exactly which ones were failing on py3.7, but the most important tests for pooling/concurrency is `test_concurrency.py` and `test_transactions.py`.
Also if you can get this one working fine: https://github.com/tortoise/tortoise-orm/commit/51b8d208eca645581aa05b6ec6465899a2b5630d#diff-1ac94b73e8bd2ab0fdcd9d23877a3665R30-R38
Then we are probably all good to go.

PR #74, opened 2018-12-22
Support field subclassing

arlyon:
This PR adds support for subclassing the already existing fields, as long as the signatures of the `to_db_value` and `to_python_value` functions are the same. It also includes an example implementation of an enum field that extends the string field, and uses it when serializing and de-serializing with the database. 

Currently, the `schema_generator` looks up the type in the `FIELD_TYPE_MAP` which makes subclassing impossible (and throw a KeyError). Now, it will continue looking up the chain (looking only at the "first" parent) until it finds a parent that is a supported `Field`, and use that.

### Some things to consider

- Currently the only way to subclass is to completely override the serialization function. Maybe, instead of overriding, having a hook that the parent calls would be a better approach?
- This could pave the way for data validation (such as only even numbers for example) but it really depends on where you want to draw the line
- Wasn't quite sure where to put the tests, so let me know where they need to go
- The GIS stuff is coming along, but as with most things, it's taking a while...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+0.10.12
+-------
+- Support subclassing on existing fields
+
 0.10.11
 -------
 - Pre-build some query & filters statically, 15-30% speed up for smaller queries.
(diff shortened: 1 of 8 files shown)


grigi:
FYI, The test failures should be replicable by running `make ci`, in particular it seems to be `isort` related, so `isort <offending-file>` should fix it.

grigi:
And update `CHANGELOG.rst` please :-)

arlyon:
I discovered a pain point with the filters but documented it. Currently all the filters`Model.filter(field__isnull=True)` go through the `to_db_value` and `to_python_value` which means overwriting these functions requires you to account for all the possible filter values (bool, list, str at least)

PR #75, opened 2018-12-24
Many small updates.

grigi:
This takes the useful parts from #69:

Bugfixes:
* contextvars update & testrunner isolation fixes
* autocommit for MySQL
* re-ordered execute-SQL log to be inside connection context manager for less confusing logs.

Enhancements:
* db_url improvements
* minor test enhancements
* test dependency updates


abondar:
`autocommit for MySQL``
Does that mean that currently MySQL backend doesn't commit at all?

grigi:
Almost. Mysql committed in transactions.
But otherwise yes, data wasn't being committed.
Which makes me wonder how much it was used...

PR #76, opened 2018-12-25
Add "safe" schema generation

arlyon:
This PR adds a "safe" mode when creating the schemas (essentially identical to [the safe parameter on peewee's `create_table` function](http://docs.peewee-orm.com/en/latest/peewee/api.html#Model.create_table)).

## Description

Adds a safe parameter to the `generate_schemas` function that inserts the `IF NOT EXISTS` clause when creating the table.

## Motivation and Context

Peewee ORM has the option to generate schemas "safely" by inserting `if not exists` clause to the SQL. This feature is a nice quality of life, especially when deploying to heroku for example where maintaining some state about whether you've created the tables can be a problem. Currently the only way to emulate this is to catch the exception and pass, but this means that if you were to try to add a model to an existing database, it would except after trying to add the first, and fail on the others.

## Issues

Peewee defaults to True on this setting, so I have done the same. However, setting it to insert the clause by default could be seen as a breaking change. Maybe it's fit for version 0.11 in that case? Or should it instead default to False? Either is valid. On one hand, following peewee's example is probably a good idea, however it seems backward to explicitly state you *don't* want to include the clause.

## How Has This Been Tested?

All it does is add the `IF NOT EXISTS` clause. It's just a boolean and an if so there are two tests to check if the clause is inserted when set to True, and not otherwise.

To preserve the functionality of the tests, the safe parameter is [set to `False`](https://github.com/tortoise/tortoise-orm/pull/76/files#diff-2b8db23f3d3d9db1db66e77053e67911R21).

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+0.11.2
+------
+- Added "safe" schema generation
+
 0.11.1
 ------
 - Test class isolation fixes & contextvars update
--- a/README.rst
+++ b/README.rst
@@ -136,7 +136,9 @@ Tortoise ORM currently supports the following databases:
 * PostgreSQL (requires ``asyncpg``)
 * MySQL (requires ``aiomysql``)
 
-``generate_schema`` generates schema on empty database, you shouldn't run it on every app init, run it just once, maybe out of your main code.
+``generate_schema`` generates the schema on an empty database. Tortoise generates schemas in safe mode by default which
+includes the `IF NOT EXISTS` clause, so you may include it in your main code.
+
 
 After that you can start using your models:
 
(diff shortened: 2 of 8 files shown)


grigi:
Also, we have stated that from version 1.0 we are going to have a stable api, but this change is done "compatibly" (see https://emptysqua.re/blog/api-evolution-the-right-way/) so I see no reason to force a gradual behaviour change for this.

arlyon:
Ah no I renamed the string formatting variable name from "contents" to "fields" last minute and forgot to reflect it in the mysql backend. Fixed now.

abondar:
@arlyon Also, could you please edit README regarding `generate_schema` - at the moment it states that you can run it only once

Issue #86, opened 2019-01-16
Capability declaration / detection on dialects and engines


arlyon:
This was mentioned in the GIS discussion but never made it into a seperate issue.

Having a client being aware of the capabilities of its connection is beneficial in a number of situations.

I am considering how to do this + #72 together since they are somewhat related and thought some discussion would be good first around how the two would interact. Some unordered thoughts about how it would work:

### Defining Capabilities

- `{dialect}+{engine}://...` syntax is nice
  - `postgis+asyncpg://...`
  - `sqlite+aiosqlite://...`
  - `mysql+odbc://...`
  - can still omit the engine to fall back on the default engine for a given dialect: `sqlite://`
- it seems obvious dialects should have a set of supported features defined:
  - postgres `DISTINCT ON` or JSON fields
  - postGIS / spatialite functions
- what about engines? 
  - example, aiosqlite did not support loading sqlite extensions, which is important for spatialite
  - odbc may need to limit capabilities
  - any situations where a client would *add* capabilities?
- if a user picks a silly combination (ex. `postgis+mysql://...`) should we warn them or let it fail? maybe an engine should define which dialects it supports (and postgis would just subclass pg)

### Using Capabilities

- a client is a combination of a dialect and an engine
- capabilities are calculated and kept on the client, representing everything that can be done on the connection
- capabilities can be queried:
  - does this connection support GIS operations?
  - does this connection support table inheritance
- Then, if for example, you were to use a GIS function on a connection that doesn't support it we can be aware of it before making the query.

### Proposed Changes

- rename the backends to the actual engine they are using
  - `mysql` -> `aiomysql`
- move them into an `backends.engines` package
- add a `backends.dialects` package that contains the supported dialects and their features
- create a new client class that "picks" a dialect and an engine
- generate a combined "capabilities" object on the client that passes the capabilities of the dialect through the engine and allow it to remove (or add) things as needed
- allow the query generator to access these capabilities

Then, with the GIS module in mind, it would be nice to be able to register new dialects. For example, having `spatialite` subclass `sqlite` to add the location capability.

grigi:
You are welcome to do it.
I was considering doing this with #72, and have a few thoughts already:

# I'm thinking something along the lines of:
* rename `tortoise.backends` → `tortoise.dialects`
* move the `client.py` out to a `tortoise.connectors` package
* move the `base.config_generator` out to the top package (and extend it to understand dialect/connector)
[...]

PR #89, opened 2019-01-27
WIP: Implement field indexes

### LINKED ISSUE #78: db_index parameter?
I miss the ability to mark fields as indexed. I don't know if this is a limitation of pypika. I need this feature to ensure that DB queries are decently fast. :)

florimondmanca:
## Description
- Add a new `index` boolean parameter to the base `Field` constructor.
- When `True`, an index is created for that field. This is implemented by adding a `CREATE INDEX` statement in the table creation query.
- If `safe=True`,  the index is only created if it doesn't exist already. For this we use a deterministic index name. Not supported on MySQL.

## Motivation and Context
Fixes #78.

If we want to implement multi-field indexes in the future, the index creation generation code may need to be refactored.

## How Has This Been Tested?
- I have changed `Tournament`'s `created` field to declare `index=True`
- New tests verify that a `CREATE INDEX` statement is added for the field when the schema is generated.

## Checklist:

## TODO

- [x] Fix issues with `EXPLAIN` being database-dependant, or find a new way to verify an index was created.
- [x] Use the `safe` flag to only create indexes if they don't exist already.


grigi:
Thanks for this Florimond. Generally the code looks good :+1:

The issue with the tests is due to either there not being a standard explain format #90 #91, or Tortoise not supporting basic driver capability reporting #86 

Ultimately I think BOTH should get solved, but either will unblock this :+1:

florimondmanca:
Yes, the failing jobs are when using MySQL and PostgreSQL which do have a different syntax for the EXPLAIN statement than SQLite.

grigi:
@florimondmanca Please review #92 to see if it would help unblock this?

florimondmanca:
Yes it would: I'd mark the test with `@requireCapability(dialect="sqlite")` so it is skipped on MySQL and PostgreSQL. 😄

tortoise/backends/base/schema_generator.py
```diff
@@ -83,13 +105,23 @@ def _get_table_sql(self, model, safe=True) -> dict:
...
         table_fields_string = ', '.join(fields_to_create)
         table_create_string = self.TABLE_CREATE_TEMPLATE.format(
             exists="IF NOT EXISTS " if safe else "",
             table_name=model._meta.table,
             fields=table_fields_string,
         )
 
+        for field_name in fields_with_index:
+            table_create_string += self.INDEX_CREATE_TEMPLATE.format(
+                index_name=self._generate_index_name(model, [field_name]),
```
florimondmanca: Note to self: need to add an `IF NOT EXISTS` check here if `safe == True`.

tortoise/backends/mysql/schema_generator.py
```diff
@@ -17,9 +20,8 @@ def __init__(self, *args, **kwargs) -> None:
         )
 
         self.FIELD_TYPE_MAP.update({
-            fields.FloatField: 'DOUBLE',
-            fields.DatetimeField: 'DATETIME(6)',
-            fields.TextField: 'TEXT'
+            fields.FloatField: 'DOUBLE', fields.DatetimeField: 'DATETIME(6)', fields.TextField:
+            'TEXT'
```
florimondmanca: @grigi I turned on yapf in my text editor and this is how it reformatted the code. Looks strange to me, but I haven't used yapf yet so I don't know if this is normal?

grigi: We tried code formatters, both black and yapf, and was unsatisfied with either of them.
We decided it was better to stick with human formatted code, as it was much more readable.
Generally we are not too strict re style, flake8 checks for the worst offenders, and then it should be readable. 😐

florimondmanca:
Actually, now that `.explain()` has been implemented, I was able to generalise the tests so they can run on all 3 databases.

However, I'm running into the fact that MySQL doesn't have an `IF NOT EXISTS` option for indexes. 😞  I've looked for ideas on how to work around that, but it's particularly nasty. The [most upvoted answers](https://dba.stackexchange.com/questions/24531/mysql-create-index-if-not-exists) use queries I'm not even sure we can run in the schema creation code.

Any clues @grigi ?

florimondmanca:
Yes, my intuition is that there's something bigger that needs to be handled to solve this issue with MySQL. It seems using some sort of migration-based schema generation instead of "forcing" the schema to be created on every initialisation would allow us to just create the index once and not have to deal with this kind of issue.

Anyway, for now I will use the new capabilities API and note that MySQL doesn't support field indexes. Then if using MySQL, raise a warning during schema generation saying that we're skipping index creation.

grigi:
Well, there is a step between just emitting code, and full on migrations. That is to synchronize. 
I think it is fine to emit the index SQL when `safe=False`, else warn that we don't support safely creating indexes in MySQL, so we ask the user to apply them manually for now.
You could emit the "dangerous" statement as part of the warning. or as a second warning.

florimondmanca:
FYI, the build is currently failing due to an issue with `@requireCapability` which I've submitted a fix for in #98.

florimondmanca:
… Tests pass! :tada: ✅

tortoise/backends/base/schema_generator.py
```diff
@@ -83,13 +114,34 @@ def _get_table_sql(self, model, safe=True) -> dict:
...
             exists="IF NOT EXISTS " if safe else "",
             table_name=model._meta.table,
             fields=table_fields_string,
         )
 
+        # Indexes.
+        field_indexes_sqls = [
+            self._get_index_sql(model, [field_name], safe=safe) for field_name in fields_with_index
+        ]
+        if safe and not self.client.capabilities.safe_indexes:
```
grigi: :+1: Nice to have generic implementation based off capabilities

tortoise/contrib/test/__init__.py
```diff
@@ -258,10 +261,28 @@ def requireCapability(connection_name: str = 'models', **conditions: Any):
...
-            for key, val in conditions.items():
-                if getattr(db.capabilities, key) != val:
-                    raise SkipTest('Capability {key} != {val}'.format(key=key, val=val))
+            checkCapability(connection_name, **conditions)
             return test_item(*args, **kwargs)
         return skip_wrapper
     return decorator
+
+
+def checkCapability(connection_name: str = 'models', **conditions: Any):
```
grigi: The terminology is unexpected.
What this does is skip the test in the middle of it, if the capability is not set up.
And the only use-case is for testing capabilities when testing schema generation.
Makes me think this is not really necessary for being part of the API?

florimondmanca: So, should I instead get the connection, check for capabilities and skip the test (essentially copying the behavior of `requireCapability` but not as a decorator) in these specific tests?

florimondmanca:
> Can you please also look at making the `safe=True` optional parameters for the schema functions in `tortoise/utils.py` non-optional.

Done!

tortoise/contrib/test/__init__.py
grigi: I am in two minds about it. Primarily I'm unhappy with the naming, then I thought about if we really want to expose a piece of API that we probably don't need to expose.
So I'd say copy the code into the test module.

grigi: I thought Travis was set up that way... Let me check.
It is like that. It only auto cancels a job that haven’t been started unfortunately.

grigi:
Thanks for your good work. I released v0.11.3 :+1:

PR #91, opened 2019-01-28
Add `QuerySet.explain()`

### LINKED ISSUE #90: Explain support
EXPLAIN is an SQL operation that outputs the execution plan and allows to know how exactly the database will perform a query. It's very useful for debugging.

Calling `.explain()` on a queryset should return the execution plan as a string:

```python
>>> plan = Tournament.filter(id=4).explain()  # type: str
```

As it is meant for debugging, the actual content of the execution plan can (and will) vary across databases.

I've also noted that:

- Not all databases use the same prefix, e.g. MySQL et PostgreSQL use `EXPLAIN` while SQLite uses `EXPLAIN QUERY PLAN` (it also has `EXPLAIN` but it returns lower-level information).
- The output of an `EXPLAIN` query may vary between databases. For example, [MySQL specifies the EXPLAIN output columns](https://dev.mysql.com/doc/refman/8.0/en/explain-output.html#explain-output-columns) while [SQLite says it may even vary across releases](https://www.sqlite.org/lang_explain.html).
- There are various `EXPLAIN` formats:
   - [MySQL](https://dev.mysql.com/doc/refman/8.0/en/explain.html) has `TRADITIONAL` and `JSON`
   - [PostgreSQL](https://www.postgresql.org/docs/9.1/sql-explain.html) has a few, but [asyncpg already has an explain() method](https://github.com/MagicStack/asyncpg/blob/78ea13a8582038b931dad74c746b94ed7606ffc6/asyncpg/prepared_stmt.py#L111) (accessible through `connection.prepare().explain()`) that hardcodes the `JSON, VERBOSE` format. However, it should be easy to re-implement `explain()` with variable format ourselves, if needed.

Because of the above, I'm not sure whether making us able to specify an `EXPLAIN` format (e.g. `queryset.explain(format="YAML")` would be worth it because, again, `.explain()` would be for debugging so I'm afraid users may start relying on the actual output if we let them specify a `format`.

**Additional context**
I wanted to use it to test field indexes (#89) but I am running into issues because the exact format is database-dependant, so I figured a separate issue would be better.

florimondmanca:
## Description
This PR adds support for executing an `EXPLAIN` statement against the query represented by a `QuerySet`.

## Motivation and Context
Fixes #90 

## Checklist:

## TODO:

- [x] Implement the base explain API (`QuerySet`, `Executor`, `AsyncDBClient`)
- [x] Implement for PostgreSQL
- [ ] Better format the PostgreSQL output
- [ ] Implement for MySQL
- [ ] Implement for SQLite
- [ ] Add tests
- [ ] Add docs

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -105,6 +105,7 @@ class QuerySet(AwaitableQuery):
         self._annotations = {}  # type: Dict[str, Aggregate]
         self._having = {}  # type: Dict[str, Any]
         self._custom_filters = {}  # type: Dict[str, dict]
+        self._explain = False  # type: bool
 
     def _clone(self) -> 'QuerySet':
         queryset = QuerySet.__new__(QuerySet)
@@ -127,6 +128,7 @@ class QuerySet(AwaitableQuery):
         queryset._annotations = copy(self._annotations)
         queryset._having = copy(self._having)
         queryset._custom_filters = copy(self._custom_filters)
+        queryset._explain = self._explain
         return queryset
 
     def _filter_or_exclude(self, *args, negate: bool, **kwargs):
@@ -372,6 +374,11 @@ class QuerySet(AwaitableQuery):
                 queryset._prefetch_map[first_level_field].add(forwarded_prefetch)
         return queryset
 
+    def explain(self, format: Optional[str] = None, **options) -> 'QuerySet':
+        queryset = self._clone()
+        queryset._explain = True
+        return queryset
+
     def using_db(self, _db: BaseDBAsyncClient) -> 'QuerySet':
         """
         Executes query in provided db client.
@@ -412,12 +419,18 @@ class QuerySet(AwaitableQuery):
 
     async def _execute(self):
         self.query = self._make_query()
-        instance_list = await self._db.executor_class(
+
+        executor = self._db.executor_class(
             model=self.model,
             db=self._db,
             prefetch_map=self._prefetch_map,
             prefetch_queries=self._prefetch_queries,
-        ).execute_select(
+        )
+
+        if self._explain:
+            return await executor.execute_explain(self.query)
+
+        instance_list = await executor.execute_select(
             self.query, custom_fields=list(self._annotations.keys())
         )
         if not instance_list:
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -85,7 +85,7 @@ class QuerySet(AwaitableQuery):
     __slots__ = ('fields', '_prefetch_map', '_prefetch_queries',
                  '_single', '_get', '_count', '_db', '_limit', '_offset', '_filter_kwargs',
                  '_orderings', '_q_objects', '_distinct',
-                 '_annotations', '_having', '_custom_filters')
+                 '_annotations', '_having', '_custom_filters', '_explain')
 
     def __init__(self, model) -> None:
         super().__init__(model, None)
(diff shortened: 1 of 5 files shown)


grigi:
The one thing that concerns me with explain, is that the output of it varies a lot between different databases, and even within the same database.
Some DB's even change their output depending on the data in the table, and statistics built from it.

I see two use cases of this:
1) Tests
2) Users perusing the DB manually and trying to tune performance

Therefore I'm feeling that this would be a perfectly valid feature if it is a partially implemented feature forever.

And it use case in tests may very well be database dependant, meaning we once again are blocked by capabilities. I'll have to see about getting that unblocked?

florimondmanca:
Yes, when we get database capabilities, we can add a "supports EXPLAIN" to be able to distinguish between clients that do/do not support this functionality. However, as of today, all databases supported by Tortoise (PostgreSQL, MySQL, SQLite) do support `EXPLAIN` (although they have a slightly different syntax, which I was planning to specify at the async db client level).

Also, as I said in #90, I do not believe that `.explain()` should be anything else than a debugging feature.
It should not be used in anything automated because, as you pointed out, its output does depend on the technology and on the data (for example, with / without indexes).
So, in that sense, I wouldn't even advice users using it in tests either (plus, I'm not even sure in what kind of tests looking at an EXPLAIN output would be useful).

abondar:
How about adding `.sql()` method to get resulting sql? 
I think it could also be useful alongside with `.explain()`

florimondmanca:
Yes, `.sql()` would definitely be useful!

tortoise/queryset.py
```diff
@@ -105,6 +105,7 @@ def __init__(self, model) -> None:
         self._annotations = {}  # type: Dict[str, Aggregate]
         self._having = {}  # type: Dict[str, Any]
         self._custom_filters = {}  # type: Dict[str, dict]
+        self._explain = False  # type: bool
```
grigi: I don't want to add more parameters to QuerySet, as the copying of this is a known performance issue.
Rather make the `.explain()` method immediately execute the explanation?

tortoise/queryset.py
```diff
@@ -372,6 +374,11 @@ def prefetch_related(self, *args: str) -> 'QuerySet':
                 queryset._prefetch_map[first_level_field].add(forwarded_prefetch)
         return queryset
 
+    def explain(self, format: Optional[str] = None, **options) -> 'QuerySet':
+        queryset = self._clone()
+        queryset._explain = True
+        return queryset
```
grigi: As I mentioned, I would rather this be an `async def` and return the explain immediately, so as to minimise adding complexity into `QuerySet`

tortoise/queryset.py
florimondmanca: Yes, that makes sense even from a user perspective, I'll change that. :-)

tortoise/queryset.py
florimondmanca: Yup!

florimondmanca:
Impact on `QuerySet` has been minimized @grigi, as a result the code is much simpler. :+1:

I also added a docstring to be rendered in the docs, which mentions the caveats we've agreed upon, i.e. that `.explain()` is only meant for debugging or performance tuning:

screenshot 2019-01-30 at 15 57 21

grigi:
There is two last little things:
1) Add your name to CONTRIBUTORS.rst
2) Add a reference to the 0.11.3 section in CHANGELOG.rst

PR #92, opened 2019-01-28
Added basic capabilities, test helper, and docs

grigi:
Added basic capabilities with a test decorator `@requireCapability()`

This is a super-basic version of capabilities just to get started.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,9 @@ Changelog
 
 0.11.3
 ------
-* Test runner now has the ability to skip tests conditionally, based on the DB driver Capabilities:
+* Added basic DB driver Capabilities.
+
+  Test runner now has the ability to skip tests conditionally, based on the DB driver Capabilities:
 
   .. code-block:: python3
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,15 @@ Changelog
 
 0.11.3
 ------
-- Performance fix with PyPika for small fetch queries.
+* Test runner now has the ability to skip tests conditionally, based on the DB driver Capabilities:
+
+  .. code-block:: python3
+
+      @requireCapability(dialect='sqlite')
+      async def test_run_sqlite_only(self):
+          ...
+
+- Performance fix with PyPika for small fetch queries
 - Remove parameter hack now that PyPika support Parametrized queries
 - Fix typos in JSONField docstring
 - Added `.explain()` method on `QuerySet`.
@@ -101,7 +109,7 @@ Changelog
 
 0.10.0
 ------
-- Refactored ``Tortoise.init()`` to init all connections and discover models from config passed
+* Refactored ``Tortoise.init()`` to init all connections and discover models from config passed
   as argument.
 
   .. caution::
@@ -159,12 +167,12 @@ Changelog
 
 0.9.0
 -----
-- Added support for nested queries for ``values`` and ``values_list``:
+* Added support for nested queries for ``values`` and ``values_list``:
 
-.. code-block:: python3
+  .. code-block:: python3
 
-    result = await Event.filter(id=event.id).values('id', 'name', tournament='tournament__name')
-    result = await Event.filter(id=event.id).values_list('id', 'participants__name')
+      result = await Event.filter(id=event.id).values('id', 'name', tournament='tournament__name')
+      result = await Event.filter(id=event.id).values_list('id', 'participants__name')
 
 - Fixed ``DatetimeField`` and ``DateField`` to work as expected on SQLite.
 - Added ``PyLint`` plugin.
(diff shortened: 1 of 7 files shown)


tortoise/tests/test_capabilities.py
```diff
@@ -0,0 +1,33 @@
+from tortoise import Tortoise
+from tortoise.backends.base.client import Capabilities
+from tortoise.contrib import test
+
+
+class TestCapailities(test.TestCase):
```
florimondmanca: Typo:

```suggestion
class TestCapabilities(test.TestCase):
```

tortoise/contrib/test/__init__.py
```diff
@@ -238,3 +239,29 @@ class TestCase(SimpleTestCase):
     async def _tearDownDB(self) -> None:
         _restore_default()
         await self.transaction.rollback()
+
+
+def requireCapability(**conditions):
```
florimondmanca: Curious question: why the `camelCase` here (and in other functions in this module)?

tortoise/backends/base/client.py
```diff
@@ -7,6 +7,18 @@
 from tortoise.backends.base.schema_generator import BaseSchemaGenerator
 
 
+class Capabilities:
+    def __init__(self, dialect: str, *, connection: dict = {}, **kwargs) -> None:
+        self.dialect = dialect
+        self.connection = connection
+
```
florimondmanca: Thought experiment here: if we want to know whether client X has capability Y, we'd add a flag here for the base case, e.g. `self.supports_y = False`, and set it to `True` on clients that do support it?

grigi: Yes, that's the idea :-)

tortoise/contrib/test/__init__.py
grigi: the unittest module is full of camelcase. Some examples:
* setUp
* skipIf
* skipUnless
* tearDown
* expectedFailure

tortoise/tests/test_capabilities.py
grigi: :man_facepalming:

arlyon:
These changes look good. Seems fairly simple to override them if needed, which is what we need really. Been busy recently (sorry!) but going to pick up work again on GIS after fosdem this weekend.

As far as I'm concerned this closes #86 :)

CHANGELOG.rst
```diff
@@ -3,9 +3,19 @@ Changelog
...
+
+      @requireCapability(dialect='sqlite')
+      async def test_run_sqlite_only(self):
+          ...
+
+- Performance fix with PyPika for small fetch queries
 - Remove parameter hack now that PyPika support Parametrized queries
 - Fix typos in JSONField docstring
+- Added basic DB driver Capabilities
+- moo
```
abondar: moo?

tortoise/backends/base/client.py
```diff
@@ -7,6 +7,18 @@
 from tortoise.backends.base.schema_generator import BaseSchemaGenerator
 
 
+class Capabilities:
+    def __init__(self, dialect: str, *, connection: dict = {}, **kwargs) -> None:
+        self.dialect = dialect
+        self.connection = connection
```
abondar: Why does capability need connections params?

Also don't think using mutable collections as default value is good. 
Better be safe and set it to `None` and create new dict inside `__init__`:
`self.connection = connection or {}`

CHANGELOG.rst
grigi: :-( Don't write code when sleepy Nick

grigi: There is more than one mistake there. The first & last point is the same.

tortoise/backends/base/client.py
grigi: I treated the Capability object as a config object. Not sure if that is wise or not.
Now that I think of it, it should in-fact be a a read-only informational property.
And yes, mutable defaults are bad. In fact I think Capability should be immutable by default.

grigi: Now that I think of it, the reason is that capabilites can change depending on the connection itself. e.g. we need >= <some-version> of a DB to get the capabilities. or the GIS plugin needs to be installed?

florimondmanca: +1 about the `{}` default, mutable defaults cause all sorts of weird things in Python.

grigi: I changed `Capabilities` to not leak changes made in there back to the DB Client.
Then went a step further and tried to make it behave somewhat immutable (as much as Python allows one to make things immutable). This conflicted with typing and autocompletion, hence my weird `_mutable` thing there.

grigi: NamedTuples are unfortunately more trouble than they are worth. (https://rhye.org/post/python-cassandra-namedtuple-performance/, https://www.attrs.org/en/stable/why.html#namedtuples)
They are immutable, but their content is still references, meaning if you put a mutable type inside it (e.g. dict) then it would allow you to alter state. So to try and prevent accidentally changing it, we would have to pass in deepcopies of everything manually.

I wanted to use attrs, but feel it unfair to introduce as a dependency for just this use case.

Data Classes would also work, but is only part of Py3.7 and backportable to Py3.6, which unfortunately makes them a non-option.

So I settled on a class.

Also, in this case explicit isn't better, as we want to minimise breakage of external drivers.
We state that capabilities are pessimistic by default.

And Yes, since we only case about capabilities we explicity know about, I should probably drop the kwargs part.

florimondmanca: Understood! Makes sense.

grigi: @abondar Any further comments?

PR #94, opened 2019-01-30
WIP: Basic serializers framework

### LINKED ISSUE #13: Serialisation support
Would be great if we could serialise/deserialise the Objects to a dict (and eventually JSON).

It would be perfectly fine if it could not jump a ManyToMany relationship at all.

I really like the way that `attrs` does this, where you can send in arb data that then gets sent to a conversion function and that then instantiates the actual object.
e.g. you can send in an ISO8601 string to a datetime, and configure a converter to auto-convert it to a `datetime.datetime` object for you.

Particuarly we need to have something that would make a CRUDL API simple to implement. I feel actual serialisation could be left as a later concern.

Maybe consider making this NOT be part of the core, but allow an external module to inspect Models easily via a schema-like interface?

florimondmanca:
## Description

This is a basic implementation of a serializers "framework" which would allow to perform the following operations:

```python
from tortoise import fields
from tortoise.serializers import ModelSerializer


class Book(Model):
    title = fields.CharField(max_length=200)
    content = fields.TextField()


class BookSerializer(ModelSerializer):

    class Meta:
        model = Book
        options = {"backend": "pydantic"}


async def main():
    serializer = BookSerializer()

    # Serialization:
    data = serializer.dump(book)

    # Deserialization:
    data = {"title": "foo", "content": "bar"}
    book = await serializer.create(data)
    updated_book = await serializer.update(book, data)

    # Validation only:
    data = {"title": "foo", "content": "bar"}
    validated_data = serializer.validate(data)
```

We have a base class, `BaseSerializer`, which defines anything that can be defined without making further assumptions on the implementation, i.e. the basic API and workflow.

Then, two concrete serializers:
- `Serializer` performs serialization, deserialization and validation as provided by third-party libraries. This is implemented by the concept of "serialization backend", which are used to implement `.to_internal_value()` and `.to_representation()`. It can be defined on the 
- `ModelSerializer` ties everything up and defines how deserialized objects should be saved to the DB by implementing `.create()` and `.update()`.

This is flexible enough to allow using serializers outside the scope of models.

This PR also comes with a serialization backend based on Pydantic. It works by generating Pydantic schemas based on the `Meta.model` attribute.

I'm pretty sure there will be many subtleties and API tweaks to figure out (e.g. how to allow to make the most out of the library behind the serialization backend?), but at least this would provide basic ser/de capabilities.

## Motivation and Context
Fixes #13?

## How Has This Been Tested?
Not tested yet.

## Checklist:

## Other TODOs

- [ ] Add support for dealing with a list of objects, e.g. using a `many=True` option.
- [ ] How can we define a default serialization backend? We wouldn't want to have to specify it everytime. Or should it be auto-detected based on what's installed?

### DIFF
--- a/tortoise/contrib/pydantic/__init__.py
+++ b/tortoise/contrib/pydantic/__init__.py
@@ -1,19 +1,19 @@
 import datetime
 import decimal
 import json
-from typing import Any, Dict, Optional, Set, TypeVar, Tuple, Type, Union  # noqa
+from typing import Any, Dict, Optional, Set, Tuple, Type, TypeVar, Union  # noqa
 
 from tortoise import fields
 from tortoise.fields import Field
 from tortoise.serializers.backends import SerializationBackend
 from tortoise.serializers.datatypes import Data
-from tortoise.serializers.exceptions import ValidationError
+from tortoise.serializers.exceptions import MissingDependencies, ValidationError
 from tortoise.serializers.field_utils import is_read_only
 
 try:
     from pydantic import BaseModel, ValidationError as PydanticValidationError, create_model
 except ImportError as exc:
-    raise exc from None
+    raise MissingDependencies('pydantic') from exc
 
 T = TypeVar('T')
 Definition = Union[Any, Tuple[Type, Any]]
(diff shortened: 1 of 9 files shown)


csrftoken:
It would be nice if the framework could be similar to the django rest framework.

grigi:
Thank you for this, it was used as inspiration for #276 pr. I'm closing now as not needed.

PR #96, opened 2019-01-30
Add `required` property on Field

### LINKED ISSUE #88: Allow to know whether a field is required
At the moment, there is no way to know whether a field is required (e.g. cannot left empty when creating a model instance) on a `Model`.

Alter the behavior of `default`: if not provided to the `Field`'s constructor, it should be set to a sentinel value, e.g. `DEFAULT_NOT_PROVIDED`. Then, we can have `required` (or `blank`, to follow the Django API?) be a property computed as `return self.default is not DEFAULT_NOT_PROVIDED`.

The API would look like this:

```python
from tortoise import fields

class BlogPost(Model):
    title = fields.CharField(max_length=100)
    description = fields.TextField(default=None, null=True)

assert BlogPost._meta.fields_map["title"].blank
assert not BlogPost._meta.fields_map["description"].blank
```

Add a `blank` parameter to `Field`'s `__init__()` and simply store it on the field. However, we don't have any validation mechanism for now so we wouldn't be able to act upon `blank` which may be counter-intuitive. I suppose for now it's better to compute it from whether a `default` was passed.

**Additional context**
I'm playing around with building Django REST Framework-like serialization on top of Tortoise, with Pydantic as a validation backend. My approach is to dynamically build a Pydantic model out of a Tortoise `Model`. Pydantic has built-in validation for required/non-required fields, so knowing whether a field is required is a blocker for now.

Also — I am willing to work on this issue. 😄  I'd be glad to have pointers at where I should be careful when altering the behavior of `default` (e.g. assigning to a sentinel value), although from what I've seen it should be quite easy to figure out.

florimondmanca:
## Description
Add a `required` read-only property in the base `Field` subclass.
The logic is: a field is required if no `default` was given. This is implemented using a sentinel object. 

Currently, `required` won't update if the `default` changes, but I made the assumption that field attributes are not meant to be modified. If that's an issue I can fix this with a `setter` for `default`.

## Motivation and Context
Fixes #88 

## How Has This Been Tested?
I added a `TestRequired` class for various cases.

## Checklist:


florimondmanca:
Yes, I didn't think about that edge case which means the solution you gave is more correct.

grigi:
Ha! that was fast!
Waiting for builds...

PR #98, opened 2019-02-01
Fix/require capability connection

florimondmanca:
## Description
Add a `connection_name` parameter to `@requireCapability()`.

## Motivation and Context
The connection name which the database client is retrieved from was hardcoded to `"models"`, but I ran into the issue that schema generation tests use `"default"` instead. Generally it's better to make this configurable with `"models"` being a sensible default.

Blocking #89 

## How Has This Been Tested?
Added a test that is expected to fail if the provided `connection_name` refers to a non-existing connection.

## Checklist:

### DIFF
--- a/tortoise/tests/test_capabilities.py
+++ b/tortoise/tests/test_capabilities.py
@@ -20,6 +20,12 @@ class TestCapabilities(test.TestCase):
         with self.assertRaises(AttributeError):
             self.caps.dialect = 'foo'
 
+    @test.expectedFailure
+    @test.requireCapability(connection_name='other')
+    def test_connection_name(self):
+        # Will fail with a `KeyError` since the connection `"other"` does not exist.
+        pass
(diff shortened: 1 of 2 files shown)


tortoise/tests/test_capabilities.py
```diff
@@ -20,6 +20,12 @@ def test_immutability_1(self):
         with self.assertRaises(AttributeError):
             self.caps.dialect = 'foo'
 
+    @test.expectedFailure
+    @test.requireCapability(connection_name='other')
+    def test_connection_name(self):
+        # Will fail with a `KeyError` since the connection `"other"` does not exist.
+        pass
```
grigi: :+1:

grigi:
Thanks for this. In hindsight it seems obvious that we should have had a better way of choosing the "connection" to check.

PR #101, opened 2019-02-04
Added comprehensive unit tests for ForeignKey and ManyToMany fields

grigi:
* Tests for relation fail cases
* Tests for FK create/access
* Tests for FK uninstantiated conditions
* Tests for FK unfetched conditions
* Tests for FK queryset manipulation
* Tests for M2M create/access/delete/clear
* Tests for M2M uninstantiated conditions

In the process found a few unexpected behaviours and bugs:
* [x] FIXED: Inconsistent lack of `await` allowed on references
* [x] FIXED: `.all()` actually returns the `_query` property as was documented.
* [x] CHANGED: New models with FK don't automatically fail to resolve any data. Done this due to the behaviour seeming non-obvious.
* [x] FIXED [CAPS]: Some DB's don't support OFFSET without Limit, added caps to signal workaround, which is to automatically add limit of 1000000
* [x] FIXED: Pylint plugin to know about default `related_name` for ForeignKey fields.
* [x] CHANGED: Simplified capabilities to be static, and defined at class level.
* [x] REMOVED: Removed `distinct()` as it could only happen after a complicated queryset, it made no sense by itself there.

Doc work:
* [x] Update docs for Foreign key/m2m uses (and there are many)
* [x] Add `Capabilities` to docs.
* [x] DB driver parameter docs.
* [x] Updated CHANGELOG


grigi:
@abondar I'm considering removing the sync portion from relationships.
Currently we have this weird and odd API:
```py
tour = await testmodels.Tournament.create(name='Team1')
rel = await testmodels.MinRelation.create(tournament=tour)
await tour.fetch_related('minrelations')
print(tour.minrelations[0])
```
Whereas I feel this would be more consistent to use:
```py
tour = await testmodels.Tournament.create(name='Team1')
rel = await testmodels.MinRelation.create(tournament=tour)
print(await tour.minrelations[0])
```
But to do that we need to sacrifice the entire sync portion on relationships, like `__contains__`, `__iter__`, `__len__`, `__bool__` as the mix of sync and async here does confuse me.
I also feel that `fetch_related` falls under optimization, and right now it gets in the way of a clean API, and the cacheing it does is unexpected, and will be a source of bugs in code.

Did you have a specific idea in mind of the sync interface?
Can I break the API?

abondar:
```tour = await testmodels.Tournament.create(name='Team1')
rel = await testmodels.MinRelation.create(tournament=tour)
print(await tour.minrelations[0])
```

I don't quite get how that would be implemented and what API it will have.

`await tour.minrelations[1]` would generate query with `offset=1 limit=1`?

Right now I am opposed of tearing down sync API, it was added with one important purpose:
Consider you have your web service where you have some kind of serialisation (e.g. marshmallow)
With this sync API you can models like this:
```
class Parent(models.Model):
    name = fields.TextField()
    

class Child(models.Model):
    parent = fields.ForeignKeyField('models.Parent', related_name='childs')
    name = fields.TextField()
```

And you want to serialise them into such structure:
```
{
    'name': 'parent',
    'childs': [
        {
            'name': 'child1',
        },
    {
            'name': 'child2',
        },
    ]
}
```

With current API you just write such schemas:
```
class ChildSchema(Schema):
    name = fields.String()


class ParentSchema(Schema):
    name = fields.String()
    childs = fields.Nested(ChildSchema, many=True)
```

And have serialisation done like this:
```
await parent.fetch_related('childs')  # it could be queried with childs from start await Parent.filter(id=1).first().prefetch_related('childs')
res, err = ParentSchema().dump(parent)
```

If we remove sync methods users won't be able to use Tortoise with existing serialisation libs, which I think is big loss.

grigi:
`If we remove sync methods users won't be able to use Tortoise with existing serialisation libs, which I think is big loss.`
Thanks for that info. That is why I asked.

We'll just have to put big warnings around the usage of `RelationQueryContainer` so that we make it clear when one should use what method.

grigi:
@abondar Please review.

PR #103, opened 2019-02-11
Differentiate between timezone aware and timezone naive datetimes

arlyon:
SQLite supports any valid ISO8601 datetime string when defining dates (amongst others) which may or may not include the timezone. This means that a field may have a mixture of timezone aware and timezone naive values. Postgres differentiates between timezone aware and timezone naive fields and so code that is valid when using the ORM with sqlite may cause errors in other engines.

Ideally, the DatetimeField, since it is not timezone aware, would catch any timezone aware datetimes and provide a descriptive error instead trying (and failing) to insert them. Another option is to provide the desired timezone that the datetime should be stored in, so that the field can convert a timezone-aware datetime into the timezone-naive datetime for the supplied timezone.

```python
class Event(Model):
    ...
    start_time = fields.DatetimeField(in_timezone=timezone.utc)
```

In the example above, any timezone-aware datetime is converted to a timezone-naive utc time before it is saved. If the timezone is supplied, datetimes retrieved from the model are timezone aware.

Is this reasonable? Or should we remove the `in_timezone` flag and create a separate timezone aware field? For some context, I am receiving strings that may or may not be timezone-aware and it is a pain having to manually guard and convert them over. So if I were able to define on the model that all datetimes once they hit the database are guaranteed UTC then it would vastly simplify the logic.

### DIFF
--- a/tortoise/fields.py
+++ b/tortoise/fields.py
@@ -1,6 +1,7 @@
 import datetime
 import functools
 import json
+from datetime import timezone
 from decimal import Decimal
 from typing import Any, Optional, Union
 
@@ -179,20 +180,35 @@ class DatetimeField(Field):
         Always set to ``datetime.utcnow()`` on save.
     ``auto_now_add`` (bool):
         Set to ``datetime.utcnow()`` on first save only.
+    ``in_timezone`` (timezone):
+        Convert all supplied timezone aware datetimes to
+        to this timezone before insertion in the database.
     """
-    __slots__ = ('auto_now', 'auto_now_add')
+    __slots__ = ('auto_now', 'auto_now_add', 'in_timezone')
 
-    def __init__(self, auto_now: bool = False, auto_now_add: bool = False, **kwargs) -> None:
+    def __init__(
+            self,
+            auto_now: bool = False, auto_now_add: bool = False,
+            in_timezone: Optional[timezone] = None, **kwargs
+    ) -> None:
         if auto_now_add and auto_now:
             raise ConfigurationError("You can choose only 'auto_now' or 'auto_now_add'")
         super().__init__(datetime.datetime, **kwargs)
         self.auto_now = auto_now
         self.auto_now_add = auto_now | auto_now_add
+        self.in_timezone = in_timezone
 
-    def to_python_value(self, value: Any) -> datetime.datetime:
-        if value is None or isinstance(value, self.type):
+    def to_python_value(self, value: Any) -> Optional[datetime.datetime]:
+        if value is None:
             return value
-        return ciso8601.parse_datetime(value)
+
+        if not isinstance(value, self.type):
+            value = ciso8601.parse_datetime(value)
+
+        if self.in_timezone is not None:
+            value = value.replace(tzinfo=self.in_timezone)
+
+        return value
 
     def to_db_value(self, value: datetime.datetime, instance) -> datetime.datetime:
         if self.auto_now:
@@ -203,6 +219,11 @@ class DatetimeField(Field):
             value = datetime.datetime.utcnow()
             setattr(instance, self.model_field_name, value)
             return value
+        if hasattr(value, "tzinfo") and value.tzinfo is not None:
+            if self.in_timezone is not None:
+                value = value.astimezone(self.in_timezone).replace(tzinfo=None)
+            else:
+                raise ValueError(f"Timezone aware datetime {value} supplied to DatetimeField")
         return value
 
 
--- a/tortoise/fields.py
+++ b/tortoise/fields.py
@@ -223,7 +223,9 @@ class DatetimeField(Field):
             if self.in_timezone is not None:
                 value = value.astimezone(self.in_timezone).replace(tzinfo=None)
             else:
-                raise ValueError(f"Timezone aware datetime {value} supplied to DatetimeField")
+                raise ValueError(
+                    "Timezone aware datetime {} supplied to DatetimeField".format(value)
+                )
         return value
 
 
--- a/tortoise/tests/test_fields.py
+++ b/tortoise/tests/test_fields.py
@@ -1,4 +1,4 @@
-from datetime import date, datetime, timedelta
+from datetime import date, datetime, timedelta, timezone
 from decimal import Decimal
 from time import sleep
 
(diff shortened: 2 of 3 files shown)


grigi:
I think it is right to set timezone on the `DateTimeField` as a parameter.

We need to handle cases where the timezone of the DB is different to the timezone of the server. Or one of them is naive.
E.g. the legacy MySQL db at work has a timezone of SAST, but I need to work with UTC fields in the API.
So we may need to specify both?

In-DB it would be either TZ-aware field, or stored naive as «some timezone».
We then should then have another flag set if we want tz-aware or naive in Python (current timezone)?
Then I would default it to TZ-aware for both.

How does that sound?

arlyon:
I can see 4 use cases essentially:

- I have a timezone-naive field with timezone-naive datetimes. Everything is as it was before.

- I have a timezone-naive field with potentially aware datetimes. We specify that tortoise should store all timezone-aware datetimes as "X" timezone using the `python_timezone` parameter. All timezone-naive datetimes are assumed to be in that timezone. Tortoise returns timezone-aware datetimes in X timezone.

- I want to interact with a timezone-aware database. I set the `database_timezone` parameter to the appropriate timezone Y. This implicitly sets the `python_timezone` to the same value and any naive datetimes are assumed to be from that timezone. Tortoise returns timezone-aware datetimes in Y timezone.

- I want to interact with a timezone-aware database in a different timezone that my own. I set both `database_timezone=X` and `python_timezone=Y` as needed. At this point, any naive datetimes going into the database are assumed to be in `python_timezone` and all datetimes tortoise returns will be timezone-aware in the `database_timezone`.

You'll notice that in all but the first of these cases the database returns timezone-aware datetimes. I think this makes the most sense. In the first two cases, the field in the database would be a `timestamp` and in the last two a `timestamptz`.

I do think this risks over-complicating things. Maybe an additional field would be better? It's not exactly intuitive to understand what each of those parameters does.

grigi:
It is a very thorough motivation of the use cases.

Case 1 is naive, and should probably not be used.
Case 2 is very specialised, and a way of using a legacy/erroneous Case 1.
Case 3 should be the default, or should it be Case 4? What is really the difference?
I would normally want to do everything as UTC in the DB, and then be bothered with "local" timezone only for presentation. So I don't see much benefit for Case 4... 

So, should case 1 actually ever be used? Or could we consider it a special case of Case 2? (a timezone of `None`?)

Then, should this be separate fields, or one with special options?
If we drop Case 4, and consider both Case 1 & 2 to be very specialised, then we should consider their use only under `ADVANCED USAGE: here be dragons`?
And since Case 2 is so specialised, Should we not just consider it a manual specialisation of Case 1?

Or am I talking in circles here?

grigi:
Definitely talking in circles.

At root we need a way to specify DB-level timezone (or absense) and ensure that all datetimes are converted to that timezone before saving?
And default to a timezone of `UTC`.

Would that solve your problem? Because maybe we should think about limiting feature creep, as we are starting to feel it, and we are not even at version 1.0 yet :stuck_out_tongue:

abondar:
I think we should just switch to timezone aware datetimes and add option to config, that would be responsible for current timezone.

I don't think that we would want to change DB settings changing it's timezone, so we could convert from db timezone to config timezone in `to_python_value` and making naive datetime aware in `to_db_value`.

Also, implementing warnings about naive datetimes and some utils for users and us to ease working with timezones would be nice, since default python workflow with timezones is quite lacking and not very straightforward.

arlyon:
Django handles timezones similarly, with a flag in settings.py `USE_TZ` which stores everything in UTC in the database and is then converted to the appropriate timezone. I don't see any issue with defaulting the database to UTC for example and then like you said providing an override in the config. Are there any databases that _don't_ support timezones?

arlyon:
Alright I'll change the impl. to use timezone aware datetimes (defaulting to UTC) and allow an override in the config.

grigi:
Thank you Alexander, looking forward to having this :-)
If you need any help, just ask and I'll make some time.

arlyon:
Finally got some time to take a crack at this. Needs a little bit of work, as mysql sets the timezone on a _per-connection_ basis, storing everything in utc meaning multiple timezones in the db isn't possible while postgres and sqlite handle it just fine. 

Solution is to explicitly set each mysql connection to UTC, and then insert "naive" (utc) datetimes for mysql 

Postgres and sqlite were pretty straight forward though!

grigi:
Wow thats frustrating.
I realized we are doing more an more odd workarounds for MySQL than the other two RDBs added up together.

arlyon:
Going to close until I have time to take another shot at this

grigi:
Hi, could you write a small blurb with your learnings? Did you run out of time to work on this?
Just curious.

arlyon:
No problem. Finding a "nice" way to do it was difficult, and university got quite hectic and stayed that way up until exams last week. I'm starting a job over the summer this week which (once I get settled in) should free up some time in the evenings. I'm still sitting on the work for the geospatial database too that I'd like to merge in the coming weeks but I figured there's no use me "claiming" this feature if I can't get it done in a timely manner.

The main problem I kept running into was just finding a way that felt logical and that worked reliably over all databases without compromising flexibility. I would try some idea and always find something that didn't quite work primarily because of the issue mentioned above: MySQL doesn't handle timezone aware stuff very nicely compared to the other backends and needs per-connection config. I may be missing something that someone with more experience may catch but my experience with the various databases is average at best.

grigi:
Hope your exams went well :-)

Yes, MySQL and "nice" can get in the way easily. Generally the rule of thumb with feature "emulation" is to do a best effort, and document the shortcomings.

My tirade:
If MySQL stored datetimes as naive, and then converts them to/from datetime based on some DB param that is set on connection, then that is how we should do it, as in set the db_timezone as a connection param (and have a reasonable default, maybe use the local systems timezone as the default).
Then we just try and override the to/from db converters for MySQL and datetimes to be converted depending on parameters, etc...
And document the emulation as a gotcha in the MySQL section?

I'm happy to sacrifice some flexbility in the name of simplicity. We have 2 cases we must support:
* Timezone-aware (default)
* Timezone-naïve

Rationale is that the user-dev could easily transform these themselves if they want to do something very specific, and it is a rare edge-case.

We can then later on support the other cases if they are really required.

Are you OK if I tackle a basic timezone implementation? v0.12 is already doing large changes with how PK's are done, it would make sense to put two breaking changes in at the same time.

jorgeecardona:
Why not just simply add a field `DateTimeTZ` that works for SQLite and PostgreSQL but not in MySQL?

PR #114, opened 2019-03-20
Fix 'unique_together' for foreign keys

REVIMI:
## Description
Fixed 'unique_together' for foreign keys

## Motivation and Context
Adding 'unique_together' with FK field name was broken, there was only an ability to add it using a field name like '{foreign_key_name}_id'

### DIFF
--- a/tortoise/backends/base/schema_generator.py
+++ b/tortoise/backends/base/schema_generator.py
@@ -123,12 +123,23 @@ class BaseSchemaGenerator:
             if field_object.index:
                 fields_with_index.append(field_name)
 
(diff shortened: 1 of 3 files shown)


grigi:
Thank you, on retrospect this should have been an obvious thing to do.

Will this work with M2M fields, or does that not make sense? :thinking:

REVIMI:
M2M relation is not presented as a column in DB table, so it makes no sense. As I implemented it, unique together on M2M field won't pass the validation.

grigi:
Yes, perfectly right. Thanks for the work

Issue #113, opened 2019-03-20
OperationalError after retrieving row from db by id, manually assigning the value from IntEnum to the relevant attribute of the model instance and calling `save()` on it.


croshchupkin:
Basically, subj. To repro, please use the following code:
```
# models.py
from enum import IntEnum

from tortoise import Model, fields


class ContactTypeEnum(IntEnum):
    home = 1
    work = 2
    other = 3


class Contact(Model):
    id = fields.IntField(pk=True)
    phone_no = fields.CharField(30, null=False, default='')
    type = fields.IntField(required=True, default=ContactTypeEnum.other)
```
Then run the script:
```
from tortoise import run_async, Tortoise

from models import Contact, ContactTypeEnum


async def run():
    await Tortoise.init(
        db_url='sqlite://db.sqlite3',
        modules={'models': ['models']}
    )
    await Tortoise.generate_schemas()

    await Contact.create(phone_no='222', type=ContactTypeEnum.work)

    breakpoint()
    contact = await Contact.get(id=1)
    for name, value in {'phone_no': '1111', 'type': ContactTypeEnum.other}.items():
        setattr(contact, name, value)
    await contact.save()


if __name__ == '__main__':
    run_async(run())
```
I get the following exception: tortoise.exceptions.OperationalError: no such column: ContactTypeEnum.other. But before the update, the new model instance was saved successfully.

croshchupkin:
After having stepped through the code for a bit, it seems, that pypika constructs the following SQL query: ```UPDATE "contact" SET "phone_no"='1111',"type"=ContactTypeEnum.other```

croshchupkin:
I assume that the query will work fine if I do int(ContactTypeEnum.other) - but I just find it strange that the entity is created without issue while the update has problems.

croshchupkin:
After digging around some more, I think this is most likely not a bug, but a "feature" of Enum's default `__str__`. This can be tweaked by creating the enum like so:
```
class ContactTypeEnum(IntEnum):
    home = 1
    work = 2
    other = 3

    def __str__(self):
        return str(self._value_)
```

grigi:
Creates are done using parametrized queries, updates/selects are not yet.
related to #81

grigi:
Could you add this as some failing tests that are marked with `@test.expectedFailure` that does insert/update and filter on an enumfield?
The same feature also blocks dealing with binary blobs.

croshchupkin:
[...]

PR #116, opened 2019-03-22
Fixed Field.to_db_value method (#113)

REVIMI:
## Description
Fixed Field.to_db_value method
Added converting field's value to the field's type

## Motivation and Context
As for now, using inappropriate field type leads to hardly readable OperationalError.

As a side effect, setting enum as a value for IntField will be fixed. (#113)

### DIFF
--- a/tortoise/fields.py
+++ b/tortoise/fields.py
@@ -53,7 +53,9 @@ class Field:
         self.model_field_name = ''  # Type: str
 
     def to_db_value(self, value: Any, instance) -> Any:
-        return value
+        if value is None or type(value) == self.type:
+            return value
+        return self.type(value)
 
     def to_python_value(self, value: Any) -> Any:
         if value is None or isinstance(value, self.type):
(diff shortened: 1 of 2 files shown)


REVIMI:
@grigi my code didn't pass the check:
Using type() instead of isinstance() for a typecheck.

But actually type() and isinstance() are not the same. For the particular case with Enum attributes:
```
isinstance(ContactTypeEnum.work, int)  # returns True
```
though such Enum field can't be straight used as an integer.

Using
```
type(ContactTypeEnum.work) == int  # returns False
```
is a workaround for enums.
That is rather an enum package bug.
@grigi what do you think? Should we have in mind that users are using Enums in the way @croshchupkin does?

grigi:
I think that using enums is very much normal. I do feel one would rather want to do it like Django and pass the enum you are using in to the field on definition, and then it would cast to/from that enum on load/save.

I'm in two minds if this is the right solution for the problem:
1) It will "work"
2) It won't cast back to the enum directly.

The codacy issue seems to be related to PyLint, so you can add a comment of `  # pylint: disable=«error-codes»` if you just want to silence it.

REVIMI:
I think that the value of the IntField doesn't have to automatically cast back to the enum. Getting int from IntField is what is expected and it will be confusing to get anything else.

PR #120, opened 2019-04-02
Black formatter integration

abondar:
Hey @grigi 
I tried running `black` formatter against code and I already liked results even with default settings.

What do you think of it?

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -5,4 +5,4 @@ multi_line_output=3
 include_trailing_comma=True
 force_grid_wrap=0
 use_parentheses=True
-line_length=88
+line_length=100
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,5 +1,8 @@
 [settings]
-line_length=100
-multi_line_output=0
 known_third_party=aiosqlite,ciso8601
 not_skip=__init__.py
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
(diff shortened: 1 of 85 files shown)


grigi:
Could I ask that you keep the line length of 100? The 88 seems quite narrow when I look at the code.
And explicitly configure target version to py35 as that is our minimum supported version.

The only real concern is how much I personally dislike double-quotes. I know it is recommended by everyone, but it just doesn't read well at all for me. I am curious how the diff changes when `--skip-string-normalization` is used on the original sources.

grigi:
Why did you skip on Py35?

abondar:
Black does not support running from python earlier than 3.6, so I thought it would be fine, as long as we have whole matrix, to test formatting only on higher versions.

Regarding double quotes - while writing I am also usually use single quotes, but mostly because it's easier to type it so :) 
But [paragraph](https://github.com/ambv/black#strings) on black readme was enough to make me give up on it in this case. I think as public piece of code Tortoise should stick more to public conventions and recommendations.
Also, good thing about black, that you don't have to write them as double-quotes, you just have to call formatter before commit and you are all good.

grigi:
I spent a bit of time to go over the formatting changes, and generally they are pretty good.
I did notice that about ⅔ of the line diffs are just changing the quotes, so in general we kept a tight enough ship.

Please remove the references to `yapf` and you are good to go :+1:

Issue #119, opened 2019-04-02
Trying to run tortoise with quart  async rest api


shaloba:
**Describe the bug**
Hi, 

I have simple tortoise example that work fine when i create the python event loop and run it.
but when i run Quart API server I get strange error, I braked my head trying to figure out what is the issue 
(I'm running python3.7)
The script I ran is below: 
```
import asyncio
from tortoise import Tortoise, fields
from tortoise.models import Model

from quart import Quart, request, abort

app = Quart(__name__)
app.config.from_object(config)

class Users(Model):
    id = fields.IntField(pk=True)
    status = fields.CharField(20)
    def __repr__(self):
        return str(self.id)

class Workers(Model):
    id = fields.IntField(pk=True)
    status = fields.CharField(20)
    def __repr__(self):
        return str(self.id)

@app.before_first_request
async def initial_db():
    await Tortoise.init({
        'connections': {
            'default':  "postgres://localhost:5432/db",
            'second':  "postgres://localhost:5432/db",
        },
        'apps': {
            'users': {
                'models': ['__main__'],
                'default_connection': 'default',
            },
            'workers': {
                'models': ['__main__'],
                'default_connection': 'second',
            }
        }
    })


def run_quart():
    app.run(port=5002)


def run_tortoise():
    loop = asyncio.get_event_loop()
    loop.run_until_complete(initial_db())

if __name__ == '__main__':
    # run_quart() # not working
    run_tortoise() # working
```

**To Reproduce**
run one time the "run_quart" see the error
then run "run_tortoise" see the run succeed

**Expected behavior**
tortoise ORM will use the Quart event loop

**Actual behavior**
Tortoise raise "NotADirectoryError" error

**Additional context**
This is my first time using Tortoise ORM and its looking very cool, I will appreciate any help with this issue 

Thanks in advanced!

grigi:
I have not used Quart before, I know with aiohttp you have a init function where you can add async initialisers to use. It also contained the object state in the app, so you had no other globals.

That code looks like it should work, I will have to try it locally.

grigi:
Here is a working sample app. (It probably is un-quark-ish as it is my first Quark app)
The info you were looking for is: https://pgjones.gitlab.io/quart/startup_shutdown.html
as https://tortoise-orm.readthedocs.io/en/latest/setup.html says one needs to set-up and tear-down the ORM properly.

I hope this helps :smile: 

```py
import asyncio
import json
from random import choice

from quart import Quart

[...]

PR #121, opened 2019-04-02
[WIP] Sample Quart integration

grigi:
Sample Quart integration

* [x] Build a more sensible example
* [x] Build a small plugin that does tear-down automatically
* [x] Add it to documentation
* [x] Add CLI command `generate-schemas`
* [ ] Ensure we solve #119


examples/quart_example.py
```diff
@@ -0,0 +1,66 @@
...
+
+@app.after_serving
+async def close_orm():
+    await Tortoise.close_connections()
+
+
+@app.route("/")
+async def list_all():
+    users, workers = await asyncio.gather(Users.all(), Workers.all())
+    return json.dumps(
```
pgjones: Quart has a `jsonify` function that should be used instead, as it sets the content-type header to json.

pgjones:
It tortoise asyncio only or can Trio be used? If the latter is possible it might be nice to provide an example that could work with either (quart-trio also exists).

Thanks for adding a Quart example 🎉

examples/quart_example.py
grigi: Thanks, will change.

grigi:
I doubt Trio will work anytime soon. We use asyncio directly in a few places, and I suspect we may run into issues with the DB drivers.

grigi:
@pgjones Could you please re-review, specifically look at: `examples/quart/` and `tortoise/contrib/quart/`
https://github.com/tortoise/tortoise-orm/pull/121/files#diff-059e35c4f26307fa382f487d0388a432
https://github.com/tortoise/tortoise-orm/pull/121/files#diff-f719a1e73bcf968f213cff7a2c990ce3
https://github.com/tortoise/tortoise-orm/pull/121/files#diff-7326eaad88442bfed6b2cbc3e7a3a986

Thanks in advance

pgjones:
👍 Looks good to me.

Issue #97, opened 2019-01-31
Field's schema customising


abondar:
Currently information on how to generate schema from field is stored in `schema_generator` and in order to make custom field for some DB type, that is not covered by us, user would have to create custom  SchemaGenerator and DBClient pointing to it, which is quite hard deal.

We could simplify SchemaGenerator by delegating description of schema to models and fields themselves. So fields would have method `get_sql_type` where it would return just string that would go in schema generating sql. Also there is advantage with this approach, that this way fields like `DecimalField` would be generated much more elegantly, than it does now.

Problem with that approach is how to alter that string depending on dialect. I think we could pass dialect (or capabilities) down to `get_sql_type` functions, so that they can alter behaviour based on dialect.

Also I think it should be developed alongside with #39

grigi:
We should be able to override the default Field type for a db type, and replace it transparently if provided.

E.g. Using the native functionality of PostgreSQL for JSON and UUID:
```py
from typing import Optional, Union
from uuid import UUID

from tortoise.fields import JSONField as GenericJSONField
from tortoise.fields import UUIDField as GenericUUIDField


class JSONField(GenericJSONField):
    def to_db_value(self, value: Optional[Union[dict, list]], instance):
        return value

    def to_python_value(self, value: Optional[Union[str, dict, list]]):
        return value


class UUIDField(GenericUUIDField):
    def to_db_value(self, value: Optional[UUID], instance):
        return value

    def to_python_value(self, value: Optional[UUID]):
        return value
```

grigi:
Related issue #55

grigi:
FWIW, currently all the DB-specific field logic sits in the executor (and DDL in schema_generator). There is a few places in the code I noticed where we DON'T use the executor. We need to move it all in.
Also, we currently re-create the executor for every operation, and I have a cacheing system to make construction cheaper. If we make it required and created on first use of DB-MODEL combo to support complicated Field-DB logic, we can then avoid the cacheing infrastructure for a simpler solution.

One of the ideas could be:
when one defines a Field, one could define many getters/setters and field info depending on dialects?

e.g:
```py
class CharField(Field):
    __slots__ = ("max_length", "case_sensitive")
    DB_NATIVE_TYPE = True

    def __init__(self, max_length: int, case_sensitive=True, **kwargs) -> None:
        if int(max_length) < 1:
            raise ConfigurationError("'max_length' must be >= 1")
        self.max_length = int(max_length)
[...]

PR #122, opened 2019-04-03
[WIP] Generic primary key implementation

### LINKED ISSUE #36: Generic PK field that references real PK field
Currently tortoise can work normally if your PK is int field named `id`. It should instead work with `pk` field like in django, that references real PK field that can have various types and names

abondar:
Closes #36 

Hey @grigi 
I wrote down this implementation and, surprisingly, it's even works (on sqlite).
I'll try to do cleanup on it in following days and write some more tests, but could you review it by then?
This PR carries many changes and I am afraid there could be some places which I could have done better, but missed them in all this work.

I hope you with fresh view will be able to propose some changes for the better here


grigi:
Wow, this PR does a lot. Adds new datatypes even, and pg-specific optimisations, and makes uuid PK-able :-)
Can you also make CharField PK-able too?

abondar:
Well, it's already should be pk-able, like any hashable value field. It's just that I was already beaten on last line of code and decided to write tests for other field types later

tortoise/backends/sqlite/client.py
```diff
@@ -90,7 +95,10 @@ def _in_transaction(self) -> "TransactionWrapper":
     async def execute_query(self, query: str) -> List[dict]:
         async with self.acquire_connection() as connection:
             self.log.debug(query)
-            return [dict(row) for row in await connection.execute_fetchall(query)]
+            print(query)
+            res = [dict(row) for row in await connection.execute_fetchall(query)]
+            print(res)
```
grigi: remove prints when you are done.

tortoise/fields.py
```diff
@@ -303,6 +306,28 @@ def to_python_value(
...
+class UUIDField(Field):
+    """
+    UUID Field
+
+    This field can store uuid value. Postgresql will store value with
+    native data type, while others will store it as string
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(type=UUID, *args, **kwargs)
```
grigi: Should generated not also be auto-set when pk=True?

grigi:
@abondar Do you need help in completing this? It has been sitting around for four weeks now?

abondar:
Hi, sorry for dragging this for so long, couldn't find time to finish it and now I am on vacation until middle of may. 
If you have courage and time I would be happy if you will finish it, otherwise I am going to finish it after coming back

grigi:
Cool, enjoy your vacation :smile: 
I'll have a crack at some of it, partly as I want to understand everything here :+1:

grigi:
@abondar I rebased on latest master, fixed the linting/typing related issues, fixed the M2M table creation to use the right FK key type, and fixed postgres to work again.

I see there is a new attribute `fetch_inserted` that seems like it should control wether we fetch the inserted values, but I wonder why this is an attribute, as we should do the most optimal variety of SQL for insert and getting generated value?

I also see you are trying to replace the field classes for postgres, so we can use the native types easier, I didn't try and implement an auto-replace yet.

And, we need quite a few more tests and lots of documentation changes.

grigi:
Could we just finish the test cases for primary keys, and then merge this without the replacing field classes on a per-db basis? This PR does enough and I feel the field refactoring should be a separate issue? Because it is quite large as well.

grigi:
Turns out both MySQL and SQLite handles ROWID primary key tables and non-ROWID tables very differently.
Also the current implementation will force a ROWID on those tables, and then always collect it afterwards.
This is OK, but causes problems for PostreSQL because it can generate other types on the DB, e.g. UUID. What makes it better right now is that we don't support those cases.

This is one of the many things that is going to complicate the field refactoring #97

grigi:
@abondar Please review my changes.

I removed the custom PostreSQL uuid field type, as until we have parametrized queries for everything, we are likely to run into escaping issues when the to_db_value doesn't treat it as a string. PostgreSQL seems smart enough to parse UUIDs so for now it is a working limitation.

Also, have not updated the documentation.

From a code POV I'm happy with what we have here now.

Also, can we discuss what is to happen for v0.12.0 ?

abondar:
Thank you!
Changes looks all good to me, so I think the only concern left is documentation updates. 
If you don't have opportunity to write it, I think I'll try to manage some time on next week for it.

Regarding 0.12.0 I think this PR is big enough to be sole member of it.

grigi:
?? That last commit broke stuff spectacularly in Travis, but I don't know why. as it doesn't happen locally.
In the MySQL test runs it is emitting PG sql, so I suspect some sort of resource leak?

grigi:
Ok, that looks much better. 
Time has been very short on my side, but it appears to be getting a bit better.

grigi:
@abondar Please vet the docs

docs/models.rst
```diff
@@ -65,20 +65,35 @@ Every model should be derived from base model. You also can derive from your own
...
-Sadly, currently only simple integer primary key is supported, there is plans to enhance this in the not too distant future.
+Any of these are valid primary key definitions in a Model:
+
+.. code-block:: python3
+
+    id = fields.IntField(pk=True)
+
+    checksum = fields.CharField(pk=True)
+
+    guid = fields.UUIDField(pk=True)
```
abondar: I think we can also mention, that filtering by `.filter(pk__in=...)` or whatever other filters is also supported

PR #123, opened 2019-04-04
Converge and update config files

grigi:
* Merged `.isort.cfg`, `.green` and the `flake8` section in `tox.ini` into a `setup.cfg`. Also added the `mypy` config to it.
* Updated test/dev dependencies and tried to make `black` install depending on `>=py3.6` through `pip`.
* Cleaned up the makefile a bit.
* Updated the "target python" for Black to py3.5, and it did a few minor changes.
* Deleted `.style.yapf`.
* Tuned up the strictness of `mypy` a little.
* `sphinx 2.0` causes rendering issue with `cloud_sptheme`. Limit to version<2 for now.
* Added logs when Tortoise-ORM starts up and tears down.


grigi:
@abondar Just some small updates/changes. Please review.

grigi:
Yes, we have both been doing squash and merge for a while now :-)

Issue #85, opened 2019-01-13
How do I add functionality to all of my models without repeating myself?


monospacedmagic:
I want to add a `_d` attribute and a few `classmethod`s and some regular methods to models by creating a subclass of `tortoise.models.Model` that users of my framework can use. However, subclassing `Model` would create a new database model, which is not what I want. From what I've seen, the `Model` class contains a lot of logic, and it seems not feasible to create a new class with `metaclass=ModelMeta`. What are my options here? Am I missing something?

abondar:
Hi

There is `abstract=True` option for models meta, it's mentioned there https://tortoise-orm.readthedocs.io/en/latest/models.html

Is it what are you looking for?

monospacedmagic:
Ohhh! Thanks, I somehow completely skipped that. It's exactly what I was looking for. :)

tcmal:
When using `abstract=True`, it seems like the child classes don't inherit fields, ie:

```
class Entity(Model):
    x = fields.IntField()
    y = fields.IntField()

class ChestEntity(Entity, Model):
    level = fields.IntField()
    opened = fields.BooleanField()
```

Just leaves the chestentity table as id, level & opened.

Am I misunderstanding this feature?

grigi:
Try just inheriting from `Entity` instead of both `Enitity` and `Model`?

grigi:
I realised we don't have a detailed example in the docs for this, nor is it part of the test suite... :disappointed:

tcmal:
Still doesn't seem to work.

grigi:
@tcmal Can you please check with release v0.12.4 ?
new docs: https://tortoise-orm.readthedocs.io/en/latest/models.html#inheritence

PR #146, opened 2019-06-30
Inheritence for fields

FirstKlaas:
Now also class attributes of type Field of super classes (and the whole mro) are taken into account when building the models.

## Description
When defining models, only the class attributes of the same class where used. No attributes of any super class are added. This was due to the design of the meta class in Model. I added a scan for every base class and for every base class all classes in the mro.

So @grigi, what do you think?

Now you can do something like this:

```
class IDMixin():
    id = IntField(pk=True)
    name = TextField()
    created_at = DatetimeField(null=True, auto_now_add=True)
    modified_at = DatetimeField(null=True, auto_now=True)

class NamedMixin():
    name = TextField()

class User(IDMixin, NamedMixin, Model):
    pass
```

## Motivation and Context
This is more an improvement. Cause you could achieve the same result by retying the same field to every model class. But now you can leverage from (multiple) inheritance.

## How Has This Been Tested?
I tested it currently in another project. I haven't written any tests so far. If the idea and the implementation approach has been approved, I will also update the documentation and add some tests.
 
## Checklist:
- [X] My change requires a change to the documentation.

### DIFF
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -15,6 +15,7 @@ from tortoise.fields import (
 from tortoise.filters import get_filters_for_field
 from tortoise.queryset import QuerySet
 from tortoise.transactions import current_transaction_map
+from typing import Iterable
 
 MODEL_TYPE = TypeVar("MODEL_TYPE", bound="Model")
 # TODO: Define Filter type object. Possibly tuple?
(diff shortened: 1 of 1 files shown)


grigi:
I see, this would be a simpler "abstract" like feature? 
Both this and the existing "abstract" features have no tests, and I would like to guarantee that stays working in the future use case.
We should ensure that this works with the existing "abstract" feature. The reason why this cannot entierly replace the "abstract" model class, is that it allows one to override some of the standard `Model` methods.

See the current discussion in #85 re how "abstract" is currently broken.

FirstKlaas:
It's no an replacement of the abstract feature. But the implementation also uses the fields of an abstract (or none abstract) super class.

FirstKlaas:
So the implementation (on branch Inherited-fields) in my opinion solves the problem of #85 . At least in my test, as I modeled something similar. The use of Mixin classes is more a personal preference. But is also works on plain inheritance.

PR #147, opened 2019-06-30
Search for fields in (all) super classes.

FirstKlaas:
Also include fields define in super classes.

## Description
I changed the way the meta model class is looking for fields. Now also the base classes, as well as their mro classes are taken into account. Only new fields are added. 

## Motivation and Context
the use case is pretty much the same as described in #85. But not only the super classes (with abstract Meta classes) will work. Every class will work. These classes even don't have to   be subclsses of Model

## How Has This Been Tested?
I used my own project (where I'm testing the tortoise orm functionality) to verify if this works. If the approach is approved. I could add some test cases. As well as changing the documentation (together with some real life examples) 

## Checklist:
- [x ] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.
- [ x] I have read the **CONTRIBUTING** document.
- [ x] All new and existing tests passed.


grigi:
👍
First some minor comments:
Please run `make ci` and fix any errors, as this is in essence a replication of the CI test suite. (I see the contribution guide is a bit out of date, there is some useful commands as part of `make` to be documented.)

Also, could you make this compatible with multiple levels of inheritance?
e.g. C inherits B which inherits A 
Might have to add a recursive lookup? But it would be better to write a test case and see if it works?

Second some larger requests:
Tests?
Docs?

For tests you could put your test models in `tortoise/tests/testmodels.py` then create a new file, e.g. `test_mixins.py` and put some tests in there to verify that the added fields ARE in fact added. (I think that should be it?)

FirstKlaas:
Ok. Recursive checks in multiple levels are already included. I will run the ci. I will add some tests and update the documentation. I think id would be good to add some meaningful examples. Should be done by eow.

FirstKlaas:
So, make ci works without problems. See what I can do about test and documentation. Maybe I will start with the documentation.

FirstKlaas:
@grigi : Hi Nikolas, I have added a first version of documentation for this feature in models.rst. It would be great if you can have a quick look and give me feedback, if I'm heading the right direction. I tried to be concise in writing style. C&Cs are welcome.

tortoise/models.py
```diff
@@ -192,6 +192,30 @@ def __new__(mcs, name: str, bases, attrs: dict, *args, **kwargs):
...
+            """
+            Searching for class attributes of type fields.Field
+            in the given class.
+
+            If an attribute of the class is an instance of fields.Field,
+            then it will be added to the attrs dict. But only, if the
+            key is not already in the dict. So derived classes have a higher
+            precedence. Multiple Inheritence is supported from left to right.
+
+            After checking the given class, the funkction will look into
```
abondar: I think `funkction` is typo for `function`?

grigi:
I'll do a review tonight

docs/models.rst
```diff
@@ -87,7 +96,8 @@ We currently support single (non-composite) primary keys of any indexable field
 
 One must define a primary key by setting a ``pk`` parameter to ``True``.
 
-If you don't define a primary key, we will create a primary key of type ``IntField`` with name of ``id`` for you.
+If you don't define a primary key, we will create a primary key of type
+``IntField`` with name of ``id`` for you.
```
grigi: I'm not sure shortening these has any benefit. ReST ignores insignificant whitespace (Newlines are only significant when there is at least 2) and we have a project-wide "standard" of about-100-characters wide.

docs/models.rst
```diff
@@ -99,6 +109,62 @@ Any of these are valid primary key definitions in a Model:
...
+        first_name = CharField(20, null=True)
+
+        class Meta:
+            table = "user"
+
+
+    class RoleModel(TimestampMixin, NameMixin, MyAbstractBaseModel):
+
+        class Meta:
+            table = "role"
```
grigi: Is `table` allowed to not be specified?

Can I do `class Simple(TimestampMixin, Model)` ?

docs/models.rst
```diff
@@ -140,9 +206,14 @@ The ``Meta`` class
 In event model we got some more fields, that could be interesting for us.
 
 ``fields.ForeignKeyField('models.Tournament', related_name='events')``
-    Here we create foreign key reference to tournament. We create it by referring to model by it's literal, consisting of app name and model name. `models` is default app name, but you can change it in `class Meta` with `app = 'other'`.
+    Here we create foreign key reference to tournament. We create it by
+    referring to model by it's literal, consisting of app name and model
+    name. `models` is default app name, but you can change it in `class Meta`
+    with `app = 'other'`.
```
grigi: Too much churn, if this was an automated editor, see if you can turn off the hard wrapping? We run the risk of breaking things without noticing.

tortoise/models.py
```diff
@@ -192,6 +192,30 @@ def __new__(mcs, name: str, bases, attrs: dict, *args, **kwargs):
...
+            the classes according to the mro.
```
grigi: Please define `mro`

tortoise/models.py
FirstKlaas: No. intention. Just kidding. I will correct it of course. Anyway, this was not meant as the final version. Just a wip to get feedback on detail level and writing style. Is it understandable, etc. Thanks for finding the typo.

docs/models.rst
FirstKlaas: Hi @grigi, ok. I overread the 100-character-wide standard. I used doc8 to test the quality of the rst file and than you get 'line-too-long' errors. But I can live with a 100-character-wide rule. Do you want we to revert the line breaks? Or keep them here and I will never touch a long line again in the future?

docs/models.rst
FirstKlaas: Real question or something you suggest to make more clear in the documentation? Of course you do not need the Meta class at all. I didn't change anything in the standard behaviour. But I will add an example of what would be the smallest derived model.

docs/models.rst
FirstKlaas: Ok

tortoise/models.py
FirstKlaas: Do you have any project wide standard regarding linking to other resources? mro is a standard feature of python and I think it should explained in depth, but I agree, that for a better understanding, I should add a short explanation.

docs/models.rst
grigi: Suggest to make clearer in the documentation :-)

docs/models.rst
grigi: Please revert the line breaks, it makes it really difficult to find if there is any changes.
Of course fixing up a few ridiculous ones is fine :-)

tortoise/models.py
grigi: No, I think whatever you come up with is fine. maybe use ALL-CAPS to indicate it is an acronym at minimum?

FirstKlaas:
Added a testcase and reverted line breaks in model.rst. This should be it. Testmodels.py giv and conflict, which I cannot resolve. Version changed to 0.12.4 as 0.12.3 was released.

grigi:
A simple way to resolve conflicts is to merge master into this branch, and then you push the merge?
FYI

grigi:
I did https://github.com/tortoise/tortoise-orm/commit/f0f651e09be4d2b6cee29a095ebc9900d00ca2a3 and is merging this now :+1: Thanks for the work!

grigi:
Oh grr, Github is not allowing me to :-(
I can't change the incoming branch, I can't push to it (of course), but I can change each of your messages and comments...
Seems its priorities are a bit challenged.
Can you please do the equivalent of https://github.com/tortoise/tortoise-orm/commit/f0f651e09be4d2b6cee29a095ebc9900d00ca2a3 so I can merge?

FirstKlaas:
Hi @grigi , I'm not sure, what exactly you what me to do. Can you give me a hint, what to merge or revert or whatever where?

grigi:
Hi @FirstKlaas 
I want to merge this, but without the `.vscode/` and with the docs line widths reverted.
I made a commit https://github.com/tortoise/tortoise-orm/commit/f0f651e09be4d2b6cee29a095ebc9900d00ca2a3 that would do just that, but Github won't allow me to merge it and keep you as a commiter.

So, if you could either:
* Change this PR from `firstklaas:Inherited-fields` to `tortoise:Inherited-fields`
* or Cherry pick that commit into your branch
* or add your own commit to do the changes (as described above)

Then I _should_ be able to keep your history.

Apologies for the confusion.
My wife jokingly says I suffer from "over-explaining" and hence, being inarticulate.
I fear she is completely right  :grimacing:

grigi:
Thank you. I released v0.12.4 :smile:

Issue #148, opened 2019-06-30
Support for COMMENT


Baysul:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Make it so the `description` parameter is used to COMMENT on columns when `Tortoise.generate_schemas` is called, or have a separate field entirely named `comment`. For tables, I'm not sure what the best solution would be, maybe a metadata field?

**Describe alternatives you've considered**
Issuing the COMMENT commands manually.

grigi:
Thank you for the issue report.
We just added a `description="..."` parameter to all fields, and we can use it to omit `COMMENT '...'` when generating DDL.

This actually should be quite easy to do :-)
I confirmed all our supported databases support that syntax. The only "consideration" is that we need to handle basic escaping (at least we don't deal with user input here)

harshanarayana:
@grigi  Can I pick this one up?

grigi:
Sure!

grigi:
Thanks to @harshanarayana we have initial support for this in v0.12.3 \o/

PR #149, opened 2019-07-02
WIP: feat: enable row and table comments while generating schema

harshanarayana:
## Description
Enable row and table level comments while generating the schema based on #148 

## Motivation and Context
Having well commented and documented tables and rows in a DB table is always useful. This feature enables a way to define row and table level comments while defining your database Models. 

## How Has This Been Tested?
1. Changes were tested using `tox` command for all define env. 
2. Changes are tested against sqlite3, mysql and Postgresql databases. 

## Checklist:

> NOTE: This is a WIP PR. 

## To Do
- [x] Enable basic escaping to handle quotes and basic escape characters
- [x] Add additional tests to account for quoted string corner cases
- [x] Examples and Documentation update

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Changelog
 0.12.3
 ------
 * Added description attribute to Field class. (#124)
+* Added the ability to leverage field description from (#124) to generate table column comments and ability to add table level comments
 
 0.12.2
 ------
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -13,6 +13,7 @@ Contributors
 * Vitali Rebkavets ``@REVIMI``
 * Shlomy Balulu ``@shaloba``
 * Klaas Nebuhr ``@FirstKlaas`` 
+* Harsha Narayana ``@harshanarayana``
 
 Special Thanks
 ==============
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -86,7 +86,8 @@ Common parameters for fields:
     Defaults to ``False``.
 ``description`` (str):
     Human readable description of the field. Defaults to ``None``. This allows consumers 
-    to build automated documentation tooling based on the declarative model api. 
+    to build automated documentation tooling based on the declarative model api. This field is also
+    leveraged to generate comment messages for each database columns.
 
 Read-only properties:
 
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -114,6 +114,11 @@ The ``Meta`` class
 
         Set this to configure a manual table name, instead of a generated one
 
+        .. attribute:: table_description
+            :annotation: = ""
+
+            Set this to generate a comment message for the table being created for the current model
+
     .. attribute:: unique_together
         :annotation: = None
 
--- a/examples/basic_comments.py
+++ b/examples/basic_comments.py
@@ -11,7 +11,9 @@ from tortoise.models import Model
 class Event(Model):
     id = fields.IntField(pk=True)
     name = fields.TextField(description="Name of the event that corresponds to an action")
-    datetime = fields.DatetimeField(null=True, description="Datetime of when the event was generated")
+    datetime = fields.DatetimeField(
+        null=True, description="Datetime of when the event was generated"
+    )
 
     class Meta:
         table = "event"
--- a/examples/basic_comments.py
+++ b/examples/basic_comments.py
@@ -1,0 +1,40 @@
+"""
+This example demonstrates most basic operations with single model
+and a Table definition generation with comment support
+"""
+import asyncio
+
+from tortoise import Tortoise, fields
+from tortoise.models import Model
+
+
+class Event(Model):
+    id = fields.IntField(pk=True)
+    name = fields.TextField(description="Name of the event that corresponds to an action")
+    datetime = fields.DatetimeField(null=True, description="Datetime of when the event was generated")
+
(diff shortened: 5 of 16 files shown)


grigi:
Thank you for this :smile: 

I had a look, and the comments are done the same for MySQL and SQLite, and different for PostgreSQL.
There is in essence two comment types:
* In-line comments
* Separate comment commands

The PostgreSQL comment support is very implementation specific right now, and  the other two do it in a more "standard" way.

How about adding a capability called `inline_comments` that is set to `True` for MySQL and SQLite, and defaults to `False`.
If True, then the comment gets inserted in the "base" schema_generator.
https://github.com/tortoise/tortoise-orm/blob/master/tortoise/backends/base/client.py#L11
and
https://github.com/tortoise/tortoise-orm/blob/master/tortoise/backends/asyncpg/client.py#L77 ...

For PostgreSQL you could just have comments emitted as a post-table command much like you are already doing.
e.g. have a `_post_table()` function that could do whatever is needed AFTER a table decleration. e.g. comments or sharding setup etc... 
This should make it easier to add comment support for more databases later on?


Otherwise:
* I like your tests, simple and correct :+1:
* Add to documentation: https://tortoise-orm.readthedocs.io/en/latest/fields.html#reference stating that `description` will be generated as table/field comments
* Also document the comment for Model: https://tortoise-orm.readthedocs.io/en/latest/models.html#the-meta-class
* Add yourself to `CONTRIBUTORS.rst` and add a short description of the feature in the `README.rst` under 0.12.3

harshanarayana:
@grigi Thanks for getting back to me with such a short turn around time. I will do the necessary changes as per your comments and push the changes later today. 

PS: I need to find out why the inline comments did not work for me when I tested the sqlite3 case yesterday in that case.

grigi:
You didn't override the SQLite generation?
Note that SQLite literally just stores the entire CREATE TABLE string, and does nothing else with it. So whilst comments are valid, they are basically ignored.

One extra request: Please make your test case comment have a character that needs escaping, e.g. `\n` and `'`. To ensure that we do a best effort of escaping those values.

And the escaping rules are actually different for different DB's. e.g. MySQL is "helpful" and will replace unicode characters with their similar non-unicode variant. e.g. `¸` and `´` Which makes SQL injection even easier. But since this is DDL specified by developer we could get away with just handling the simpler ascii charaters.

harshanarayana:
> You didn't override the SQLite generation?

I tried to do the same way I did with mysql and it did not like the `COMMENT 'msg'`  part of the Schema definition. So I copied the string directly and tried to run it manually and found the same issue. Let me find the sqlite's documentation that explains the commenting for table/column  level and see what I did wrong the first time around.  Mind sharing the document you used as a reference to check the commenting behavior on sqlite part?

>  Please make your test case comment have a character that needs escaping, 
This is something I am working on. I am trying to formulate a simple enough mechanism to do this without a lot of additional overhead. Let me see what I can come up with. I will add necessary test cases around it when I implement the escaping behavior as part of this PR. 

I am only planning to handle the most basic cases of escaping and nothing too over the top. Simpler the better when it comes to maintainability.

harshanarayana:
@grigi  `sqlite` does support `-- comment` format in-line with table and columns. Would that be an agreeable replacement instead of using `COMMENT 'msg'` in case of `sqlite`?

grigi:
I just pasted some schema with COMMENT into my sqlite, and it didn't complain? I never actually looked at docs. Looking now there is no mention of `COMMENT`. 
I wonder if it is cause my SQLite has the spatialite patches applied?
It has no real value for SQLite as there are no real concept of DB management, so you can skip it for SQLite.

That renders some of my comments invalid... Apologies.
Looke some more an this DOES work (as SQLite literally copies the schema creating SQL into a text buffer):
https://stackoverflow.com/questions/7426205/sqlite-adding-comments-to-tables-and-columns

Still the value is dubious. Entierly your call.

grigi:
I saw your last comment after I did mine. It still stands `--` comments or None are perfectly acceptable replacements.

Now the work is just on how to make the comments as "standard" in code as possible. Because each of them do it TOTALLY differently...

`"...".format(comment='...')` at least works perfectly fine if there is no `{comment}` section. So I think that makes in-line comments optionally work right easily.
Just the post-create comments of PostgreSQL that needs to be different...

Maybe create a Column & Table schema with and without comments.
If there is a comment it uses the comment variant and passes the comment in.
Default is for them to be the same (so never emit comments).

Then PostgreSQL's implementation will have to stay unique?
And push everything through an `escape_comment()` method, so we can override comment method on a per-db level if needed?

harshanarayana:
@grigi Adding a `--` inline comment has one more side effect I did not account for earlier. When we generate the schema used for testing, it’s a collection of all table schema definition in one single string that is executed. So adding a `--` based comment marks everything after that line as a comment. So I am going to move to the `/* .. */` style of comments instead.

grigi:
I'll make a point of getting to it tomorrow. I had a big go-live today and is exhausted.

grigi:
I feel that it would be great to also generate the DDL formatted nicely. But that definitely out of scope...

harshanarayana:
> I feel that it would be great to also generate the DDL formatted nicely. But that definitely out of scope...

Absolutely. It can be really useful to have a neatly formatted DDL dumped so that you can just copy paste it

grigi:
And have a little command to generate and emit the DDL, to make it easier when you have to do all DDL through separate db admins.

PR #153, opened 2019-07-09
Relation key error

FirstKlaas:
Throwing ConfigurationError with a descriptive and helpful message.

## Description
see motivation

## Motivation and Context
When defining relations, you have to reference the related model by a string in the format <app>.<model>. If you have a typo, you only get an KeyError with a non helpful message. Now ConfigurationError are thrown with a helpful message. 

## How Has This Been Tested?
make ci. In another personal project, where I had this type. As this is not really a change, as the semantic of the code hasn't changed. Just providing more hints for developers.

## Checklist:


grigi:
Ah, this could really help newer users! Thank you! :tada: 
Please remember to add a quick description to the CHANGELOG
(no need to bump the version in the PR, unless a bugfix which should go out asap)

Could you also please add a test like this one: https://github.com/tortoise/tortoise-orm/blob/master/tortoise/tests/test_init.py#L42-L61
Create your own models file with a bad config and test that the right error message gets generated.
Just to ensure that we don't regress :-)

FirstKlaas:
Ok. Changelog for 0.12.5 right? As 0.12.4 was released. I will ad a
Testcase as you proposed.

Am Di., 9. Juli 2019 um 20:32 Uhr schrieb Nickolas Grigoriadis <
notifications@github.com>:

FirstKlaas:
So @grigi , added four Testcases. Wrong app name, wrong model name, missing '.' in reference, two '.' in reference.

tortoise/tests/test_bad_relation_reference.py
```diff
@@ -0,0 +1,94 @@
...
+        except ConfigurationError:
+            pass
+        Tortoise._inited = False
+
+    async def tearDown(self):
+        await Tortoise.close_connections()
+        await Tortoise._reset_apps()
+
+    async def test_wrong_app_init(self):
+        with self.assertRaises(ConfigurationError):
```
grigi: Could you do a `self.assertRaisesRegec(ConfigurationError, "<important substring of the message>"):` here?
It's useful to ensure that the message is correct for the test case.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "restructuredtext.confPath": "${workspaceFolder}/docs"
+}
```
grigi: I'm sure `.vscode/` is in the `.gitignore`. How did this get committed back?
Please try to not commit editor config. If we allow it, we may eventually end up with everyones editor config and having to handle those conflcts.

FirstKlaas:
Ok I'll remove it.  Sry.

Am Di., 9. Juli 2019 um 22:08 Uhr schrieb Nickolas Grigoriadis <
notifications@github.com>:

.gitignore
```diff
@@ -51,7 +51,6 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
-.vscode/
```
grigi: I think you misunderstood?

The `.gitignore` is files or paths that git itself should ignore.
So with `.vscode/` in the `.gitignore` it won't automatically add any files under that path.
Which is what we want.
(The docs about .gitignore: https://git-scm.com/docs/gitignore )

The comment was about the `.vscode/settings.json` file, which should not be committed.
And the entry in the .gitignore should stay.

FirstKlaas: Hi @grigi, yes I misunderstood. I guess I didn't read carefully enough your remarks. Personal settings shouldn't be committed to github. I wasn't aware of the fact, that I did. Which would have been clear if I had read carefully. *loop*! But I'm happy that `vscode/` is an entry in `.gitignore`. I reverted to the previous commit for this file.

grigi:
Don't forget to remove `.vscode/settings.json` please :-)

grigi:
Cool, I'm happy :+1: 
Once again, thank you :bowing_man:

PR #159, opened 2019-07-20
__models__ attr

superloach:
If a module contains the variable `__models__`, use that list to load the models in `_discover_models`, rather than all of the module's attributes. Use duck typing to check for existence and iterability.

In some cases, it may be nice to define models within another class:
```python
class Database(Tortoise):
    def __init__(self):
        print(self.InAClassEntry)

    class InAClassEntry(Model):
        pass

# InAClassEntry will be loaded from within Database
__models__ = [Database.InAClassEntry]
```
Additionally, it may be nice to manually exclude certain models temporarily:
```python
class GoodEntry(Model):
    pass

class BadEntry(Model):
    # some bad code
    pass

# BadEntry will not be loaded
__models__ = [GoodEntry]
```

## How Has This Been Tested?
I created `test__models__`, `test__models__good`, and `test__models__bad` - the tests pass.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,10 @@
 Changelog
 =========
 
+0.12.6
+------
+* Handle a __models__ variable within modules to override the model discovery mechanism.
+
 0.12.5
 ------
 * Using non registered models or wrong references causes an ConfigurationError with a helpful message.
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -14,6 +14,7 @@ Contributors
 * Shlomy Balulu ``@shaloba``
 * Klaas Nebuhr ``@FirstKlaas`` 
 * Harsha Narayana ``@harshanarayana``
+* Terra Brown ``@superloach``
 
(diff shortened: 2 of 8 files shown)


superloach:
Hooray, finally! 😅
Sorry about that, I wasn't really used to your test system.
(And I was a bit lazy at first in writing the attribute handing, but shhhh)

grigi:
Hi @superloach 
Thank you for the PR.
The test system is linting heavy, yes, and I think it is a good thing, as linters are good at minimising bikeshedding re style, and finding areas of confusion. And then because we need to test interaction between an external DB, most of our tests are integration-heavy ヽ(ー_ー )ノ

Generally the PR looks good. Could you please add yourself to CONTRIBUTORS.rst and create a new entry in CHANGELOG.rst

superloach:
Yeah, the testing is honestly fantastic, I just wasn't expecting so much.
And sure, I'll make those changes now!

grigi:
Looks good :+1: 
Thank you for your work @superloach

PR #160, opened 2019-07-21
Pass Postgres username as None instead of empty string

hongquan:
## Description

When connecting Postgres via Unix domain socket, we provide URL as `postgres:///somedb`, where username is omitted. Currently, with this URL, Tortoise ORM parses username as empty string (`""`) and pass it to `asyncpg`, which cause error.

Based on current implementation of `asyncpg`, the username should be passed as `None`, for it to be retrieved from environment variable and OS user:

https://github.com/MagicStack/asyncpg/blob/ae5a89db81dc97a762055cd574b3215559b3d59f/asyncpg/connect_utils.py#L322-L325

## Checklist:


grigi:
Thanks for this, I did not know about this functionality. :smile: 

Could you please ensure the same for host/port/user/pass/database as per: https://magicstack.github.io/asyncpg/current/api/index.html#connection
And document it a bit more verbosely.

So, e.g. `postgres://` would read all those data off environment variables?

Also, Please add a short description in the CHANGELOG.rst (new section, 0.12.7) and please add yourself to CONTRIBUTORS.rst

### One concern:

I have not tried specifying all details but for a unix socket via the URL. How would that work? `postgres://user:pass@/unix/path/database` → The file-path and db name is mixed? Would one have to do `postgres://user:pass@/unix/path?database=db` ? Is there a better way? :thinking: 

I see at https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING they discuss this same problem right above section `34.1.1.3` and recommend two options. The %-encoded option looks to be reasonably easy to adopt?

hongquan:
> So, e.g. postgres:// would read all those data off environment variables?

Right. Here are the variables: `PGHOST`, `PGPORT`, `PGUSER`, `PGPASSWORD`, `PGDATABASE`. If even `PGHOST` is not set, Postgres will fallback to Unix domain socket, as per [documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING):

> a Unix-domain socket connection is chosen if the host part is either empty or starts with a slash
-----
> The %-encoded option looks to be reasonably easy to adopt?

Yes. But this case is so rare that I doubt `asyncpg` even supports it. There is also a similar issue https://github.com/MagicStack/asyncpg/issues/419, this time with `:` character instead of `/`.

grigi:
Hmm, good point. In that case we just recommend the user uses the dict-format to configure.

tortoise/backends/base/config_generator.py
```diff
@@ -88,17 +93,21 @@ def expand_db_url(db_url: str, testing: bool = False) -> dict:
     vmap.update(db["vmap"])
     params[vmap["path"]] = path
     if vmap.get("hostname"):
-        params[vmap["hostname"]] = str(url.hostname or "")
+        params[vmap["hostname"]] = None if not url.hostname else str(url.hostname)
```
grigi: Is url.hostname not none if not defined? Or you could do `url.hostname or None`
This way seems a bit verbose?

hongquan: Thank. I've just made code less verbose. No need to do `str(url.username)` because `url.username` won't be other type than string and `None`.

grigi:
Could you please run `make style` as per Travis's output: https://travis-ci.com/tortoise/tortoise-orm/jobs/218158099 and then `make ci` to ensure that the build is basically sane?

Then I can merge this :smile:

hongquan:
@grigi Fixed linting error.

grigi:
Thank you @hongquan for you work :+1:

PR #162, opened 2019-08-06
[WIP] Fix self relations → Now TLC?

grigi:
This PR started off as an investigation as to the issues with v0.12.7, but whilst not finding the issue , it ended up being mostly TLC.

(Turns out the issue for 0.12.7 is only on the version of code in PyPi, not the tagged version in git, so probably I botched the release?)

Features:
* [x] ``.values()`` and ``.values_list()`` now default to all fields if none are specified.
* [x] ``generate_schema()`` now generates well-formatted DDL SQL statements.
* [x] Added ``TruncationTestCase`` testing class that truncates tables to allow faster testing of transactions. This allows tests to run 20% faster on SQLite, and about 7× faster for MySQL on my local system. (PG is 2×)
* [x] Cleaned up the examples a bit
  * [x] Re-did generate_schema example
     * [x] turned it into a test case
  * [x] A new recursive-lookup example
     * [x] turned it into a test case

Bugfixes:
* [x] Fixed state leak between databas drivers which could cause incorrect DDL generation.
* [x] Fixed missing table/column comment generation.
* [x] Fixed comment generation to escape properly for SQLite
* [x] Fixed comment generation for PostgreSQL to not duplicate comments

Chores:
* [x] Minor tweaking to tests to run with one process to simplify stack traces
* [x] Lots of small formatting/linting fixes
* [x] Fixed requirements generation to work right on Py3.7
* [x] Some documentation cleanup
* [x] Sphinx 2.1 breaks badly with the documentation theme, and <2 currently gives the best result, so force it to an older version for the forseeable future.

TODO:
* [ ] Clean up the examples some more
* [x] Do a more thorough M2M example & test case
* [ ] Explode aggregation tests


grigi:
Oh, py3.5 doesn't have insert-ordered dicts… Can I guarantee order, or should I make the DDL tests more lenient? (I actually WANT it strict, maybe just skip those on Py3.5?)

grigi:
Py3.5 doesn't support async comprehensions! Argh!
I'm seriously considering dropping py3.5 now. Grrr

grigi:
@abondar This PR is gotten massive enough could you please PR.

Also, I'm seriously considering dropping py3.5 support now. There is much missing syntax/feature support that I really want to use:
* variable annotations
* async generators (much cleaner implementations, at least)
* async comprehensions
* f-strings
* preserved order attribute decelerations and dics
* https://docs.python.org/3.8/whatsnew/3.6.html#asyncio
(possibly more?)

grigi:
@abondar Would you have any objection to us declaring py35 deprecated, and removing it on 0.14? Eg give one major version notice?

abondar:
I think it is good time to deprecate 3.5 python
I hope it will ease development for us :)

Issue #156, opened 2019-07-18
Distinguish between attribute fields and relation fields


FirstKlaas:
**The Problem / Issue**
I want to auto generate rst documentation for the models. In documentation it makes sense to describe _plain_ attributes, like IntField or CharField in a different block then relation attributes. Programmatically I cannot distinguish these to types. Of course I could do something like: `isinstance(...)` an list all corresponding classes. But if tortoise will introduce a new field type, the code breaks.
 
**Describe the solution you'd like**
My idea would be to introduce two new classes `AttributeField` and `RelationField`. Both are a subclass of Field. And all the current fields will inherit either from `AttributeField` or from `RelationField`. So `class IntField(Model):` will become `class IntField(AttributeField):`.

The approach is pretty straight forward, does not change any semantics and should have no performance impact. 

**Describe alternatives you've considered**
An alternative could be a property.

grigi:
I don't think there is a need to change the fields inheritance structure. We have ample metadata and any field will be mapped appropriately.
Also, there is no plans to add a new field type for relations that doesn't map cleanly to fk or m2m.
There is also no plans to change the _meta so it can be used for introspection. Hmm, it should actually be documented…

## Suggestion
In the model's `_meta` class you should have all the info you need?
e.g. `User._meta.fetch_fields` is all the fields that define relations.
Literally defined as `self.m2m_fields | self.backward_fk_fields | self.fk_fields` (it is a set)

Quick description of the _meta fields one can use for introspection: (all sets except `fields_map` which is a dict)
* `fields_map` → name:field dict of fields.
* `db_fields` → Columns on that model
* `fetch_fields` → Relations (could overlap with `db_fields`)
  * `fk_fields` → A fk column (will also occur in `db_fields`)
  * `backward_fk_fields` → A backwards fk relation (not in `db_fields`)
  * `m2m_fields` → a m2m field, either forward or backwards (not in `db_fields`)

Simply you could first take:
`db_fields - fetch_fields` → Non-relational
`fetch_fields` → Relational

We don't guarantee order, but on Py3.6+ you could use the order in `fields_map` to keep order as defined.

Also, the primary key is not always at the top of definition, you could force it to the top by using `pk_attr` which is a string containing the name of the primary key.

So, order of evaluation could be:
* PK: 
  `_meta.pk_attr`
* Other DB fields in order of definition: 
[...]

PR #170, opened 2019-08-17
Describe model api

### LINKED ISSUE #39: Schema discoverability
Define a method on Tortoise singleton and/or Model classes that will allow one to discover the exact schema, and any extra metadata.

This should be usable for generating migrations, serialisers and hypothesis strategies and similar.

Follow up from #14

grigi:
Fixes #39 & #156 

Done:
* [x] Implemented interface proposed in #156
* [x] Output is both JSON-serializable and Python native
* [x] Tests for JSON-able describe
* [x] Tests for Python-native describe
* [x] Documentation
* [x] Relational fields are represented using mostly the same fields, but with extra/missing fields as applicable.
* [x] Fixes for comment propagation & generation in DDL


grigi:
@abondar Please review.
This is currently missing the `db_type` data, as that is dependant on #97 but otherwise is good.
Please check that the description format is sane from your point of view, and that I didn't miss anything else.

PR #173, opened 2019-08-22
Default MySQL charset to utf8mb4. If a charset is provided it will also force the TABLE charset to the same.

grigi:


abondar:
Is it related to some issue? Or what is background on this PR?

grigi:
Oh, my bad. This is related to unicode issues I had on MySQL.
I was using a cloud provided MySQL, and they just left it at default config which means it defaults to latin-1
Then when I set it to utf8, it sort-of worked, but my typical test case of throwing emojis in broke.
Turns out because my LC_ALL is set to `en-utf8` it also defaulted to utf8, which is in fact utf8mb3, which is partial unicode support.

Hence utf8mb4.

And explicitly setting the table schema is a backup, since there is uncertainty if a table created will adopt the connection encoding or only the server encoding if unspecified. So I made it specify it explicitly.

PR #174, opened 2019-08-24
Fix PK generation for Bigint (PostgreSQL, MySQL) and added Smallint PK support

grigi:
Whilst working on #173 I noticed that a generated PK for a `BigIntField` was just on type `INT`.

Obviously this is wrong. Refactored autonumber PK DDL generation, and modified the tests so that those get tested.
I also added PK support for `SmallIntField` as it was properly supported in both PostgreSQL and MySQL.

SQLite only supports `INTEGER` autonumbers, which are just aliases for `__rowid__` which is always a 64-bit number.
Considering that SQLite doesn't actually honour any int type other than 64-bit unsigned, I left the minor discrepancy where foreign keys are still `BIGINT` and `SMALLINT`, even though the PK they link to is just `INTEGER`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ Changelog
 - Simple update is now ~3-6× faster
 - Partial update is now ~3× faster
 - Delete is now ~2.7x faster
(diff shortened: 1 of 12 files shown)


grigi:
@abondar Please review (ping)

PR #175, opened 2019-08-25
Ensure that default PK is added to the top of the attrs.

grigi:
It was slightly annoying that the default PK was not at the top of the generated DDL.

This will only preserve order for Py3.6+


grigi:
@abondar Please review (ping)

PR #176, opened 2019-08-29
Hotfix/add connection pool for postgresql only

Quinn-Fang:
Add connection pool and make it work correctly.

## Description
So basically, I am using Tortoise in my Project which needs to handle more than 20000 QPS,  thus the asyncpg connection pool is required. So I've tried to incorporate PR 141's asyncpg/client and base/client into tortoise, so at lease I can get a temporary connection pool version of tortoise to get my project work. But after a little pressure test, (somewhere around 120 requests/s, 3 create per transaction, 40 transactions/s), error occurs, says a "transaction is already closed", which means the ddl s inside the same transaction were using different connections. it's because the execute_insert function inside AsyncpgDBClient is creating new connections wether it's inside a transaction or not, so in order to let all ddl s use same connection, I use a contextvars to store the transactionwrapper instance in the current coroutine context.

### DIFF
--- a/tortoise/backends/asyncpg/client.py
+++ b/tortoise/backends/asyncpg/client.py
@@ -25,7 +25,7 @@ if TYPE_CHECKING:
     from typing import AsyncContextManager
 
 from contextvars import ContextVar
-current_transaction = ContextVar("Current TransactionWrapper obj")
+current_transaction = ContextVar("Current TransactionWrapper obj", default=None)
 
 
 def retry_connection(func):
--- a/tortoise/backends/asyncpg/client.py
+++ b/tortoise/backends/asyncpg/client.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 from functools import wraps
-from typing import List, Optional, SupportsInt  # noqa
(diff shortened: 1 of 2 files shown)


grigi:
Interesting work. I'm glad you managed to work around that same issue I got when I was looking at it months ago.

So is it only `execute_insert` that forced a new connection? I'm pretty sure that all the `execute_*` ones do… (The `self.acquire_connection(…)` does it?)
I see you added two more contextvars, `current_transaction` (global) and `self._current_transaction`? Why is that?

grigi:
I spent some time looking at this and I think I see what you did:

The transaction wrapper check:
```py
transaction_wrapper = current_transaction.get()
if transaction_wrapper:
    transaction_wrapper._connection
```
is logic that _should_ really be run by:
```py
```

Also, you used your own connection-local connection wrapper object, instead of the global one.

:thinking:

grigi:
Hi @Quinn-Fang We fixed the context leak, and have PR #229 just about done right now that should implement pooling.
Thank you for this PR, it was very helpful in helping to find the root cause.

Could you please have a look at & test #229 ?

grigi:
Closing this as we have a pooling implementation that seemingly works in master (but not yet released)

PR #179, opened 2019-09-01
queryset.update() uses the DB-specific to_db_value()

### LINKED ISSUE #178: sqlite boolfield update error
**Description**
Update operations on models with boolfields generates "OperationalError: no such column: true
" error on sqlite.

**To Reproduce**
```python
from tortoise import fields


import tortoise


class BlogEntry(Model):
    id = fields.IntField(pk=True)
    name = fields.TextField(default="new entry", null=False)
    body = fields.TextField(default="replace with some text here...", null=False)
    is_published = fields.BooleanField(default=False, null=False)


async def init():
    await tortoise.Tortoise.init(
        db_url="sqlite://db.sqlite3", modules={"models": ["models"]}
    )
    await tortoise.Tortoise.generate_schemas()


async def test_update():
    entry = await BlogEntry.create(name="demo", is_published=False)
    result = await BlogEntry.filter(name="demo").update(
        is_published=True
    )  # OperationalError  no such column: true

    assert result.is_published == True

```

**Expected behavior**
Update changes bool values.

**Additional context**
```
OperationalError: no such column: true
> /site-packages/tortoise/backends/sqlite/client.py(28)translate_exceptions_()
     26             return await func(self, query, *args)
     27         except sqlite3.OperationalError as exc:
---> 28             raise OperationalError(exc)
     29         except sqlite3.IntegrityError as exc:
     30             raise IntegrityError(exc)

ipdb> query                                                                                                                                                                                                                                               
'UPDATE "blogentry" SET "is_published"=true WHERE "name"=\'demo\''

```
It seems that [backends.sqlite.executor.to_db_bool](https://github.com/tortoise/tortoise-orm/blob/0.13.1/tortoise/backends/sqlite/executor.py#L12) maybe missing from the update code?

grigi:
Fixes #178

* Ensure that `«queryset».update()` is properly tested, and that it calls the right `to_db_value()`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,7 +13,7 @@ use_parentheses=True
 line_length=100
 
 [green]
-processes     = 1
+#processes     = 1
 no-skip-report= True
 initializer   = tortoise.contrib.test.env_initializer
 finalizer     = tortoise.contrib.test.finalizer
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,7 +13,6 @@ use_parentheses=True
 line_length=100
 
 [green]
-#processes     = 1
 no-skip-report= True
 initializer   = tortoise.contrib.test.env_initializer
 finalizer     = tortoise.contrib.test.finalizer
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 
 Changelog
 =========
+0.13.3
+------
+- ``«queryset».update(…)`` now correctly uses the DB-specific ``to_db_value()``
+
 0.13.2
 ------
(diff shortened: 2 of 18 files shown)


setup.cfg
```diff
@@ -13,7 +13,7 @@ use_parentheses=True
 line_length=100
 
 [green]
-processes     = 1
+#processes     = 1
```
abondar: What about removing it at all? Or why it have to stay?

grigi: Er, typo?

grigi:
@abondar This PR expanded in scope, it now also tests related issues that I found whilst working on it originally. Please review again.
Or should I break this up?

Issue #19, opened 2018-07-04
Discussion: v1.0.0 target


grigi:
## MUST HAVE for v1.0.0:
* [x] MySQL (#37)
* [x] Tortoise.init() refactor (#14 #29)
* [x] ~JSONField (#20)~
* [ ] Fix DB pools (#41)
* [ ] Fields Schema generation refactoring
* [ ] Better documentation
  * [x] ~Document init/config (#35)~
  * [x] ~Document DB_URL (#35)~
  * [ ] Marketing material
  * [ ] Contribution guide
    * [ ] Policy re extra DB backends (#31)
* [x] ~Test framework overhaul (#29 #33)~
* [ ] Better testing (#38)
  * [x] ~Test all data fields (#20)~
  * [x] ~Test `.values()` and `.values_list()` (#20)~
  * [x] ~Test transactions (#29 #33)~
  * [x] Filter tests for data fields (#38)
  * [ ] Test all relation fields (#38)
  * [ ] Filter tests for relation fields (#38)
  * [x] ~Test all Model methods (#34)~
  * [x] ~Test all QuerySet methods (#25)~
  * [x] ~Test the test classes (#25)~
  * [x] ~Test the `db_url` resolver (#25)~
  * [x] ~Test init (#35)~
* [x] ~Clean up Tortoise exception structures (#24)~
  * [x] ~Replace `assert` and `AssertionError` with appropriate Tortoise exception. (#24)~
  * [x] ~Formalise exception inheritance (#24)~
* [ ] Cleanup of "Issues" and "Projects"

## SHOULD HAVE for v1.0.0:
* [ ] Better documentation (see https://realpython.com/documenting-python-code/#documenting-your-python-projects)
  * [ ] Document `app` concept
  * [x] ~Expand documentation re Fields (#25)~
  * [x] ~Expand documentation re Querysets (#25)~
  * [ ] Expand documentation re Filtering
  * [x] ~Expand documentation re Exceptions (#25)~
  * [ ] Expand documentation re Database backends
  * [ ] Expand documentation re Models
  * [ ] Cleaned up examples
    * [ ] Simple usage examples as doctests: https://docs.python.org/2/library/doctest.html#simple-usage-checking-examples-in-a-text-file
    * [ ] Add doctests to test suite
    * [ ] What to do with larger examples?
  * [x] ~Update README with travis/coveralls badges (#24)~
* [x] ~Python 3.7 support (#22)~
[...]

PR #186, opened 2019-09-06
starlette example

nsidnev:
## Description
Added integration with `Starlette` and usage example.

## Motivation and Context

Help with #19.

## How Has This Been Tested?

Manually by running a script and sending requests.

## Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,7 +4,7 @@ exclude =
 ignore = W503
 
 [isort]
-known_third_party=aiosqlite,ciso8601,quart,sanic
+known_third_party=aiosqlite,ciso8601,quart,sanic,starlette
 not_skip=__init__.py
 multi_line_output=3
 include_trailing_comma=True
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,7 +4,7 @@ exclude =
 ignore = W503
 
 [isort]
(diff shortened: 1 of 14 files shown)


setup.cfg
```diff
@@ -4,7 +4,7 @@ exclude =
 ignore = W503
 
 [isort]
-known_third_party=aiosqlite,ciso8601,quart,sanic
+known_third_party=aiosqlite,ciso8601,quart,sanic,starlette
```
grigi: I think add `uvicorn` to fix py3.5

grigi:
Thank you for this, it looks pretty standard. Seems the Py3.5 build is failing, but that is probably just because Starlette doesn't support py3.5 (And we plan to drop support for that very, very, soon).

It is getting to the point where we have near identical code on three different examples, and now I'm thinking if there is a way to increase commonality between then? :thinking: 
Although the actual contrib portions are quite small right now. Mostly just documentation...

Could you add it to the generated sphinx documentation? something like this: https://tortoise-orm.readthedocs.io/en/latest/contrib/quart.html

Otherwise, looks good :+1:

setup.cfg
nsidnev: Yeap, it helped. Forgot about adding this. Thanks!

PR #192, opened 2019-09-21
More tests

grigi:
Ported the tests over from #190, including some bugfixes.


grigi:
Coveralls seem very confused. Adding more tests reduces coverage, but when you look at the detail page, coverage is up...

PR #193, opened 2019-09-22
py.test and moving tests out of lib path

grigi:
Done:
* [x] Pulled the py.test changes from #141
* [x] Moved tests out of the library path.
* [x] Moved the of the dev-oriented requirements out to the tests folder.
* [x] Collapsed Travis/tox runs to do full test suite for each supported language
* [x] Do backup tests via green instead of py.test
* [x] Moved extra_requirements.txt to tests/
* [x] `make ci` now runs tests against ALL supported DB's
* [x] made py.test runs multi-process for a significant speedup (still slower than green, though)

This should do a few things:
* Less CI runs on Travis should let it complete faster.
* Cleaned up the root folder
* Tests no longer in the library module path, so a little easier to keep separate.
* Since coveralls broke in handling multiple test runs, at least now the coverage reported will be correct.
* Default to py.test due to a bit more features (e.g. collect warnings, etc) Although performance is notably slower than green, it is still worth it for the extra flexibility.


grigi:
@abondar Please review. I took the relevant changes from your PR #141

Issue #155, opened 2019-07-15
Case-insensitive equals


gausie:
**Is your feature request related to a problem? Please describe.**
Since we have case-insensitive contains and case-insensitive starts/ends with, it would be really nice to just be able to case-insensitive compare!

**Describe the solution you'd like**
Perhaps `value__iequals="ThE VaLuE"`

**Describe alternatives you've considered**
Using startswith/endswith would obviously be error prone

grigi:
This is an interesting request.

I did some quick research and found the following:
* MySQL is by default case insensitive, but may not be (depends on how the DB was setup). Enforcing case sensitivity can be enforced by special `BINARY` keyword. (Can be controlled by the `COLLATE` encoding on a per-column basis) Performance is best when using `COLLATE`.
* PostgreSQL is by default case sensitive, but has a fully functional `LOWER()` function, catch is it needs an index if you want to avoid a tablescan `INDEX ON groups (LOWER(name))`
* SQLite has incomplete support for for `UPPER()` or `LOWER()` (always a tablescan) and incomplete support for `COLLATE NOCASE` character encoding. Un-wildcarded `LIKE` also works. Only works for `ascii` characters, and other is always compared case sensitively. Performance info is extremely conflicting, but I suspect may be best `COLLATE` or `LIKE`?

It seems that for good performance (or any kind of predictability of function for MySQL), we should actually know if case-insensitive searches would be done at table-creation time.
The PostgreSQL implementation seems the simplest, and MySQL has the most environmental factors that can interfere, (or give us the functionality for free).

A generic, but non-optimal solution would be to do an unwildcarded `LIKE` (or `ILIKE` for PostgreSQL).

hongquan:
PostgreSQL has [`CIText`](https://www.postgresql.org/docs/current/citext.html) data type, which perform case-insensitive match. It would be nice if TortoiseORM support some way to define `CICharField` to make use of this data type.

grigi:
Good point, we could also consider different DB types for the same field depending on use?

e.g. in PostgreSQL if we want a case-sensetive charfield with exact match only, we just use a regular B-tree index. if we want indexed LIKE, we use a different/extra index.
If we want it case insensitive we could use CIText or index on LOWER(field) (will have to evaluate which one works better).
If you want full-text searching, even more complicated indexes with collation.

And if one could indicate expected use of field, then we could generate a good DDL for it.

[...]

PR #194, opened 2019-09-22
A case-insensitive filter "__iexact" was implemented

pilat:
__iexact filter implementation.

## Description
That filter will cast your fields to VARCHAR, then it will perform UPPER() to field values and desired value. So, the final query will be looked like that:
```sql
... WHERE UPPER(CAST("email" AS VARCHAR))=UPPER('AAA@localhost')
```

## Motivation and Context
That request may resolve impossibility of case-insensitive filtering. I hope that it's the simplest solution for cases like that https://github.com/tortoise/tortoise-orm/issues/155.
`__iexact` filter exists in [Django ORM](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#iexact) (but it uses ILIKE syntax there)

## How Has This Been Tested?
That PR was tested with built-in unittests and also it was tested in my machine with Postgres. Unfortunately I haven't test it with the others environments.

## Checklist:

If it will be necessary I may change documentation and examples as well.


grigi:
Hi @pilat Thank you for this PR!

Could you please add yourself to `CONTRIBUTORS.rst` and add a section detailing the new feature in `CHANGELOG.rst` create a new section e.g. 0.13.7?

Otherwise, this looks good :+1:

I suppose that #155 will now just change to ensuring case insensitive operations are performant?
(Will have to expand the benchmark to test that we use the DB's correctly, instead of right now where I have been focusing on finding Tortoise inefficiencies.)

grigi:
Thank for your good work :+1:

Issue #8, opened 2018-06-28
Migrations


abondar:
(edited by @grigi )
Migrations is currently planned as a post-v1.0 feature.

Some useful resources for getting this working right now:
* https://github.com/tortoise/tortoise-orm/issues/8#issuecomment-534946871 (Native python, MySQL)
* https://github.com/tortoise/tortoise-orm/issues/8#issuecomment-575982472 (Go, MySQL/PostgreSQL/SQLite)

## Forward migrations
Best guess at this time for a complete solution, not a "quick" solution:
- [ ] Make the `describe_model()` contain all the data one needs to generate DDL from.
- [ ] Generate a sequence of high-level DDL instructions from `describe_model()`
- [ ] Port `generate_schema` to use this high-level DDL instructions to generate a schema in the requested dialect. (This would require some decoupling from the driver instance, look at #72)
- [ ] Build a persistance model (read/write) for high-level DDL insctuctions much like Django Migrations
- [ ] And need a way to run "management" commands. So possibly a standard command-line utility.
- [ ] Build a full model (e.g. same format as `describe_model()`)from a series of high-level DDL instructions (as in read the migration persistance model)
- [ ] Diff models, and generate a diff DDL instruction set.
- [ ] If a Diff requires a default not pre-specified, we need to ask the user for something.
- [ ] Have some way of determining the "version" of the models persisted in the DB.
- [ ] Put it all together and make it work

I'm not particularily happy about many migration systems storing state in the DB itself, as for some environments that won't work, but it is a very good place from a consistency PoV. We should have a backup for when we generate DDL that one would pass on to the DB team in cases where DDL changes have an established process (e.g. enterprise/consulting)

## Data migrations
Proper data migration management makes this very useful:
- [ ] Allow custom user scripts as migrations
- [ ] Allow data-migrations to get a version of a model as it was at that time. So ability to build a model class from the intermediate `describe_model()` data
- [ ] Handle DB edge cases, like Postgres being able to do DDL changes in a transaction OR data changes in a transaction, but not both.

## Backward migrations
There is some cases where backwards migrations cannot be done as a forward migration clobbered/destroyed some data.
In cases where we detect that we need mising data, and request a sane default.
What do we do about data migrations? Is it safe to just emit a warning?

Marlysson:
Which models did you think it's necessary or which design this feature need..

Zeliboba5:
[...]

PR #200, opened 2019-09-26
MySQL engine fixes

### LINKED ISSUE #197: Foregin keys on MySQL
**Describe the bug**
Foreign key constraints are not working on MySQL/MariaDB

**To Reproduce**

models/countries.py:
```python
class Countries(Model):
    id = IntField(pk=True)
    code = CharField(max_length=2, unique=True, description="ISO 3166-1")
    name = CharField(max_length=50, index=True)
```

models/users.py:
```python
class Users(Model):
    id = BigIntField(pk=True)
    firstname = CharField(max_length=128)
    lastname = CharField(max_length=128)

    country = ForeignKeyField(model_name="models.Countries")
```

**Expected behavior**
Foreign keys should be created.

**Additional context**

The created SQL is like this:
```sql
CREATE TABLE `countries` (
    `id` INT UNSIGNED NOT NULL PRIMARY KEY AUTO_INCREMENT,
    `code` VARCHAR(2) NOT NULL UNIQUE COMMENT 'ISO 3166-1',
    `name` VARCHAR(50) NOT NULL,
    KEY `countries_name_991101_idx` (`name`)
) CHARACTER SET utf8mb4;

CREATE TABLE `users` (
    `id` BIGINT UNSIGNED NOT NULL PRIMARY KEY AUTO_INCREMENT,
    `firstname` VARCHAR(128) NOT NULL,
    `lastname` VARCHAR(128) NOT NULL,
    `country_id` INT NOT NULL REFERENCES `countries` (`id`) ON DELETE CASCADE
) CHARACTER SET utf8mb4;
```
1st problem is that pk fields are unsigned, but ForeignKeyFields are not. When I tried to add a foreign key by SQL from a mysql gui, I got the following error: "Foreign key constraint is incorrectly formed"
More info: https://stackoverflow.com/questions/8434518/mysql-foreign-key-constraint-is-incorrectly-formed-error

The 2nd is it seems MySQL can't handle the REFERENCES in the same row as the field (it just ignores it), but it needs a separate FOREIGN KEY row inside CREATE TABLE.

grigi:
This PR attempts to fix #197 

There is a possibility that our MySQL support was troublesome in part because we only tested `InnoDB` on `MariaDB`. Extend the testing to `MyISAM`.

Done:
* [x] Ensured that int PK's are not unsigned (we don't support unsigned ints at this stage)
* [x] Enable the MySQL db client to specify the testing engine
* [x] Test if engine is not InnoDB, then mark `transaction_support` as `False`
* [x] Skip any transaction-specific test cases when DB doesn't support transactions.
* [x] Added `MyISAM` engine to CI
* [x] Enable `STRICT_TRANS_TABLES` sql_mode by default as is recommended. (Apparently only needed for older versions, but supposedly fix many runtime issues. Django also does it.)
* [x] Investigate if in-line REFERENCES just works in MariaDB or also old MySQL? *Definitely needed*
* Refactor SQL schema generation by putting key/indexes into to bottom of create statement for MySQL to conform to recommended MySQL spec:
  * [x] For FK
  * [x] For M2M through tables
  * [x] Indexes as per https://github.com/tortoise/tortoise-orm/issues/8#issuecomment-534946871

Defer proper refactoring work to 0.14

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,11 @@
 Changelog
 =========
 
+0.13.9
+------
+- Foreign keys and indexes are now defined correctly in MySQL so that they take effect as expected
+- MySQL now doesn't warn of unsafe index creation anymore
+
 0.13.8
 ------
 - Fixed bug in schema creation for MySQL where non-int PK did not get declared properly (#195)
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-checkfiles = tortoise/ examples/ tests/ setup.py conftest.py
+checkfiles = tortoise/ examples/ tests/ conftest.py
 black_opts = -l 100
 py_warn = PYTHONWARNINGS=default PYTHONASYNCIODEBUG=1 PYTHONDEBUG=x PYTHONDEVMODE=dev
 
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,8 @@ test: deps
 _testall:
 	-$(py_warn) TORTOISE_TEST_DB=sqlite://:memory: py.test -q --cov-report=
 	-python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:@127.0.0.1:5432/test_\{\} py.test -q --cov-append --cov-report=
-	-$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}" py.test -q --cov-append
+	-$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test -q --cov-append --cov-report=
+	-$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}?storage_engine=INNODB" py.test -q --cov-append
 
 testall: deps _testall
 
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ _testall:
 	-$(py_warn) TORTOISE_TEST_DB=sqlite://:memory: py.test -q --cov-report=
 	-python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:@127.0.0.1:5432/test_\{\} py.test -q --cov-append --cov-report=
 	-$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test -q --cov-append --cov-report=
-	-$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}?storage_engine=INNODB" py.test -q --cov-append
+	-$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}" py.test -q --cov-append
 
 testall: deps _testall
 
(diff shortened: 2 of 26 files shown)


grigi:
I now have a dilemma, there is a bug (which makes it urgent) and there is a quick fix, and a proper fix.

The quick fix would be to shoe-horn the schema generation in like https://github.com/tortoise/tortoise-orm/issues/8#issuecomment-534946871
(Using magic placeholders, I mean).

The proper fix would be to do the entire Field refactoring in #97, extend the `Tortoise.describe_model(…)` to generate the correct DB types off the refactored fields, and then use that to generate schemas based on a simple implementation that just iterates over that, and generates a set of data:
```py
safe_create: bool           # IF NOT EXISTS
table_name: str             #
field_statements: List[str] # Field definitions
tail_statements: List[str]  # Keys/Constraints/Indexes be inside the block, but after fields
table_params: List[str]     # Table parameters, such as encoding, etc...
post_statements: List[str]  # Statements that should be after the create table (PG has lots of these)
```
Which is very easily templateable. And it would allow us to extend the same datastructures for eventually Migrations.

So, @wallneradam and @abondar What do you recommend?
I feel I should do the quick fix here, get it out, and do the proper refactor for v0.14?

wallneradam:
Thanks for asking my opinion...
If I understand well, it is only a problem in MySQL (atm). My "quick fix" uses magic placeholder just beause I didn't want to override the whole `BaseSchemaGenerator._get_table_sql` method. I just wanted to make as short fix as possible into my application. 

I think that method needs a refactoring. It should have more small methods to be able to override at the needed point by the DB specific schema generator. Maybe the index creations should be split into e.g. `_get_field_index_sql` and `_get_separated_index_sql` or some better names.

I think these should be placed into the specific SchemaGenerator classes, not into Field classes as you specified it in #97. Fields should just describe the common parameters and generator should generate them as the database engine needs them. So in my opinion every differences between engines should handle separately. But this is just my opinion and I don't know Tortoise source well yet.

Another problem with these `_get..._sql` methods that they get only the name of the field, not the object nor the model. I created another quick fix (the UNSIGNED problem), and it was hard because of this. I couldnt easily check if a field is a foreign key or not. (I solved it by saving the model in `_create_string` and get the object from the model by field name in the `_get_index_sql`).
So the solution would be to be able to generate the SQL based on the field type here, not in the Field classes.

grigi:
Indeed @wallneradam, I'm going to do the "quick fix" for this PR, to resolve the issue for you. This may be the last release before we drop Py3.5 support.

And then do fields over properly for 0.14
This will include a complete refactor of how fields are declared, and the DB-specific workarounds for them, and building the DDL off the described models ONLY.

Then we can stop hacking up more `_get...sql` methods in schema_generator. Also, I would want the dialect/driver discovery to be easier.

Firstly I wanted to make fields a bit more declarative, so that the need to override functions is less required. There is many problems with the current setup:
* Some field overrides are in the executor
* Some field overrides are in the schema_generator
* To create a DDL one needs to have an instantiated client so that one can resolve the correct schema_generator and executor
* Hence to add each field, one needs to define logic for it in a minimum of 2 places, median of 3 and a current worst case of 6 places. 

I think it really needs a redesign, which is more invasive than a simple refactor.

Some requirements I think:
* Both fields and dialects should be able to be extendede OUTSIDE the Tortoise codebase:
  * One should be able to add a field with DB overrides without having to alter existing dialects.
  * One should be able to add a dialect that has field overrides without having to alter existing fields.
* All schemas should be generated off public interfaces without having to do any introspection at all.
* When requesting model description we should be able to get all DB-related data as well.
* One should be able to get DB-related data for each known dialect effortlessly. (including DDL generation). It should work without requiring the DB client connectors installed.

grigi:
@abondar Please review

Issue #196, opened 2019-09-26
Field operator support


wallneradam:
**Is your feature request related to a problem? Please describe.**
Take the following example:
```python
class Users(BaseModel):
    id = BigIntField(pk=True)

    firstname = CharField(max_length=128)
    lastname = CharField(max_length=128)

    @property
    def full_name(self):
        return self.firstname + " " + self.lastname
```

The IDE shows an error, that CharField does not define `__add__` method.

**Describe the solution you'd like**
It would be good to make fields implement basic, possible operators.

**Describe alternatives you've considered**

It may be a solution to make basic fields also inherit from base Python types, e.g. CharField inherit not just Field, but str.

Of course I can workaround the above issue:
```python
return str(self.firstname) + " " + str(self.lastname)
```

grigi:
This is more of a metaprogramming issue.

What happens is that the type system sees the field as a class variable, and assumes that is what is running when you use the model.

What actually happens is that the fields as class variables are removed, and the constructor will instead populate an instance variable of the Python type. Which for CharField is str.

So at runtime you work with a plain Python string, as one would expect.

But the IDE type system doesn't know it.

I built a pylint plugin to help with this, but thesedays ides may need their own plugin. I know mypy plugins is now a thing, but unsure if it will help at all.

1. It works as you expect
2. IDEs need help with metaprogramming

wallneradam:
Thanks, now I understand how it works.
Hello, I have an idea to make IDEs know the Python class of fields without plugins:

```python
class FieldMeta(type):
    def __new__(cls, name, bases, attrs):
        # Remove fake base classes
        if len(bases) > 1:
            bases = (bases[0],)
        return type.__new__(cls, name, bases, attrs)
        

class Field(metaclass=FieldMeta):  # Use metaclass to modify subclasses before creation
    __slots__ = ('p1', 'p2')
    def __init__(self, p1=None, p2=None):
        self.p1 = p1
        self.p2 = p2


class IntField(Field, int):  # Acts like an int object for IDEs
    def __init__(self, p1=None, *args, **kwargs):
        super().__init__(p1, *args, **kwargs)
        

class CharField(Field, str):  # Acts like an str object for IDEs
    __slots__ = ('p3')
    def __init__(self, p1=None, *args, **kwargs):
        super().__init__(p1, *args, **kwargs)
```

In brief:
- You add the type you want the class act as a 2nd base class. (you can add more classes)
[...]

PR #202, opened 2019-09-30
Set field types by 2nd base class

wallneradam:
## Description
This makes IDEs happy by seeing fields acts as a base class (like int, str, float...), and bit more intuitive to set the Python class of a Field subclass.

## Motivation and Context
It is discussed in #196

## How Has This Been Tested?
I tested it in my application only.

## Checklist:

### DIFF
--- a/tortoise/fields.py
+++ b/tortoise/fields.py
@@ -15,6 +15,7 @@ from tortoise.utils import QueryAsyncIterator
 if TYPE_CHECKING:  # pragma: nocoverage
     from tortoise.models import Model
 
+
 CASCADE = "CASCADE"
 RESTRICT = "RESTRICT"
 SET_NULL = "SET NULL"
@@ -25,13 +26,26 @@ JSON_DUMPS = functools.partial(json.dumps, separators=(",", ":"))
 JSON_LOADS = json.loads
 
 
-class Field:
+class _FieldMeta(type):
+    def __new__(mcs, name, bases, attrs):
+        if len(bases) > 1:
+            # Instantiate class with only the 1st base class (should be Field)
+            cls = type.__new__(mcs, name, (bases[0],), attrs)
+            # All other base classes are our meta types, we store them in class attributes
+            cls.type = bases[1] if len(bases) == 2 else bases[1:]
+            return cls
+        else:
+            return type.__new__(mcs, name, bases, attrs)
+
+
+class Field(metaclass=_FieldMeta):
     """
     Base Field type.
     """
 
+    type = None  # Type is a readonly property for the instance, it is set by _FieldMeta
+
     __slots__ = (
-        "type",
         "source_field",
         "generated",
         "pk",
@@ -48,7 +62,6 @@ class Field:
 
     def __init__(
         self,
-        type=None,  # pylint: disable=W0622
         source_field: Optional[str] = None,
         generated: bool = False,
         pk: bool = False,
@@ -61,7 +74,6 @@ class Field:
         description: Optional[str] = None,
         **kwargs
     ) -> None:
-        self.type = type
         self.source_field = source_field
         self.generated = generated
         self.pk = pk
@@ -89,7 +101,7 @@ class Field:
         return self.default is None and not self.null and not self.generated
 
 
-class IntField(Field):
+class IntField(Field, int):
     """
     Integer field. (32-bit signed)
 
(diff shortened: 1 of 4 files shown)


grigi:
Weird:
```py
>>> bool
<class 'bool'>
>>> bool().__class__
<class 'bool'>
```

Then again both mypy and PyLint is complaining about the same thing. But not about `int`. Both are literals :man_shrugging: 

Put a `# pylint: disable=E0239` at the top of the Fields file to silence PyLint/Codacy

Consider putting `  # type: ignore` in the line with bool. This is a case where we are abusing the Python syntax possibly too much?

wallneradam:
>. Still it behaves just fine? 
Yes it works, because it is just used as a kind of type hint and not a real subclass. It just needs to be suppressed.

grigi:
I like what you did here :+1:
Please add yourself to `CONTRIBUTORS.rst` and add a quick description of the change in `CHANGELOG.rst` (create a new 0.13.9 section.)

tortoise/fields.py
```diff
@@ -375,7 +379,8 @@ class ForeignKeyField(Field):
                 Can only be set is field has a ``default`` set.
     """
 
-    __slots__ = ("model_name", "related_name", "on_delete")
+    # Here type will be set later, so we need a slot to be able to write it
+    __slots__ = ("type", "model_name", "related_name", "on_delete")
```
grigi: Slotted instance variables and class variables cause conflicts. I have learnt to be very careful with this.
Consider injecting type as a class variable in the metaclass?
```py
attrs['type'] = bases[1] if len(bases) == 2 else bases[1:]
```
And then remove the `type = None` in the base class?

grigi:
I tried disabling the linters re bool, and they are not having any of it :-(

I think we may have to just not do this for `bool`.
Maybe set type explicitly as a class variable, and then don't subclass for bool?

tortoise/fields.py
wallneradam: It works.
I think you mean this:
```python
class A:
    __slots__ = ('x',)
    x = 1
```
Which doesn't work, because x is readonly. More here: https://stackoverflow.com/questions/5757924/python-slots-inheritance-and-class-variables-attribute-is-read-only-b

But this does:
```python
class A:
    x = 1

    
class B(A):
    __slots__ = ('x',)


a = A()
print(a.x)
a.x = 2
print(a.x)
```
And we use the latter here.

wallneradam:
> Then again both mypy and PyLint is complaining about the same thing. But not about `int`. Both are literals

https://stackoverflow.com/questions/10061752/which-classes-cannot-be-subclassed?lq=1
Only bool is not subclassable, though I not 100% agree with the reasons. But here technically I did not subclass it, so it really worked that way, just the linter was the problem.

grigi:
> > Then again both mypy and PyLint is complaining about the same thing. But not about `int`. Both are literals
> 
> https://stackoverflow.com/questions/10061752/which-classes-cannot-be-subclassed?lq=1
> Only bool is not subclassable, though I not 100% agree with the reasons. But here technically I did not subclass it, so it really worked that way, just the linter was the problem.

Thats an informative link. It makes sense to me that fake "classes" that have a small, finite set of global constants probably can't be subclassed, and leave the interpreter to perform well.

Also, I see the travis builds `mypy` is having a fit of epic proportions :-(
Run `make check test` locally and see if you can get that all to pass?

wallneradam:
Thanks. It was a lot of things to learn. I had not used mypy before this. It seems too rigorous to me. And definetely not designed for this kind of things. I tried to do this with less lying, but if I solved one thing it generated more other problems.

As I saw, you will drop Python 3.5 support. I think it could be (should be) refactored for Python 3.6 type annotations, and then will be less lying.
Most of my problems were with ForeignKeys, ManyToManyField and BackwardFKRelation, because they use `.type` in different meaning. They put Model class into it, not Field. And mypy cried because of this. It should really use different property name (Model?), but it would have been too much refactoring for this PR. Also `type` is not a good name, because it is reserved in Python. Now it can be a "private" property (also readonly), because it just used internally, so `_type` would be good name.

grigi:
It's always nice to figure something out :smile: 

Generally I have had to wrestle mypy less than I had to wrestle PyLint. Both have their own strong points.
Mostly I use Linters, because even with warts and all, they find bugs for much less effort than doing unit tests. And sometimes they bring attention to better ways of doing things.

Yes, for the next release we are dropping py3.5. the `develop` branch already has all that. The plan is to merge the last 3 fixes for master back into develop, and then clean up some more.

Then we have a release that has less features (no py3.5), and the only thing to gain is a slight perf improvement.
I feel like making it a big refactor release, mostly with the aim to simplify some of the complicated interactions. But that lies slow development, so may not work out as we expect. Still every now and then we need to pay our technical debts.

PR #203, opened 2019-10-01
Names ForeignKey constraints in a consistent way

grigi:
Also discovered that the `_make_hash(…)` method did not generate a cryptographically secure hash, and it was SUPER EASY to get conflicts.

Also since we now use the modern MySQL syntax for foreign keys, I started getting new errors, such as index length issues. I opted to change my testcases over raising an exception so as to not break code straight away.

I'm deferring hash & index changes to 0.14 as they could cause compatibility issues.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,9 @@
 
 Changelog
 =========
+0.13.10
+-------
+- Names ForeignKey constraints in a consistent way
 
 0.13.9
 ------
--- a/tests/models_schema_create.py
+++ b/tests/models_schema_create.py
@@ -7,7 +7,7 @@ from tortoise.models import Model
 
 class Tournament(Model):
     tid = fields.SmallIntField(pk=True)
-    name = fields.TextField(description="Tournament name", index=True)
+    name = fields.CharField(max_length=100, description="Tournament name", index=True)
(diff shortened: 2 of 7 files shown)


grigi:
@wallneradam This will create FK's like so:
```sql
CONSTRAINT `fk_event_tourname_51c2b82d` FOREIGN KEY (`tournament_id`) REFERENCES `tournament` (`tid`) ON DELETE CASCADE
```

Please review

wallneradam:
This was fast :-) It seems good to me, and works as expected in my application.

I think maybe you could use this new `hash2` in the `_generate_index_name` as well. Though index names should be unique in the table scope, so less chance for collisions.

Also maybe it would be a good idea to make it database specific later, so not using this old 30 character limit for MySQL (it has 64 chars limit).

grigi:
I decided against using hash2 for indexes in this version, as that may lead to creating duplicate hashes, and this is a "bugfix release".
But the next major version (develop branch) is definitely going to use hash2 everywhere.

Also, re the speed, we had each others attention. Things happen when people are paying attention :-)

grigi:
Also, a good idea to have the hash possibly be a string, that it then shortens in a deterministic way if needed. That way we can have some kind of heuristic and length limit for each dialect? :thinking: 

That would change the index names completely, but make them more human readable on average.

Issue #205, opened 2019-10-04
How to get related model from a ForeignKeyField?


mojimi:
I couldn't find any way to achieve this since when you access a ForeignKeyField property from a model class it just returns a property object.

The following is an example of what I'm trying to achieve :

```python
from tortoise import fields, Model

class Base(Model):

  @classmethod
  async def new(cls, **kwargs):
    create_params = {}
    for key, value in kwargs.items():
      if key in cls._meta.fields:
        create_params[key] = value
      elif key.endswith('_id'):
        key = key.replace('_id', '')
        if key in cls._meta.fields:
          model = getattr(cls, key).model #This method doesn't exist
          create_params[key] = model(id=value)

    return await cls.create(create_params)

class Book(Base):
  title = fields.CharField(100)

class Page(Base):
  book = fields.ForeignKeyField(Book, related_name='pages')
  content = fields.TextField()

#...

#The context here is a view that has book_id in the url parameters
#There's no need to query the database for said book beforehand
page = await Page.new(content='Hello World', book_id=1)
```

How can I achieve this?

Bonus points for being able to know if it's a forward or reverse relationship (single value or list)

grigi:
I'm not understanding why you have the new method?

e.g. with the following models:
```py
class Book(Model):
  title = fields.CharField(100)

class Page(Model):
  book = fields.ForeignKeyField(Book, related_name='pages')
  page = fields.IntField()
  content = fields.TextField()
```

You can create a page like so:
```py
page = await Page.create(content='Hello World', page=2, book_id=1)
```
You can get all pages like so (ordered):
```py
# Assume book exists
pages = await book.pages.order_by('page'))
```
To get a pages's book, you do this:
```py
# Assume page exists
await page.fetch_related('book')
book = page.book
```
The `fetch_related` is needed because `book` needs to be fetched, and until it is you only access the FK property.
The better syntax is to just do `book = await page.book` but that somehow didn't get implemented yet.

Also, look at the tests here: https://github.com/tortoise/tortoise-orm/blob/develop/tests/fields/test_fk.py

mojimi:
@grigi Sorry I wasn't clear enough.

I'm trying to create a "smarter" creation method that automatically maps a `field_id` parameter to a instance if `field` is a relationship field on the table.

So, both `Page.new(content='Hello World', book_id=1)` and `Page.new(content='Hello World', book_id=book)` would work.

[...]

PR #206, opened 2019-10-07
Field refactor

grigi:
Attempts to address #97 and #205 
Also, to extend the functionality of #39 

Done:
- [x] Split fields into many modules for categorisation
- [x] Remove unnesesary `__slots__` from Fields, as this complicates work, and saves almost no memory as Fields only get instantiated as class variables (hence does not accumulate)
- [x] PyLint plugin now uses simplified Field-type inference and generates reverse FK's correctly.
- [x] Move DDL definition of fields from `generate_schema` into Fields
- [x] Make fields support overriding on a per-dialect level
- [x] Extend schema discover to know about DB types for ANY db type.
- [x] Make schema generator use types specified in Fields
- [x] Have fields define generated PK's
- [x] Have fields define custom emulation rules (e.g. DecimalField and SQLite)

Deferred for part 2 later on:
- Have fields define "native level" for DB.
- Have fields define index types per field
- Have fields define collation per field


grigi:
@abondar Please review.

This PR _also_ exploded in scope, I have already culled the scope of it, but related stuff was added in.

We are not yet completely at the stage where schema_generators don't have to know anything about field types, generated integer primary keys are still in the schema_generators. (I have decided to call it quits as this needed to get out, so we can start shipping 0.14)

tortoise/fields/base.py
```diff
@@ -0,0 +1,103 @@
...
+                if not key.startswith("_")
+            }
+            for dialect in [key for key in dir(self) if key.startswith("_db_")]
+        }
+
+    def get_db_field_types(self) -> Optional[Dict[str, str]]:
+        if not self.has_db_field:
+            return None
+        return {
+            "": getattr(self, "SQL_TYPE"),
```
abondar: Why empty string is used here as key?

tortoise/backends/base/schema_generator.py
```diff
@@ -232,34 +208,27 @@ def _get_table_sql(self, model, safe=True) -> dict:
...
 
         # Indexes.
-        field_indexes_sqls = [
-            val
-            for val in [
-                self._get_index_sql(model, [field_name], safe=safe)
-                for field_name in fields_with_index
-            ]
-            if val
+        _indexes = [
```
abondar: May be info about index declared through `index=True` in field kwargs should be returned by field too? 
I'd like to make role of SchemaGenerator as small as possible, so it will be easier to build alternatives for it

tortoise/fields/base.py
grigi: The default? Should I use something else to indicate it isn't dialect specific?

tortoise/backends/base/schema_generator.py
grigi: Yes, I wanted to do that. Because for e.g. case-insensitive or full-text functionality, it can be implemented via collations and indexes on a per-field basis.
The PR was getting colossal enough as is (And I have a lack of time to spend on this), so I wondered if I can defer that to another PR, later?

Or we pull out the non-dependants for 0.14 and release that, and get this perfect first? But considering I spent nearly a month on this, I feel that is a way to delay everything indefinitely.

It is a call we should make, are we going to hold off on features until they are complete, or push partial features in? This is changing as the project matures, and we want to rather have fewer breakages. Are we there yet? 0.14 already introduces some breakages.

grigi:
@abondar I'm happy to pull non-field refactor changes out of here, and then release 0.14 as it has plenty of changes already. But then you need to help me in specifying what the refactor should cover in entirety so we can get that over the line?

The big things for v1.0 is:
* Field refactor
* Use Executor for all db operations (so we can manage things like atomic get_or_update and getting connections managed in a single place. As the one connection pool PR proved we need to do. This will also simplify the db clients, so they become much dumber.)
* Use parametrized queries everywhere (For queryset parameters, so `.filter()` and `.update()`, everything else is done.)
* Actually have reliable connection pooling

In roughly that order, I think.

grigi:
@abondar I took out all the non-field refactor work and release 0.14
Then cleaned up this PR, and increased the scope to move as much of the schema generator to the fields themselves.

Could you please have a look at the planned scope and/or alter it?

grigi:
Plan to remove the subclass hack, we found a less terrible way of doing the same by using annotated-functions-that-return-fields.
This can also lead to more accurate auto-completion that doesn't inaccurately show field parameters.

PR #208, opened 2019-10-09
Field multiple inheritence fix

wallneradam:
There were some small problems with my last PR. It can't work well with inherited fields have mixins.

## Description

```python
class VirtualField:
    def __init__(self, *args, **kwargs) -> None:
        super().__init__(*args, **kwargs)

class VirtualIntField(fields.IntField, VirtualField):
    pass
```

1. The _FieldMeta class hasn't checked if the 1st base class was Field. So if I used a field class in the above way, only IntField was preserved.
2. Field class hasn't called the `super().__init__`, so the `__init__` of 2nd base class of VirtualIntField was never reached.

## Motivation and Context
It is a fix for #202 

## How Has This Been Tested?
I tested with the above code

## Checklist:

### DIFF
--- a/tortoise/fields.py
+++ b/tortoise/fields.py
@@ -28,13 +28,12 @@ JSON_LOADS = json.loads
 
 class _FieldMeta(type):
     def __new__(mcs, name, bases, attrs):
-        if len(bases) > 1:
-            if bases[0] == Field:
-                # Instantiate class with only the 1st base class (should be Field)
-                cls = type.__new__(mcs, name, (bases[0],), attrs)  # type: Type[Field]
-                # All other base classes are our meta types, we store them in class attributes
-                cls.type = bases[1] if len(bases) == 2 else bases[1:]
-                return cls
+        if len(bases) > 1 and bases[0] is Field:
+            # Instantiate class with only the 1st base class (should be Field)
+            cls = type.__new__(mcs, name, (bases[0],), attrs)  # type: Type[Field]
+            # All other base classes are our meta types, we store them in class attributes
+            cls.type = bases[1] if len(bases) == 2 else bases[1:]
+            return cls
         return type.__new__(mcs, name, bases, attrs)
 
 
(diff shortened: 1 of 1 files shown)


tortoise/fields.py
```diff
@@ -29,11 +29,12 @@
 class _FieldMeta(type):
     def __new__(mcs, name, bases, attrs):
         if len(bases) > 1:
-            # Instantiate class with only the 1st base class (should be Field)
-            cls = type.__new__(mcs, name, (bases[0],), attrs)  # type: Type[Field]
-            # All other base classes are our meta types, we store them in class attributes
-            cls.type = bases[1] if len(bases) == 2 else bases[1:]
-            return cls
+            if bases[0] == Field:
```
grigi: What aboud:
```py
if len(bases) > 1 and bases[0] is Field:
```
A single conditional statement is preferable if it is possible. Also when comparing types, `is` will ensure that first subclass is a field, and not looks like a field.

wallneradam: Indeed, thanks :-)

grigi:
Thank you. I release v0.13.11 with this fix :+1:

tortoise/fields.py
```diff
@@ -89,6 +89,7 @@ def __init__(
         self.model = model
         self.reference = reference
         self.description = description
+        super().__init__(**kwargs)  # type: ignore # mypy issue 4335
```
grigi: This is causing issues, so I'm going to revert it. When removing it, and using your example I can't actually get it to not call the entire super chain.

wallneradam: @grigi The problem is with the Python's method resolution order. In my example VirtualField's `__init__` will never called. 

I thought my solution could be a problem, when no multiple inheritence was in use, but it worked for me and all tests were passed.

Now I read a lot about MRO and found that, this behavior is because python first go through the  1st base class's parent chain, then the 2nd, if there are no common classes in there. In my example only common class were `object`. So the solution was to extend my `VirtualField` class from `Field` and magically __init__ is called and everything works as expected.

So you could remove that commented line. Sorry about that...

grigi: Could you please do a documentation pr describing the caveat & process?
Maybe update against develop, as there is a lot of doc changes lined up for the next major release already

wallneradam: sure, next week

Issue #217, opened 2019-10-16
Add database functions in queries to support manipulation of fields


zoliszeredi:
**Is your feature request related to a problem? Please describe.**
As of [v0.13.2](https://github.com/tortoise/tortoise-orm/tree/0.13.12) there is no support for database functions. Depending on your use-case, query fields might require extra processing: 
 - for NULLs COALESCE for comparisons and conversions
 - for text types: TRIM for trailing white-spaces, LENGTH, REPLACE, SUBSTRING, CONCAT

**Describe the solution you'd like**
Add support for functions via an importable module (e.g. `tortoise.models.functions`)
that implements database functions (like COALESCE) and can wrap model fields possibly via annotations like:
```python
from tortoise.models.functions import Coalesce

class User(Model):
    id = fields.IntField(pk=True)
    name = fields.TextField()
    email = fields.TextField()
    username = fields.TextField()
    join_date = fields.DateTimeField()

users = await User.annotate(
    screen=Coalesce("name", "username", "email"),
    clean_join_date=Coalesce('join_date', datetime.datetime(2011, 12, 31),
).filter(
    clean_join_date__gt=datetime.datetime(2012, 1, 1)
).values_list('screen', 'clean_join_date')
```

**Describe alternatives you've considered**
The two most well known ORMs for python [SQLAlchemy](https://docs.sqlalchemy.org/en/13/core/functions.html?highlight=coalesce#sqlalchemy.sql.functions.coalesce) and [Django ORM](https://docs.djangoproject.com/en/2.2/ref/models/database-functions/#coalesce) have similar APIs:

**Additional context**
The underpinning SQL library pypika [does implement](https://github.com/kayak/pypika/blob/v0.8.0/pypika/functions.py) and have [documented](https://pypika.readthedocs.io/en/latest/api/pypika.functions.html?highlight=coalesce#pypika.functions.Coalesce) the most important database functions.

From the pypika [documentation](https://github.com/kayak/pypika#strings-functions) for example of string functions:

```python
from pypika import functions as fn

customers = Tables('customers')
q = Query.from_(customers).select(
    customers.id,
    fn.Concat(customers.fname, ' ', customers.lname).as_('full_name'),
)
```

grigi:
Thank you for this well presented/thought out feature request @zoliszeredi 

We already have support for some aggregate functions, and I don't see much new required to support other functions. 
(Although the name "Aggregate" is too specific, we should probably change it to "Function")

[...]

PR #218, opened 2019-10-17
Add trim, length, coalesce db functions.

zoliszeredi:
Add support for filtering via database functions: trim, string-length and coalesce with a non-field type default 

## Description
Solution is implemented by sub-classing the [Aggregate](https://github.com/tortoise/tortoise-orm/blob/0.13.12/tortoise/aggregation.py#L12) class and mapping the appropriate pypika [functions](https://github.com/kayak/pypika/blob/0.9.3/pypika/functions.py#L67). 

### Limitations
In this implementation the second argument in `COALESCE` will be considered a simple value based default and will not resolved to a field. 
This limitation impacts implementation for other functions like `CONCAT` can have many field terms.

## Motivation and Context
Resolves support for #217 

## How Has This Been Tested?
Unit tests

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Enhancements:
 - Models, Fields & QuerySets have significant type annotation improvements,
   leading to better IDE integration and more comprehensive static analysis.
 - Fetching records from the DB is now up to 25% faster.
+- Database functions added in aggregations ``Trim()``, ``Length()``, ``Coalesce()``.
 
 Bugfixes:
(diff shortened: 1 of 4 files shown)


grigi:
Hi @zoliszeredi Thank you for this. It seems even easier than I expected.

Could you please add yourself to CONTRIBUTORS.rst & add to the changelog (under 0.14.0).

grigi:
Basically looks good to me :+1:
Thank you for the work.

PR #234, opened 2019-11-03
Safe Concurrent Transactions

grigi:
This PR should hopefully finally resolve the `ContextVars` confusion that we had previously.
Tests so far shows that this is keeping track of transactions and handles locks between them. :tada: 

## This is done by:
* Having transactions wrapped in a transaction contextmanager
* Rewriting the test layer to be more async so we can run the entire per-test transaction in one event-loop run, 
* Differentiating between root-transaction locks and operation locks.

This also means that `start_transaction` will probably never work, so we need to mark it as deprecated, and/or aggressively remove it.
Likely we will have to bump to 0.15 once we get this in. And preferably working connection pooling.

## To go multi-connection-pool: 
* We need to have a per-connection root-transaction and operation lock.
* When creating a transaction the `TransactionWrapper` will probably not change much, just get the connection-specific handle and locks.
* Root-level autocommitted operations need to pull off an available connection, so we would need to make the root level `acquire_connection` also select from the pool, but the `TransactionWrapper` should keep the current naïve `ConnectionWrapper`.

## Done:
* [x] Use context managers to manage the ContextVar
* [x] Update test runner to use this 
* [x] Get all tests to pass on SQLite
* [x] Get all tests to pass on PostgreSQL
* [x] Get all tests to pass on MySQL+MyISAM
* [x] Get all tests to pass on MySQL+InnoDB
* [x] Mark `start_transaction` as deprecated
* [x] Remove all `HANGS` skips
* [x] Remove all `BREAKS TESTRUNNER` skips
* [x] Fix Py3.6

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ _testall:
 	$(py_warn) TORTOISE_TEST_DB=sqlite://:memory: py.test --cov-report=
 	python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:$(TORTOISE_POSTGRES_PASS)@127.0.0.1:5432/test_\{\} py.test --cov-append --cov-report=
 	$(py_warn) TORTOISE_TEST_DB="mysql://root:$(TORTOISE_MYSQL_PASS)@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test --cov-append --cov-report=
-	$(py_warn) TORTOISE_TEST_DB="mysql:$(TORTOISE_MYSQL_PASS)//root:@127.0.0.1:3306/test_\{\}" py.test --cov-append
+	$(py_warn) TORTOISE_TEST_DB="mysql://root:$(TORTOISE_MYSQL_PASS)@127.0.0.1:3306/test_\{\}" py.test --cov-append
 
 testall: deps _testall
 
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -19,20 +19,3 @@ class TestBasic(test.TestCase):
                 {"id": tournament.id + 1, "name": "Test 2"},
             ],
         )
-
-    async def test_basic2(self):
-        tournament = await Tournament.create(name="Test")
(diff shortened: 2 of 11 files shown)


grigi:
@abondar @AEnterprise Please have  look at this PR.
I'm pretty confident this solves the context leak issues :grin:
And does this with little API change, so we can hopefully impact very few people with this.

There is a context leak when doing a gather over multiple transactions in Py3.6, but I could not figure out why. It could be an incompatibility with the `aiocontextvars` backport for Py3.6. Everything else is working perfectly as far as I can see. We may have to have a compatibility declaration in the docs.

So, hopefully as long as the proposed guidance has been followed we can easily enable pooling.

abondar:
It's really great to take that headache away 👍 
But I can't get why trxlock is necessary 
Could you explain it for me please?

grigi:
To guarantee consistency we need to ensure that for the same transaction (so one connection) that DB operations get serialised, else the DB gets very confused fast. And then we need to guard against different transactions.
Hence we need two locks during a transaction.

I spoke with AEnterprise on gitter some more, and there is some further discussion there.

A summary is there is 2 "modes":
* Autocommit: trx lock is N/A, operation lock is the get-pool
* Transactioned: trx lock is the get-pool, operation lock is a manual asyncio.lock

But after discussion this PR still has a race condition:
One task starts a transaction (but not yet do an operation), and then another task could do an autocommit operation, then that one uses the operation-lock (which is not set) and accidentally inserts an instruction into a different tasks transaction.

So I'll have to fix this.

grigi:
Or rather, we need to guarantee that a task that is sending a request to the db is the only one that reads the result. And that no other request is sent to clobber the cursor.
And since this is a fully async system, why can't I do a gather operation inside an atomic block?

grigi:
@abondar @AEnterprise This last commit should fix the potential transaction/autocommit pollution issue mentioned.

Also, the tortoise benchmarks had a single test that behaved a bit wonky when scaling concurrency up, but I can't get it to misbehave now :tada:

Issue #71, opened 2018-12-11
[IDEA] Alternate model syntax


grigi:
In trying to solve the DB → python type mapping, and the advent of `typing`, a possible solution is to define types that then get evaluated into both Python and DB types:

This also raises the question of a cleaner (more modern) model definition, by defining the type:

Since py3.6 we have the typed variables:
```py3
class Author(Model):
    name: str[255]
    books: reverse_fk[Book]

class Book(Model):
    title: str[255]
    author: Author
    published: datetime
    logo: bytes = None
    pages: int
    extra: JSONField = {}
    price: Decimal[18,5]
    created_at: datetime[auto_now_add]
```
Which is already standard in Py3.7 [@dataclass](https://docs.python.org/3/library/dataclasses.html)

We could extend type definitions so we can build model definitions off of types.
e.g. `str[100] = 'some default'` → `CharField(max_length=100, default='some default')`
The syntax is not comprehensive, but we could probably use type declarations for a large portion of use cases.

This would require Py3.6+ to function, so dropping Py3.5 & PyPy.

abondar:
Even though possibility is neat, I don't see any real life advantages of implementing this. 

It would be harder for us to implement both syntaxes and development of standard syntax could be held down because of type annotation limitations.

And for users, having two ways to do same thing would be confusing and will most likely lead to wrong understanding of how to use Tortoise.

grigi:
Yes, hence the "wild idea".

To get IDE's to understand how models are configured we would need to mess with `typing` in any case.
Here we define python types that get mapped to DB types (Which I see how to do, sort of), our existing syntax defines DB types, that we should then map to Python types.

The big issue is they don't map cleanly. `str` is unbounded so technically `TextField`, but in reality it is more often a `CharField`.

We need a place to put wild ideas. An issue just doesn't feel right, but I don't see where else?

abondar:
I think issues are fine for it, but we should formulate it so, that wild ideas should come with description of a problem that we are trying to address with this wild idea. So that way, discussion on how to overcome that problem would be in place and alternative solution for that problem could be found out in issue.

Regarding typing, at the moment I don't see how we can change our types for models because of static nature of `typing` module of python, while metaclasses work dynamically.
Django overcomes it with custom plugins for editors, but for Tortoise at the moment that seems like drastic overkill.

grigi:
[...]

PR #235, opened 2019-11-04
added type hints for relation managers

nsidnev:
## Description

With these changes, it is now possible to achieve better auto completion in IDEs when developing something using `Tortoise ORM`.

## Motivation and Context

I'm not sure if this change looks like the way you intended using `Tortoise ORM`, but it seems to me that it would be useful in this project.

I'm also not sure if any additional documentation changes are required for this PR.

Version `0.14.0` brought us a nice improvement in typing in working with models. It works well with direct access to referenced models or query results, but there is a problem when you want to access from a related model, like here:
```python3


class ModelA(models.Model):
    b = fields.ForeignKeyField("models.ModelB", related_name="a_models")


class ModelB(models.Model):
    name = fields.TextField(default="")


async def main():
    b = await ModelB.create()
    await ModelA.create(b=b)

    # normal typing
    await ModelA.get(b=b)  # will be ModelA
    await ModelA.all()  # will be List[ModelA]

    # bad typing
    await b.a_models.filter()  # error in PyCharm "Unresolved attribute reference 'a_models' for class 'ModelB'"
```

I use `tortoise` in some of my projects, and I think it will be very useful to have an auto completion in the IDE even with some manual work on adding this typing to models. Previously, I just used `fields.RelationQueryContainer` and `fields.ManyToManyRelationManager` directly, along with typing the results of queries, like this:
```python3
from typing import List

class ModelA(models.Model):
    b = fields.ForeignKeyField("models.ModelB", related_name="a_models")


class ModelB(models.Model):
    name = fields.TextField(default="")

    a_models: fields.RelationQueryContainer


async def main():
    b = await ModelB.create()
    await ModelA.create(b=b)

    # normal typing
    await ModelA.get(b=b)  # will be ModelA
    await ModelA.all()  # will be List[ModelA]

    # bad typing
    a_models: List[ModelA] = await b.a_models.filter()  # now there is no attribute error in PyCharm, but no auto completion on results
```

With this change, it’s now possible to write something like this and achieve normal auto completion in models.
```python3


class ModelA(models.Model):
    b = fields.ForeignKeyField("models.ModelB", related_name="a_models")


class ModelB(models.Model):
    name = fields.TextField(default="")

    a_models: fields.RelationQueryContainer[ModelA]


async def main():
    b = await ModelB.create()
    await ModelA.create(b=b)

    # normal typing
    await ModelA.get(b=b)  # will be ModelA
    await ModelA.all()  # will be List[ModelA]

    # also good typing
    await b.a_models.filter()  # will be List[ModelA]
```

I also added a `__await__` method for the `fields.ForeignKeyField`. It is required when used with the awaiting for the field with the possibility from `0.14.0`. Without this, my IDE (PyCharm) complains that `Class 'ForeignKeyField' does not define '__await__'; so the 'await' operator cannot be used on its instances`.

## How Has This Been Tested?

Since this change is not in the logic of `Tortoise ORM`, but in the improvement of typing for editors, testing was done manually in PyCharm.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ test: deps
 
 _testall:
 	$(py_warn) TORTOISE_TEST_DB=sqlite://:memory: py.test --cov-report=
-	python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:$(TORTOISE_POSTGRESS_PASS)@127.0.0.1:5432/test_\{\} py.test --cov-append --cov-report=
+	python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:$(TORTOISE_POSTGRES_PASS)@127.0.0.1:5432/test_\{\} py.test --cov-append --cov-report=
 	$(py_warn) TORTOISE_TEST_DB="mysql://root:$(TORTOISE_MYSQL_PASS)@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test --cov-append --cov-report=
 	$(py_warn) TORTOISE_TEST_DB="mysql:$(TORTOISE_MYSQL_PASS)//root:@127.0.0.1:3306/test_\{\}" py.test --cov-append
 
--- a/Makefile
+++ b/Makefile
@@ -51,9 +51,9 @@ test: deps
 
 _testall:
 	$(py_warn) TORTOISE_TEST_DB=sqlite://:memory: py.test --cov-report=
-	python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:@127.0.0.1:5432/test_\{\} py.test --cov-append --cov-report=
-	$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test --cov-append --cov-report=
-	$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}" py.test --cov-append
+	python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:$(TORTOISE_POSTGRESS_PASS)@127.0.0.1:5432/test_\{\} py.test --cov-append --cov-report=
+	$(py_warn) TORTOISE_TEST_DB="mysql://root:$(TORTOISE_MYSQL_PASS)@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test --cov-append --cov-report=
+	$(py_warn) TORTOISE_TEST_DB="mysql:$(TORTOISE_MYSQL_PASS)//root:@127.0.0.1:3306/test_\{\}" py.test --cov-append
 
 testall: deps _testall
 
--- a/tests/model_bad_rel1.py
+++ b/tests/model_bad_rel1.py
@@ -7,7 +7,10 @@ from tortoise.models import Model
 
 class Tournament(Model):
     id = fields.IntField(pk=True)
+    events: fields.RelationQueryContainer["Event"]
 
 
 class Event(Model):
-    tournament = fields.ForeignKeyField("app.Tournament", related_name="events")
+    tournament: fields.ForeignKey[Tournament] = fields.ForeignKeyField(
+        "app.Tournament", related_name="events"
+    )
--- a/tests/model_bad_rel1.py
+++ b/tests/model_bad_rel1.py
@@ -7,10 +7,7 @@ from tortoise.models import Model
 
 class Tournament(Model):
     id = fields.IntField(pk=True)
-    events: fields.RelationQueryContainer["Event"]
 
 
 class Event(Model):
-    tournament: fields.ForeignKey[Tournament] = fields.ForeignKeyField(
-        "app.Tournament", related_name="events"
-    )
+    tournament = fields.ForeignKeyField("app.Tournament", related_name="events")
--- a/tests/testmodels.py
+++ b/tests/testmodels.py
@@ -21,6 +21,10 @@ class Tournament(Model):
     desc = fields.TextField(null=True)
     created = fields.DatetimeField(auto_now_add=True, index=True)
 
+    events: fields.RelationQueryContainer["Event"]
+    minrelations: fields.RelationQueryContainer["MinRelation"]
+    uniquetogetherfieldswithfks: fields.RelationQueryContainer["UniqueTogetherFieldsWithFK"]
+
     def __str__(self):
         return self.name
 
@@ -28,6 +32,8 @@ class Tournament(Model):
 class Reporter(Model):
     id = fields.IntField(pk=True)
     name = fields.TextField()
+
+    events: fields.RelationQueryContainer["Event"]
 
     class Meta:
         table = "re_port_er"
@@ -39,9 +45,13 @@ class Reporter(Model):
 class Event(Model):
     id = fields.BigIntField(pk=True)
     name = fields.TextField()
-    tournament = fields.ForeignKeyField("models.Tournament", related_name="events")
-    reporter = fields.ForeignKeyField("models.Reporter", null=True)
-    participants = fields.ManyToManyField(
+    tournament: fields.ForeignKey["Tournament"] = fields.ForeignKeyField(
+        "models.Tournament", related_name="events"
+    )
+    reporter: fields.ForeignKeyNullable[Reporter] = fields.ForeignKeyField(
+        "models.Reporter", null=True
+    )
+    participants: fields.ManyToManyRelationManager["Team"] = fields.ManyToManyField(
         "models.Team", related_name="events", through="event_team", backward_key="idEvent"
     )
     modified = fields.DatetimeField(auto_now=True)
@@ -55,6 +65,9 @@ class Team(Model):
     id = fields.IntField(pk=True)
     name = fields.TextField()
 
+    events: fields.ManyToManyRelationManager[Event]
+    minrelation_through: fields.ManyToManyRelationManager["MinRelation"]
+
     def __str__(self):
         return self.name
 
@@ -64,7 +77,9 @@ class EventTwo(Model):
     name = fields.TextField()
     tournament_id = fields.IntField()
     # Here we make link to events.Team, not models.Team
-    participants = fields.ManyToManyField("events.TeamTwo")
+    participants: fields.ManyToManyRelationManager["TeamTwo"] = fields.ManyToManyField(
+        "events.TeamTwo"
+    )
 
     class Meta:
         app = "events"
@@ -76,6 +91,8 @@ class EventTwo(Model):
 class TeamTwo(Model):
     id = fields.IntField(pk=True)
     name = fields.TextField()
+
+    eventtwo_through: fields.ManyToManyRelationManager[EventTwo]
 
     class Meta:
         app = "events"
@@ -169,19 +186,23 @@ class UUIDFields(Model):
 
 class MinRelation(Model):
     id = fields.IntField(pk=True)
-    tournament = fields.ForeignKeyField("models.Tournament")
-    participants = fields.ManyToManyField("models.Team")
+    tournament: fields.ForeignKey[Tournament] = fields.ForeignKeyField("models.Tournament")
+    participants: fields.ManyToManyRelationManager[Team] = fields.ManyToManyField("models.Team")
 
 
 class M2MOne(Model):
     id = fields.IntField(pk=True)
     name = fields.CharField(max_length=255, null=True)
-    two = fields.ManyToManyField("models.M2MTwo", related_name="one")
+    two: fields.ManyToManyRelationManager["M2MTwo"] = fields.ManyToManyField(
+        "models.M2MTwo", related_name="one"
+    )
 
 
 class M2MTwo(Model):
     id = fields.IntField(pk=True)
     name = fields.CharField(max_length=255, null=True)
+
+    one: fields.ManyToManyRelationManager[M2MOne]
 
 
 class NoID(Model):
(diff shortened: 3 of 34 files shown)


grigi:
This is pretty awesome! Too bad there isn't an obvious/generic/easy way to let the editor/mypy know about `a_models` without having to define the reverse. _(We could do a mypy plugin, but that requires manual setup for mypy, and doesn't do anything for IDEs that I know of.)_

I think this is a reasonable pattern to encourage.

We need to add this into the unit tests, and actually get it used to ensure that this pattern doesn't break things.
Also e.g. in the future I might want to make models autogenerate `__slots__` (for memory usage reasons) and this pattern could interfere. So we need to make sure that we cover this. In fact this may be a way to solve the issue that we don't know reverse relations at class creation time, and hence generating `__slots__` breaks things because the address space of a slotted class is immutable.

Also, I think this pattern could be described in a section under `Models` reference called something like "Improving relational type hinting". :smile: 

I'm thinking we would have to create a type describing the actions for a ForeignKey, right now it is something like:
```py
Union[Awaitable[Optional[ForeignModel]], Optional[ForeignModel]]
```
Depending on whether `select_related` has been used.

e.g.
```py
class StraightFields(Model):
    eyedee = fields.IntField(pk=True, description="Da PK")  # will be 'int'
    chars = fields.CharField(max_length=50, index=True, description="Some chars")  # will be 'str'
    blip = fields.CharField(max_length=50, default="BLIP")  # will be 'str'
    fk: fields.RelationQueryContainer["StraightFields"] = fields.ForeignKeyField(
        "models.StraightFields", related_name="fkrev", null=True, description="Tree!"
    )
    rel_to: fields.ManyToManyRelationManager["StraightFields"] = fields.ManyToManyField(
        "models.StraightFields", related_name="rel_from", description="M2M to myself"
    )

    fkrev: fields.BackwardFKRelation["StraightFields"]
    rel_from: fields.ManyToManyRelationManager["StraightFields"]
```

:thinking:
Is it possible we could add more "behaviour" items to FK/M2M fields instead?
Is it possible to do something like
```py
fk = fields.ForeignKeyField[ForeignModel]("models.ForeignModel", ...)
```
?

nsidnev:
> This is pretty awesome!

Thanks! :smile: 

> Too bad there isn't an obvious/generic/easy way to let the editor/mypy know about `a_models` without having to define the reverse.

I'm not sure about *easy* way, but I think that plugin for mypy will be very useful for checking projects in CI. As for the editors, I think there might be plugins for them created by community of these project in future. Both of these ways are already used in the pydantic, and I think that they can be used here as well.

> We need to add this into the unit tests, and actually get it used to ensure that this pattern doesn't break things.

> Also, I think this pattern could be described in a section under `Models` reference called something like "Improving relational type hinting".

Yep, I can handle it.

> Also e.g. in the future I might want to make models autogenerate `__slots__` (for memory usage reasons) and this pattern could interfere.

Well, I don’t really understand how slots can break annotations. Even if we define slots for some classes, annotations are stored independently as a class attribute, while slots control instance attributes. But in any case, it will be possible to control this through modified tests, where this pattern will be used, right?

> I'm thinking we would have to create a type describing the actions for a ForeignKey, right now it is something like:
> 
> ```python
> Union[Awaitable[Optional[ForeignModel]], Optional[ForeignModel]]
> ```

Yes, I think it would be quite suitable for use. But I'm not sure where to put it.

> ```python3
> class StraightFields(Model):
>     ...
>     fk: fields.RelationQueryContainer["StraightFields"] = fields.ForeignKeyField(
>         "models.StraightFields", related_name="fkrev", null=True, description="Tree!"
>     )
>     rel_to: fields.ManyToManyRelationManager["StraightFields"] = fields.ManyToManyField(
>         "models.StraightFields", related_name="rel_from", description="M2M to myself"
>     )
> 
>     fkrev: fields.BackwardFKRelation["StraightFields"]
>     rel_from: fields.ManyToManyRelationManager["StraightFields"]
> ```

Could you clarify the point here, why the annotation for `fk` is `fields.RelationQueryContainer`, and for `fkrev` `fields.BackwardFKRelation`? I could not find the normal annotation for `fk` (what you suggested about the FK annotation seems to me to be perfect here), but shouldn’t it just be `fields.RelationQueryContainer` for `fkrev`? Or am I missing something?

> Is it possible we could add more "behaviour" items to FK/M2M fields instead?
> Is it possible to do something like
> 
> ```python
> fk = fields.ForeignKeyField[ForeignModel]("models.ForeignModel", ...)
> ```

Right now I think that it's impossible because type annotations won't work like that. But it seems to me that this will be possible with the implementation of #71.

grigi:
> I'm not sure about _easy_ way, but I think that plugin for mypy will be very useful for checking projects in CI. As for the editors, I think there might be plugins for them created by community of these project in future. Both of these ways are already used in the pydantic, and I think that they can be used here as well.

I would like to pursue both in the future as well.

> Well, I don’t really understand how slots can break annotations. Even if we define slots for some classes, annotations are stored independently as a class attribute, while slots control instance attributes. But in any case, it will be possible to control this through modified tests, where this pattern will be used, right?

`__slots__` changes interaction between class variables and instance variables, without it class varaibles can get overwritten, with it, you get read-only errors.
Yes, the purpose of tests is to control your code :stuck_out_tongue:

> Could you clarify the point here, why the annotation for `fk` is `fields.RelationQueryContainer`, and for `fkrev` `fields.BackwardFKRelation`? I could not find the normal annotation for `fk` (what you suggested about the FK annotation seems to me to be perfect here), but shouldn’t it just be `fields.RelationQueryContainer` for `fkrev`? Or am I missing something?

:man_facepalming:  I made a few mistakes here (a copy-paste error), I meant something like this:
```py
class StraightFields(Model):
    ...
    fk: fields.ForeignKeyType["StraightFields"] = fields.ForeignKeyField(
        "models.StraightFields", related_name="fkrev", null=True, description="Tree!"
    )
    rel_to: fields.ManyToManyRelationManager["StraightFields"] = fields.ManyToManyField(
        "models.StraightFields", related_name="rel_from", description="M2M to myself"
    )

    fkrev: fields.RelationQueryContainer["StraightFields"]
    rel_from: fields.ManyToManyRelationManager["StraightFields"]
```
where `ForeignKeyType` could be defined as (in fields.py):
```py
ForeignKeyType = Union[Awaitable[Optional[MODEL]], Optional[MODEL]]
```
But there are nullable and non-nullable FK's so may be best to define it twice?

> Right now I think that it's impossible because type annotations won't work like that. But it seems to me that this will be possible with the implementation of #71.

Type annotation syntax doesn't support extra metadata, but this is solved in libraries like attrs/pydantic by defining a type and assigning a metadata object to it. As this is already a pattern, it might be OK to just follow this?

nsidnev:
@grigi 
Just ran into the problem of understanding annotations using `mypy`. The `fields.ManyToManyField` is not a `fields.ManyToManyRelationManager` and `mypy` tells us about that. I think that it might be possible to rewrite M2M fields as a function that will return relation manager by annotation? This way is used in `fastapi` for solving mypy issues.

grigi:
> Just ran into the problem of understanding annotations using `mypy`. The `fields.ManyToManyField` is not a `fields.ManyToManyRelationManager` and `mypy` tells us about that.

Technically, it is correct. But we rewrite the classes, so things are not as they appear.

> Would it be possible to rewrite the M2M field as a function that will return relation manager by annotation? This way is used in `fastapi` for solving mypy issues.

You can do that? :thinking: You mean a function that returns an annotation of `fields.ManyToManyRelationManager` but instead returns a class that we just `# type: ignore`?

Hmm, can we get it to do generics? that way we could potentially change the signature to accept a class (or string) as the type specialization? Then we can completely eliminate the need to require specifying types. and also get rid of the odd typing hack we are doing for data fields...
Would be even better if we could generalise the `Optional` part...

Ok, I'm getting a bit carried away here. But that opens up possibilities :smile:

nsidnev:
> You can do that? You mean a function that returns an annotation of `fields.ManyToManyRelationManager` but instead returns a class that we just `# type: ignore`?

Yes, I mean exactly that. It can look a little strange, but I think it might be the appropriate way here. This won't break backward compatibility (or at least I think so) and give us a normal typing (as far as possible now) without mypy errors.

As for the implementation, I think that the `fields.ManyToManyField` itself can be renamed to something like `fields.ManyToManyFieldInstance`. And then write function like this:
```python3
def ManyToManyField(
    model_name: str,
    through: Optional[str] = None,
    forward_key: Optional[str] = None,
    backward_key: str = "",
    related_name: str = "",
    field_type: "Type[Model]" = None,  # type: ignore
    **kwargs,
) -> ManyToManyRelationManager:
    return ManyToManyFieldInstance(  # type: ignore
        model_name, through, forward_key, backward_key, related_name, field_type, **kwargs
    )
```
Can it break something?

grigi:
`__init__` `models` `filters` `query_utils` & `executor` all test against it. So we will have to update many places. But other than that, I don't think it will break anything.

tests/model_bad_rel1.py
```diff
@@ -7,7 +7,10 @@
 
 class Tournament(Model):
     id = fields.IntField(pk=True)
+    events: fields.RelationQueryContainer["Event"]
```
grigi: We really needn't change these `model_*` models as they are faulty on purpose for testing.

tests/testmodels.py
```diff
@@ -240,36 +261,50 @@ class ImplicitPkModel(Model):
 class UUIDPkModel(Model):
     id = fields.UUIDField(pk=True)
 
+    children: fields.RelationQueryContainer["UUIDFkRelatedModel"]
+    children_null: fields.RelationQueryContainer["UUIDFkRelatedNullModel"]
+    peers: fields.ManyToManyRelationManager["UUIDM2MRelatedModel"]
```
grigi: You definitely needn't be so thorough here in the test models. We probably will have to deal with a crazy mix of complete and incomplete annotations.

I would take it out of the UUIDPkModel and its related, but leave it in the UUIDPkSourceModel and related. (or vice versa)
Reason for this that these two sets of models are run by the exact same test cases each, and we want to ensure that the differences in specifying models don't affect actual API usage at all.

tortoise/fields.py
```diff
@@ -517,6 +528,20 @@ def __init__(
         self._generated = False
 
 
+def ManyToManyField(
+    model_name: str,
+    through: Optional[str] = None,
+    forward_key: Optional[str] = None,
+    backward_key: str = "",
+    related_name: str = "",
+    field_type: "Type[Model]" = None,  # type: ignore
```
grigi: Technically the `field_type` is an internal API, and only ever set either later or during `__init__` (If I remember correctly). So we can now due to the separation of interface split up the public/private API quite well.

grigi:
Also, you don't need to be so thorough in updating the tests, Or rather it may actually be desirable to NOT be consistent, as we can't expect users to be consistent.
I mean, most projects have many different people working on it, and a non-uniform test suite is more likely to pick up on edge-case regressions.

Still, awesome work. It does open up avenues for even better typing for free for most projects :-)

tests/model_bad_rel1.py
nsidnev: No problems, I will return them to their original state except for the models from `testmodels`.

tests/testmodels.py
nsidnev: Got it. I left only `UUIDPkModel`.

tortoise/fields.py
nsidnev: Just removed field_type and ran the tests. All tests were successful!

nsidnev:
> What about `ManyToMany` and `ReverseForeignKey` as types?

`ReverseForeignKey` more leads me to think about the possibility of getting a single model instance, rather than related models as collection. And it seems to me that this name would be more appropriate if Tortoise supported `OneToOneField`. It seems to me more appropriate for FK to be something like `BackwardRelations`/`BackwardRelationship`, what do you think about it?

Something similar, I think, will be appropriate for M2M(`ManyToManyRelationship`), but `ManyToMany` also seems to me to be one of the suitable names.

grigi:
But with a relationquerycontainer you only get one result?
Still, it is the same type on both sides of a one to one relationship, so SingleRelation may be a better name?
As long as it's consistent with the "theme".

nsidnev:
In the end, I renamed them as follows:
1. `RelationQueryContainer` -> `ReverseRelation`.
2. `ManyToManyRelationManager` -> `ManyToManyRelation`.

Then, if Tortoise will ever add `OneToOneField` type for a relation, it can be named (as you suggested) the same way as `SingleRelation`.

If you do not like this naming, then I can change it to what you suggested.

If all is well, then it remains to document the changes of this PR.

grigi:
For the sake of consistency, could you rename `ForeignKey` to `ForeignKeyRelation` ?

grigi:
Looks pretty good. Could you change the branch you want to merge to, to `master`. The current `develop` branch is getting some large changes, that may need stabilization before we can release. So I don't want to delay this more than we need to.

Makefile
```diff
@@ -51,9 +51,9 @@ test: deps
 
 _testall:
 	$(py_warn) TORTOISE_TEST_DB=sqlite://:memory: py.test --cov-report=
-	python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:@127.0.0.1:5432/test_\{\} py.test --cov-append --cov-report=
-	$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test --cov-append --cov-report=
-	$(py_warn) TORTOISE_TEST_DB="mysql://root:@127.0.0.1:3306/test_\{\}" py.test --cov-append
+	python -V | grep PyPy || $(py_warn) TORTOISE_TEST_DB=postgres://postgres:$(TORTOISE_POSTGRES_PASS)@127.0.0.1:5432/test_\{\} py.test --cov-append --cov-report=
+	$(py_warn) TORTOISE_TEST_DB="mysql://root:$(TORTOISE_MYSQL_PASS)@127.0.0.1:3306/test_\{\}?storage_engine=MYISAM" py.test --cov-append --cov-report=
+	$(py_warn) TORTOISE_TEST_DB="mysql:$(TORTOISE_MYSQL_PASS)//root:@127.0.0.1:3306/test_\{\}" py.test --cov-append
```
grigi: typo, look at `$(TORTOISE_MYSQL_PASS)` placement.

grigi:
Looks pretty good, just the one comment about a change in the makefile. (Oh, could be related to master/develop change)

grigi:
Since you didn't break the makefile (it was broken) I think I'll just merge this :-)

Thank you for your great work!

PR #239, opened 2019-11-16
OneToOneField

sinaso:
I implemented most of what is required to make OneToOneField working both in prefetch and lazy mode. I have tested most cases, although there is no unit tests yet. Just wanted you to take a look and comment. It may not be completely ready to be merged yet though.

thanks
--sina


tortoise/backends/base/executor.py
```diff
@@ -318,15 +345,23 @@ def _make_prefetch_queries(self) -> None:
...
     async def _execute_prefetch_queries(self, instance_list: list) -> list:
         if instance_list and (self.prefetch_map or self._prefetch_queries):
             self._make_prefetch_queries()
-            for field, related_query in self._prefetch_queries.items():
-                await self._do_prefetch(instance_list, field, related_query)
+            prefetch_tasks = [
+                self._do_prefetch(instance_list, field, related_query)
+                for field, related_query in self._prefetch_queries.items()
+            ]
+            await asyncio.gather(*prefetch_tasks)
```
grigi: :+1: 
Yes, since we are now properly concurrency safe, we should probably do more I/O concurrently instead of in loops

tortoise/__init__.py
```diff
@@ -158,14 +158,21 @@ def describe_field(name: str) -> dict:
...
             # These fields are entierly "virtual", so no direct DB representation
-            if isinstance(field, (fields.ManyToManyFieldInstance, fields.BackwardFKRelation)):
+            if isinstance(
+                field,
+                (
+                    fields.ManyToManyFieldInstance,
+                    fields.BackwardFKRelation,
+                    fields.BackwardOneToOneRelation,
+                ),
+            ):
```
grigi: Should probably extend the field definition by addind a "o2o_fields" list instead. o2o may be implemented much like foreign keys, but are conceptually different. So lets keep them separate in `describe_model`

tortoise/backends/base/schema_generator.py
```diff
@@ -181,11 +182,12 @@ def _get_table_sql(self, model, safe=True) -> dict:
                 else ""
             )
             # TODO: PK generation needs to move out of schema generator.
-            if field_object.pk:
+            if field_object.pk and not isinstance(field_object.reference, OneToOneField):
```
grigi: This is OK for now, but we should remember to fix this in #206

sinaso:
Thank you @grigi . I will try unit tests, will let you know if help was needed.

sinaso:
@grigi please take a look. Added test to cover the added code.

grigi:
I'll have a look today :-)

grigi:
I have a question, form a user's pov, is there such a thing as a backward One-to-One field? The behaviour should be identical to forward foreign key, right?

As in `await model.o2ofield` and `await othermodel.o2orevfield` ?

grigi:
I have a question, form a user's pov, is there such a thing as a backward One-to-One field? The behaviour should be identical to forward foreign key, right?

As in `await model.o2ofield` and `await othermodel.o2orevfield` ?

grigi:
Beautiful PR :+1: 

Could I ask that you add yourself to CONTRIBUTORS and a quick description of the change in CHANGELOG? (e.g. section 0.15.2)

The only change I could request is to merge "o2o_fields" & "backward_o2o_fields" in the describe model. As the user would use them symmetrically. Kind of like how m2M is used symmetrically.

One can still see by the type of "Backward..." how the one is implemented if you really NEED to know.

Otherwise: Awesome work! :sunglasses:

sinaso:
Thanks @grigi !

I can merge "o2o_fields" & "backward_o2o_fields" but before that, let's make sure if it is not going to be confusing down the road. Note that, contrary to M2M case, our fields on O2O are not really symmetric.

For one, in the underlying tables, backward table does not have any columns while the forward table has a column pointing to the other table. This asymmetry is also reflected in the models themselves, setting the o2o value in the forward object, means we are changing the associated column in the table, while changing the backward o2o value should be avoided (hmm, I still had to put a setter for the lazy case, but now I am thinking I need to disallow users to set the backward relation)

my guess is because of this asymmetry pretty much in all cases users might need to distinguish between forward and backward o2o even if we merge them into one.

I would be happy to discuss this more if needed, but if you feel confident, I can go ahead to merge them.

grigi:
Right, so not entirely symmetrical. One can set the o2o relation from one side only, but read from both sides the same.

Is there anything you feel you need to do before I merge?

sinaso:
Yes let do the additions you wanted and one small fix to remove setter from
the reverse side altogether.

sinaso:
@grigi I think this should do... 
Thank you

grigi:
Awesome PR :+1: Merging now.

grigi:
Release 0.15.3 with this changes :-)

sinaso:
Thank you!

PR #244, opened 2019-11-25
WIP: EnumField and CharEnumField

wallneradam:
Field types to map python's Enum class to database field.

## Description

I added EnumField and CharEnumField into fields.py.

With this you can create models like this:
```python
from tortoise import models, fields
from enum import Enum


class Service(int, Enum):
    python_programming = 1
    database_design = 2
    system_administration = 3


class Currency(str, Enum):
    HUF = 'HUF'
    EUR = 'EUR'
    USD = 'USD'


class Price(models.Model):
    service: Service = fields.EnumField(Service)
    price = fields.DecimalField(14, 6)
    currency: Currency = fields.CharEnumField(Currency, default=Currency.HUF)
```

## Motivation and Context

I use these enumeration fields in my project and found it very useful. Are you interested in merging this? If yes I will try to write tests and documentation for it.

## Checklist:


tortoise/fields.py
```diff
@@ -407,6 +408,76 @@ def to_python_value(self, value: Any) -> Optional[UUID]:
...
+
+    def __init__(self, enum_type: Type[Enum], description: str = None, **kwargs) -> None:
+        try:
+            if description is None:
+                # Check and create description
+                description = "\n".join([f"{e.name}: {int(e.value)}" for e in enum_type])[:2048]
+            else:
+                # Just check if it is numeric enum
+                assert [int(e.value) for e in enum_type]
+        except ValueError:
```
grigi: This doesn't seem right?
Only check by raising an `AssertionError` when description is set, then catch `ValueError`?

Also, could you avoid using `assert` in final code?

And doesn't python have `IntEnum` which should be used if it is an all-int enum?
https://docs.python.org/3/library/enum.html#enum.IntEnum
Could we not just require that the enum used is an IntEnum?

wallneradam: It was just a try to bypass the checking system you have. Without it the problem was to not use the result of the comprehension.

If you see the source of `IntEnum` it is just one line:
```python
class IntEnum(int, Enum):
```
@grigi, my goal was here to check type early, because it could cause trouble later (in `to_db_value`) if you use Enum with values other than int. Because enums can have any values in Python. Though we could use IntEnum as type hint and IDEs can check it and my type check may not necessary at all. Also could be good to rename EnumField to IntEnumfield. What's your opinion?

grigi: Yes, `IntEnumField` and `CharEnumField` makes sense to me.

The one concern I have re the `CharEnumField` is the auto max_len.
I feel if you don't specify anything we should have a decent minimum size.
e.g. if you start with `Y` and `N`, then later add `Maybe`, it's going to explode.
If you specify a size, then you are managing it explicitly, else we can guess, but I'd then guess a minimum of 10 chars?

And the validation on creation time of the int range is valid.
Right now we have a range of signed 16-bit, so we can validate that the enum values are int, and from 0-32767.
If not, we could consider raising a `FieldError`?

wallneradam: Ok, I'm going to modify the PR accordint to your advices. Though I do not really get your concern about the auto size. If you want you can specify the size, then it would work the same way as real `CharField`. In your exampe you can specify 10 as the size. It is just comfort feature. For me it is easier this way, because my schema sync can immediately modify the table according to the maximum value. Though if you want, I could make 10 as the minimum and maybe only the value of `True` triggers the autosize feature, or something like that.
Is it ok this way?

grigi: Hmm, valid point.
Either we make `max_len` a positional field, and if you specify a `0` value we auto-size
or
We could make it auto-size by default, and put a big fat usage warning in the docs?

wallneradam: I prefer the 2nd option. Because if you use an enum, its items acts like constants of possible values. It won't change in production. I prefer solutions where everything is automated if possible. But it is your project, your decision. And both way works for me.

grigi: Big fat usage warning in docs it is then :-)

grigi:
Hi @wallneradam Is everything allright? Do you need any help?

wallneradam:
@grigi, yes just I had to do multiple projects at the same time, and have had no time for this.

Please review it. I created tests, but I don't know how to create tests for model creation level to check validation and auto length feature.

grigi:
Looks good to me. The p3.6 test fail is my fault. I seem to have forgotten to push a fix to the develop branch.
I'll merge this and fix develop afterwards.
Thank you for the awesome work. In the port to #206 I'll try add the auto-size test.

Issue #250, opened 2019-11-27
Pydantic support


wallneradam:
**Is your feature request related to a problem? Please describe.**
It is actually not a feature request, but a question about integrating the feature in subject.
My project uses FastAPI as backend, which uses Pydantic models everywhere. It is a wonderful project uses all good things in modern Python. I'm working on a solution, to convert Tortoise model to Pydantic model runtime. It has started working, though not complete yet, but will be in a few days.
It makes my API functions very clean, because I can use my models as input and output of my API calls, no need to specify the same thing in different way. It has automatic documentation without extra work, And it solves the serialization problems in #13 as well (https://pydantic-docs.helpmanual.io/usage/exporting_models/).

**Describe the solution you'd like**
I extended the Tortoise model like this:
```python
class Model(models.Model):
    # If not empty, only fields this property contains will be in the pydantic model
    __pydantic_include__: Tuple[str] = ()
    # Fields listed in this property will be excluded from pydantic model
    __pydantic_exclude__: Tuple[str] = ()
    # Computed fields can be listed here to use in pydantic model
    __pydantic_computed__: Tuple[str] = ()

    @classmethod
    @functools.lru_cache(maxsize=128)
    def pydantic_model(cls, *, exclude: Tuple[str] = None, include: Tuple[str] = None,
                       computed: Tuple[str] = None, name=None) -> pydantic.BaseModel:
        ...
```
The `pydantic_model` static method will create a Pydantic class which can be used as schema descriptor and serializer for FastAPI. I even created a solution, which can use comments before fields as description in the automatic documentation (OpenAPI) of FastAPI.

It can generate multiple Pydantic class from the same Tortoise class, to be able to use different views for the same database object. E.g. you can use less fields for one API call than the other. Or you can have private fields. The defaults can be described in the `__pydantic` prefixed properties. And if you need more views you can use the optional parameters of the method. It may be better to include this in the `class Meta` to make it more Tortoise friendly, it was just a quick implementation.

Then the created `PydanticModel` class has an async `from_orm` method which can be used to convert the database results to the pydantic counterpart. Then you can just give it to the API to send it serialized as JSON or anything you want. The from_orm can automatically fetch the needed not already fetched foreign key and many2many fields.

**Additional context**
Are you interested in this? 
[...]

PR #260, opened 2019-12-06
WIP: Pydantic support

wallneradam:
Ability to convert tortoise model to pydantic model, and use it as a serializer.

## Description

- It creates a view as pydantic class from tortoise model class at runtime.
- It creates a schema, which can be recursive when relations are used. 
- It can include/exclude fields from the schema
- Multiple schema could be created for the same tortoise model
- Models are cached
- It is protected from infinite recursion
- The schema can serialize tortoise objects into a pydantic object
- The pydantic object can be serialized into other types e.g. JSON
- Computed field support

## Motivation and Context

It is useful as a serializer, especially in FastAPI. 

More details can be found in issue #250. Since then I made it use the `Meta` class for configuration.

## How Has This Been Tested?

It has been tested only in my project.

## Checklist:


grigi:
Ooh :upside_down_face:

So this will be used like so?
```py
class Something(Model):
    name=CharField(max_len=255)
    ...

SomethingPydantFull = Something.pydantic_model()
SomethingPydantLite = Something.pydantic_model(include=('id','name'))
```

Then you use it like
```py
s_lites = await SomethingPydantLite.from_tortoise(await Something.all())
s_fulls = await SomethingPydantFull.from_tortoise(await Something.all())
```
?

And how would the inverse work? e.g. I create a `SomethingPydantFull` and then `.save_tortoise()` or `.create_tortoise()` etc… ?

wallneradam:
Yes, something like that. I will create examples and tests next week.
The reverse is not implemented yet. And actually I don't need it. But it could be useful for inputs.

My original goal was to be able to use something like this in FastAPI
```python
class User(Model):
    """
    User Model description
    """
    class Meta:
        pydantic_exclude = ('password', 'invoice_address.user')

    # User ID
    id = BigIntField(pk=True)

    # The email address of the user, used as login name as well
    email = CharField(max_length=255, unique=True)

    invoice_address: ForeignKeyRelation['Address'] = ForeignKeyField('models.Address', null=True)

    # The password of the user
    password = PasswordField()


PUser = User.pydantic_model()


@router.post('/login')
async def login(
        email: EmailStr = Query(..., example="test@foo.org"),
        password: str = Query(..., min_length=MIN_PASSWORD_LENGHT, example="Qwe123")
) -> PUser:
    """
    Login with email address and password
    :param email: Email address of the user
    :param password: The password of the user
    :return: User object
    """
    query = User.filter(email=email, password=password).first()
    user = await query
    return PUser.from_tortoise(user)
```
(I created a custom router for FastAPI so the real example is more clean, bacause it can detect if I return a tortoise model, and it can automatically convert everything.)

From the above code, I will have a schema for the API call similar to this:
```javascript
{
    schemas: {
        Address: {
            title: "Address",
            type: "object",
            properties: {
                id: {
                    title: "Id",
                    type: "integer"
                },
                country: {
                    title: "Country",
                    enum:
                        [
                            "36"
                        ],
                    type: "integer"
                },
                city: {
                    title: "City",
                    type: "string"
                },
                zip: {
                    title: "Zip",
                    type:
                        "string"
                },
                address: {
                    title: "Address",
                    type:
                        "string"
                },
                lat: {
                    title: "Lat",
                    type:
                        "number"
                },
                lon: {
                    title: "Lon",
                    type:
                        "number"
                },
                full_address: {
                    title: "Full Address",
                    type:
                        "string"
                }
            },
            description: "Custom Pydantic BaseModel "
        },
        User: {
            title: "User",
            type:
                "object",
            properties: {
                id: {
                    title: "Id",
                    type: "integer",
                    description: "User ID"
                },
                email: {
                    title: "Email",
                    type:
                        "string",
                    description:
                        "The email address of the user, used as login name as well"
                },
                invoice_address: {
                    title: "Invoice Address",
                    allOf:
                        [
                            {
                                $ref: "#/components/schemas/Address"
                            }
                        ]
                },
            },
            description: "User Model description"
        }
    }
}
```

And the result of the API endpoint will look like this:
```json
{
        "id": 27,
        "email": "test@foo.org",
        "invoice_address": {
            "id": 0,
            "created_at": "2019-12-06T15:43:56.550Z",
            "modified_at": "2019-12-06T15:43:56.550Z",
            "country": "36",
            "city": "string",
            "zip": "string",
            "address": "string",
            "lat": 0,
            "lon": 0,
            "full_address": "string"
        },
    }
```

grigi:
That's pretty cool :-)

I had a further look, and why are you not using the `Tortoise.describe_model(SomeModel)" instead of manually introspecting the fields?
The reason I ask is because `describe_model` should report all that info for you, in a much more reliable way (e.g. no need to introspect and compare fields, we do classification in one place).

And with #206 it will also report the DB-specific config (but I don't think you need it).

wallneradam:
The most important reason is I just don't know about this `describe_model` function :-)

Another reason was the evolution of the work. 1st it just tried to use the annotations, because at start models are not fully initialized, when FastAPI routers try to use them. But later I refactored my API to initialize (import) routers after all models are initialized. Now it is a requirement for this pydantic conversion to work correctly, so I can just use your `describe_model`. Although it could be solved by creating some placeholder classes, or somethong very tricky, it could be overcomplicated.

Also my goal was to create a pydantic model with the same field order as the tortoise model was specified by the user. And `__annotations__` have the order of the fields. Though in the final version of this PR it is not fully the case either.

I'm going to try to rewrite it to use `describe_model`.  Only one problem I have. I created a feature request about it:  #262

grigi:
> The most important reason is I just didn't know about this `describe_model` function :-)

:laughing: 

> Also my goal was to create a pydantic model with the same field order as the tortoise model was specified by the user. And `__annotations__` have the order of the fields. Though in the final version of this PR it is not fully the case either.

Ordering is at least kept from Py3.6+ So you'll mostly keep ordering when using `describe_model`.
It may not be an exact fit to what you want to have, but I'm happy to make to more useful, as long as we are careful if we break backwards compatibility.

I am quite excited about full pydantic support. Unfortunately we are not using it anywhere with work, so I never really got to play around with it enough.
One of the big features of pydantic is apparently SerDes speed? Once we have a nice clean API, I'll see if there is any easy performance wins.

wallneradam:
> Ordering is at least kept from Py3.6+

Yes but the problem is normal fields and relations are separated into different dictionaries, so the orders not preserved this way. But maybe it would be good to have a list of fields in describe model with every fields and backward fields in order of specification. And the auto generated fields last.

> I am quite excited about full pydantic support.

I'll try to implement the reverse as well. I don't think it is too hard. The complecated thing was to generate the Pydantic class, not to convert the objects.

grigi:
@wallneradam I'm happy if you continue without using `describe_model()` as you already have it working well. Just add a big `TODO` that state that we would like to refactor it.
I'm saying this because I would like to use `describe_model()` to also generate the DDL in the future, and this is essentially an internal API, so less affected by legacy.

long2ice:
hello,any news here?

wallneradam:
> I'm happy if you continue without using `describe_model()` as you already have it working well. Just add a big `TODO` that state that we would like to refactor it.

@grigi  I've rewritten it. It is much cleaner and also faster than it was. I had problems with the previous version, so it was better to rewrite. I changed the hacky inspect.stack() code as well to use a simple tuple insted.  And now it can sort fields and exclude the raw "_id" fields.

About the deserialization. I don't know how to create a universal solution. My problem is, I need to know somehow if the object needs to be created or it needs to be updated in the DB. Also these objects can be nested, which makes it much more complicated. You may want to update your current object with new subobjects...
Maybe it could work if the object has the "_id" fields, than it should be updated otherwise it should be created. But now the default is to excude raw fields in my code. 
Any idea or opinion?

tortoise/models.py
```diff
@@ -783,3 +813,28 @@ class Meta:
                     table="custom_table"
                     unique_together=(("field_a", "field_b"), )
         """
+
```
grigi: The `Meta` nested class here is only for documentation. At the moment all the processed fields end up in the `_meta` (which is a `MetaInfo` class)
:thinking: Possibly we move the documentation from the `.rst` file into here, and also specify default values, then we don't have to deal with `.get()` so much... :thinking:

tortoise/pydantic.py
```diff
@@ -0,0 +1,384 @@
...
+    from tortoise.models import Model
+
+
+# Pydantic is an optional dependency
+try:
+    import pydantic
+    from pydantic import BaseModel
+
+    @functools.lru_cache()
+    def _get_comments(cls: Type["Model"]) -> Dict[str, str]:
```
grigi: This is actually a good idea, we should make this a standard.
Is it OK to populate the description from the comment if not explicitly defined?

What should be the order of preference?
1. `description=`
2. docstring
3. preceding comments

Note that a comment that starts with `#: ` sets the docstring, I think.

(Don't change this unless you REALLY want to, I'm thinking of moving this kind of thing up at a later time)

tortoise/pydantic.py
```diff
@@ -0,0 +1,384 @@
...
+                comment = re.sub(r"(^\s*#\s*|\s*$)", "", match[0], flags=re.MULTILINE)
+                # Class name template
+                comment = comment.replace("{model}", cls.__name__)
+                # Change multiline texts to HTML
+                comments[field_name] = comment.replace("\n", "<br>")
+
+        return comments
+
+    @functools.lru_cache()
+    def _get_annotations(cls: Type["Model"], method: Optional[Callable] = None) -> Dict[str, Any]:
```
grigi: Ditto, this should be standard.

grigi:
> @grigi I've rewritten it. It is much cleaner and also faster than it was. I had problems with the previous version, so it was better to rewrite. I changed the hacky inspect.stack() code as well to use a simple tuple insted. And now it can sort fields and exclude the raw "_id" fields.

It certainly makes a bit more sense to me now :-)

> About the deserialization. I don't know how to create a universal solution. My problem is, I need to know somehow if the object needs to be created or it needs to be updated in the DB. Also these objects can be nested, which makes it much more complicated. You may want to update your current object with new subobjects...
> Maybe it could work if the object has the "_id" fields, than it should be updated otherwise it should be created. But now the default is to excude raw fields in my code.
> Any idea or opinion?

I think for deserialization to be allowed at all, it needs to have the PK of every record in the field. I think it is perfectly fine to require a strict set of guidelines. Such as:

* Record MUST have pk info for any update to happen
* Record MUST have all required fields to allow creation of record
* Sub-record requires that it either has PK or all required fields

If it doesn't meet the criteria, then we should throw a deserialization exception, I think.

Also, deserialization could be done as a later PR, but probably of higher importance is actually writing tests.
If you could provide a sample of how it should be used, I can help turning it into interface tests (which is important, that allows us to refactor how it works without breaking the world for some people)

grigi:
Oh, just for clarity. I'm not requesting any changes on the code you submitted itself at this stage.

All we really need is a basic test suite.

I'm happy to get this into `develop` so we can swarm on it :+1:

tortoise/pydantic.py
wallneradam: I like this preceding comment. It is more natural and readable for me, because when I write my models I always create a description for it as a comment.  
Though the `description` parameter in my opinion is for the database fields primarily. This is why I made preceding comment support for Pydantic classes. For my API I need a better and more verbose description suport without writing the same thing more times. But I may not want the same description appear in the table field description. 
But it is only one (my) use case. For someone it maybe useful to write table comments this way.
Also note that this has multiline comment support, which is needed for API description. So you can use HTML, reStructuredText, MarkDown, or anything like that in here. The database descriptions are limited in size so don't think that they are created for this kind of documentation.

As far as I know `#:` is only in Epydoc which is not developed since 2008. Others don't have variable level  docstrings.
It is not hard to rewrite to use only comments started with `#:`if you prefer that. Just it can be a collision with reStructuredText format (which PyCharm uses by default), when you start a comment like eg. `#:return: ...`.

tortoise/pydantic.py
wallneradam: It is good because it can handle the `ForwardRef` annotations like this: `ForeignKeyRelation['Address']`.

tortoise/models.py
wallneradam: I found this `Meta` class is useful to store the defaults and  their descriptions here, but if you have better or standard place for them, please help in it. 

These  settings are different from the _meta ones, because they are independent from the Tortoise model, and should not be in the `_meta` at all imo. It could be better to separate them into a different class? e.g `class PydanticMeta:`?

wallneradam:
> Also, deserialization could be done as a later PR, but probably of higher importance is actually writing tests.
If you could provide a sample of how it should be used, I can help turning it into interface tests (which is important, that allows us to refactor how it works without breaking the world for some people)

I see, I'm starting to write a doc and examples for it. But I may need help in the tests.

tortoise/models.py
grigi: Unsure, As long as there is adequate namespacing, it should be fine.
This was me thinking of moving the field docs into `Meta` (and then also provide defaults there instead of in code via e.g. `.get("indexes", [])`).

A PydanticMeta might make more sense? Then there is a clearer separation of concerns?

tortoise/pydantic.py
grigi: Sphinx autodoc (which we use) uses `#:`, but you are right, it is a tooling convention, the Python parser doesn't care.

I see, we really need two levels of description. a short one, and a long one.
The long one can default to the short one if provided.

I think I'll introduce a `long_description` var, that either is a param, or it tries to get the preceding comments like this.

Then the sometimes HUGE descriptions generated by the enum fields could go into long_description by default?

tortoise/pydantic.py
grigi: `ForwardRef` caused me lots of pain with type annotation in docgen. Sometimes the ref isn't in scope because it is hidden behind a `if TYPE_CHECKING:` condition :'(

grigi:
To get this nicely prepared for a release, I think we need to work on a common branch. I'm giving you repo write access.
Please move this into a pydantic branch on tortoise git, and then we can both work on the docs and tests :grin: 
Or do you prefer me to merge this in first?

tortoise/pydantic.py
wallneradam: I created that automatic enum description because when you see enums in the DB, you see only numbers. So for me it is a help to be able to know which number means what. So that should be included in the DB (short_description) IMO. 

Anyway I like the idea. And that way the description can be from `describe_model` too.

grigi: I looked up comment limits, and MySQL has a limit of 255 for a column (on v5.5). It is upped to 1023 since v5.6 (which is also now quite old).
We have to just be careful to limit the comment length for the generated comments.

grigi:
I'll merge this into a `pydantic` branch on Tortoise, do a new PR, fix the build and you are welcome to help on that branch (as you now have commit access).

I want to get this relatively polished up before we do a release with it. (that means tests & docs, possibly elevating the comments & annotation functionality)

PR #276, opened 2020-01-17
Pydantic

### LINKED ISSUE #250: Pydantic support

grigi:
This PR continues on #260 in a shared repo.
Fixes #250 

TODO:
- [x] Fix CI
- [x] Example
- [x] Behavioural Tests
- [x] Recursive model
- [x] List of Models
- [x] Revise backward_annotations if required? (think not anymore)
- [x] Type all pydantic params
- [x] Computed/callable fields
- [x] Explicit name
- [x] Verify sort fields by alpha
- [x] Verify sort fields by definition
- [x] Move to `tortoise.contrib.pydantic`
- [x] Move params into a `PydanticMeta` to keep it clean, remove prefixes
- Documentation
  - [x] Examples in doc
  - [x] Basic Usage/tutorial
  - [x] Reference
- To move out:
  - [x] ~Apply default model order on related queries~
  - [x] ~Fix preserving ordering on M2M~
  - [x] ~Fix post-M2M relations not being evaluated~
  - [x] ~Fix default type of JSONField~
  - [x] ~Elevate extracting comments (and possibly description) (based on this conversation: https://github.com/tortoise/tortoise-orm/pull/260#discussion_r364375665)~
  - [x] ~Early schema initialization~


grigi:
@wallneradam Could you please drop an example, even as a comment here, Then I can con continue working on this :-)

wallneradam:
@grigi Sorry for not responding until now. I had a big problem in one of my other projects and I don't have any time for this. I hope I can finish that in the next few days, and I can come back for this. Sorry again...

grigi:
@wallneradam Don't worry too much,, I have a some examples now and is getting an idea of how to use it.
I will definitely ask you to review it :grin: 

On a separate note, I changed a few things ,and plan to change some more:
* I now serialise the primary key.
* I added a `recursion_allowed` (which defaults to False) which will automatically limit any cycles, so that it makes adding un-annotated relations safer.
* I am going to move it all to `tortoise.contrib.pydantic` and have it only depend on Tortoise.
* I am going to mode the Meta stuff to a PydanticMeta, to keep the separation clean
* Would want to serialise from a queryset, so one could return a list easily. How would I get Pydantic to serialise is as a list?

I have found quite a few bugs so far:
* We don't fetch 3'rd level or higher data.
* If I have overlapping Pydantic models: https://github.com/tortoise/tortoise-orm/pull/276/files#diff-5877c39ebc7323ff914955ee78b01125R64-R66 then the first model works, and the rest fail with a verification error? I think there is some kind of leakage?
* Some models refuse to serialise, I haven't looked into why yet.

But I am quite impressed at how powerful this is :smile:

wallneradam:
I did some unfinished fixes in the meantime too. I'm going to review and help and finish the documentation at the weekend.

grigi:
@wallneradam So far I'm happy with this work, And quite excited about the prospects of this.
There is one outstanding bug, which I found is actually an issue in the `.select_related()` where it doesn't follow relations after a M2M join.
Which is quite understandable, but due to the need to preselect everything so that serialization can happen correctly, it now becomes a blocker.
(Normally you could just await the missing objects, and they would be fetched)

Also I'm quite excited about the extra information we have that would allow more efficient query generation.

## List of models?

And lastly, resolving on a queryset (which seems obvious to me) would require us to build a special pydantic list model, Should we just attach an anonymous list model to each pydantic model? The issue then is how would one name the type? e.g:

A single model:
```Event_Pydantic = pydantic_model_creator(Event)```

Just doing a:
```List[Event_Pydantic]```
model won't cut it, as now you lose the Pydantic helpers.
So we need to create a new model:
```py3
class Event_List_Pydantic(BaseModel):
    events: List[Event_Pydantic]
```

But is it OK if it is anonymous? or how would I name it?

I could either require the `pydantic_model_creator()` to return both (so you could name it), or have a standard "list" type?
e,g, `Event_Pydantic.List` type?

Or I might just be missing something obvjous here?

grigi:
Added the List of Models case. @wallneradam What do you think of this interface?

```py
Event_Pydantic_List = pydantic_queryset_creator(Event)
pl = await Event_Pydantic_List.from_queryset(Event.filter(address__event_id__isnull=True))
print("All Events without addresses:", pl.json(indent=4))
```

Another option would be to add a from_queryset directly on the regular tortoise model, and it generates the listed version? but the schema is different?

Possibly a `Event_Pydantic.plural.? No, that feels a big dumb too.

grigi:
Maybe make the plural model more incremental? e.g. one dis a `pydantic_model_creator()` and then pass that in to a `pydantic_queryset_creator`? That would allow them to use the same internal structure (so less duplication).
And then the plural name of the model would be a bit less weird?

Or does it matter? Because that is pretty much the same interface.

grigi:
Our order of initialisation is very likely going to cause complications for users.

Consider this scenario:
1) Define models in module
2) import models
3) pydantify them
4) use their generated schema as a parameter for a decorator that enforces schema validation. This does some codegen and then wraps the response.
5) init the project

Point 3 will be garbage, because until 5 happens (init and finalize the models) we don't have the full models yet, and so eveything is broken :-(

At least I can make 3 check if the models are finalized to not lull the user into a false sense of security. Also possibly make it lazy, but in that case would move the break to step 4, which may not be enough?
Maybe we need an early model finalization? As we don't need to bind the model to a DB yet, we just need all the fields defined. (That would be outside of the scope of this, I think we go live with this limitation)

grigi:
@wallneradam Please do a review.
I still have to finish docs, but I'm happy with the functionality and interface as it is.
I also removed all the non-pydantic specific work (bugfixes, etc...) to try and minimise this already massive PR.

grigi:
Thanks for the review!
>     1. could the .dict() method be mentioned to get a dictionary instead of a json string? it's so much more useful especially if this isn't the root payload you are returning but are inserting it into another dict and you can use alternative serialization that might be faster (unless this internally already uses ujson as well?)

Will add it to an example and the docs.

>        2)i can't find any info on awaiting vs not awaiting with the from_tortoise_orm and from_orm methods, while the first one pulls in all the related fields from relations, and the other doesn't, meaning that if you use the metadata to ignore it or handle them yourself you can stop it from pulling in a ton of unneeded data as this recurses down (maybe also a warning on this as this can seriously affect the speed of things)

Yes, I will make sure to mention it, but probably would have when getting to the docs.
> 
>     2. could there maybe be a hook for the starlette register_tortoise hook to give you a callback? or even initialize these serializers for you? right now i have to use bouncer functions to lazy init them to make sure this all happens AFTER this initialized tortoise, the alternative is not using this hookin function and handle the startup and shutdown yourself but it kinda makes said hook useless then

Would early partial init not do it for you? https://github.com/tortoise/tortoise-orm/blob/develop/CHANGELOG.rst#future (second bullet point)

>     3. some docs on what all the relational type hints are? i pretty much figured out the ones i need from pure guessing, diving into the code on what all exists and trial and error, the examples include a lot of new type hits but not a full list or usecases of those

Definitely. Is there anything big missing from the examples?

grigi:
I mean how in this example: https://github.com/tortoise/tortoise-orm/blob/a6075000602aed3f659bfe391a3b7904a256ef7c/examples/pydantic/early_init.py#L37-L39 the pydantic-model is defined as a global, so one could just use it directly?

Also, I should probably build a fastapi example :thinking: But it may be best to do that as another PR.

Issue #279, opened 2020-01-30
Support auto generate UUID for non pk field


lntuition:
First, Thank you for support nice software :)
I find that UUIDField only auto generate UUID4 only that field is primary key.
How about support auto generate UUID key for non pk field?

Maybe we can use like below
``` Python
uid = fields.UUIDField(auto_generate=True) # default : False
```

Also, I'm not good at orm and database well. I want to hear about other's opinions.
I really appreciate if anyone answer to me! Thank you :+1:

abondar:
Hi!

Actually, for pk uuid fields generation is done inside of Tortoise just like that:
```
    def __init__(self, **kwargs) -> None:
        if kwargs.get("pk", False):
            if "default" not in kwargs:
                kwargs["default"] = uuid4
        super().__init__(**kwargs)
```

So, if you wanna same behaviour for non-pk uuid field you just have to do it like that:
```
from uuid import uuid4

uid = fields.UUIDField(default=uuid4)
```

lntuition:
@abondar Thank you for your friendly comment :)

What do you think about adding new keywords to `UUIDField` for auto-generate UUID for non pk field? Adding new keyword can be useful to user and also prevent unnecessary uuid import.   

If you do not agree, then how about add some example code about this issue?
I think it is hard to find upper solution just looking document. 
We must view source code to solve upper problem.

Thank u and Have a nice day ! :smile:

abondar:
Well, I think there really is issue with `default` parameter for fields documentation. It can accept functions as value and this function will be called if field has to be populated by default value.

It isn't documented in field's reference in docs, which, I think, happened mostly because it's behaviour emulates Django's same parameter, so it hasn't bothered us before. 
I think we will have to improve docs in this terms.

Regarding `auto-generate` parameter I would oppose such solution, because it's seems to be too specific and would bring more confusion to users, than be really helpful, cause people will have to find out what exactly this params mean and it means we would have to document it thoroughly.

But if we gonna document something I think we should better document `default` param behaviour :)

grigi:
Yes, I have been having to do many support queries where the docs either had something missing, or was hard to find. We need doc guidelines.

lntuition:
> Well, I think there really is issue with `default` parameter for fields documentation. 

I agree with above opinion. So i will close this issue after documentation updated. Thank u :)

grigi:
[...]

Issue #272, opened 2020-01-10
Equivalent of "UPDATE...WHERE"?


mojimi:
Just a quick question, is there a way of doing the following :

```
async with in_transaction() as connection:
    await connection.execute_query(f"UPDATE EMPLOYEE SET SALARY = 5000 WHERE OCCUPATION = DEVELOPER")
```

But without using raw sql, something like :

`await Employee.filter(occupation='developer').batch_update(salary=5000)`

I understand I could also do a for loop inside a transaction but there's no way that's the right way to do it.

AEnterprise:
you where actually pretty close:
 ``await Employee.filter(occupation='developer').update(salary=5000)``
will do that

mojimi:
Thank you, totally missed that in the docs

grigi:
Tried to make docs clearer re `.update()` usage.

Issue #270, opened 2020-01-10
ForeignKeyField can't use isnull


lpig:
isnull is not working in ForeignKeyField

models
```python
class User(Model):
    SEX = {
        0: "未知",
        1: "男",
        2: "女",
    }
    nickname = fields.CharField(max_length=200, description='昵称', null=True)
    sex = fields.IntField(description='性别', default=0)
    openid = fields.CharField(max_length=100, description='openid')
    avatar = fields.CharField(max_length=2000, description='头像', null=True)
    create_time = fields.DatetimeField(auto_now_add=True)

    class Meta:
        table = 'member_user'

    def __str__(self):
        return '%s（%s）' % (self.nickname, self.openid)


class Employee(Model):
    name = fields.CharField(max_length=200, description='名称')
    user = fields.ForeignKeyField('models.User', related_name='employees', null=True)
    create_time = fields.DatetimeField(auto_now_add=True)

    class Meta:
        table = 'member_employee'

    def __str__(self):
        return '%s' % (self.name)
```

code
```python
employees = await Employee.filter(user__isnull=True)
```

error
```
Traceback (most recent call last):
  File "/Users/lp1ig/venv/youyou-lottery-AOMm0tqX/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/Users/lp1ig/venv/youyou-lottery-AOMm0tqX/lib/python3.7/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/Users/lp1ig/venv/youyou-lottery-AOMm0tqX/lib/python3.7/site-packages/uvicorn/middleware/debug.py", line 81, in __call__
    raise exc from None
  File "/Users/lp1ig/venv/youyou-lottery-AOMm0tqX/lib/python3.7/site-packages/uvicorn/middleware/debug.py", line 78, in __call__
    await self.app(scope, receive, inner_send)
  File "/Users/lp1ig/venv/youyou-lottery-AOMm0tqX/lib/python3.7/site-packages/fastapi/applications.py", line 140, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/lp1ig/venv/youyou-lottery-AOMm0tqX/lib/python3.7/site-packages/starlette/applications.py", line 134, in __call__
    await self.error_middleware(scope, receive, send)
[...]

Issue #266, opened 2019-12-20
Indices support in database.


JuniorJPDJ:
I'd like to see an possibility to at least create one `index` or more.
It will allow us eg. to have set of columns unique.
I'm trying to rewrite plain aiosqlite code to tortoise and gain `pg`/`mysql` support but I just can't because I decided to use indices.
Example query I'd like to be able to model:
```
CREATE UNIQUE INDEX history_idx ON history(feed_id, post_id);
```

grigi:
This is supported?
https://tortoise-orm.readthedocs.io/en/latest/models.html#tortoise.models.Model.Meta.unique_together

JuniorJPDJ:
Oh thank you very much then. I'll check it.

JuniorJPDJ:
Yes, thank you.
It works, my bad.
I'd like not to create new issue, but I've question if something like that is implemented:
```
			CREATE TABLE IF NOT EXISTS feeds (
				id INTEGER PRIMARY KEY AUTOINCREMENT,
				url TEXT NOT NULL UNIQUE
			);

			CREATE TABLE IF NOT EXISTS history (
				feed_id INTEGER NOT NULL,
				post_id TEXT NOT NULL,
				FOREIGN KEY (feed_id)
					REFERENCES feeds (id)
						ON UPDATE CASCADE
						ON DELETE CASCADE
			);
```

I'm asking about `REFERENCES` section. It's supposed to delete every related key if his foreign key is deleted.

grigi:
https://tortoise-orm.readthedocs.io/en/latest/fields.html#tortoise.fields.relational.ForeignKeyField
The `on_delete` param should set that.

JuniorJPDJ:
Thank you very much :D 
I've not found this in docs.
Marking as resolved.

PR #284, opened 2020-02-04
Better Docs & Typing

grigi:
Done:
* [x] Get typing data pretty complete
* [x] Have every module indexed in Sphinx
* [x] Ensure every type is cross-linked in Sphinx
* Fix up all autodoc to have basic documentation
  * [x] Base DB client
  * [x] Hide private for data fields
  * [x] Document base field
  * [x] Complete QuerySet
  * [x] Document Q
  * [x] Document Prefetch
  * [x] Document functions/aggregates
* Add missing info from [label:documentation](https://github.com/tortoise/tortoise-orm/issues?q=is%3Aissue+is%3Aopen+label%3Adocumentation) issues:
  * [x] SSL db connections #255 
  * [x] Document that SQLite does a best effor emulation of Decimal fields #256 
  * [x] Drop database & caveats #232
  * [x] The `_id` field in FK O2O fields #259
  * [x] Make `on_delete` use case clearer #266
  * [x] Document `__isnull` needs to use `_id` for FK #270
  * [x] Queryset update operations need to be more prominent #272
  * [x] Document `default=` param for fields #279
* [x] Update contribution guide
* [x] Update roadmap


grigi:
@abondar @AEnterprise @zoliszeredi I did a large update on documentation.

Would be awesome if you can do a quick look-over and see if I made a big mistake somewhere?

docs/schema.rst
```diff
@@ -0,0 +1,20 @@
...
+===============
+Schema Creation
+===============
+
+Here we create connection to SQLite database client and then we discover & initialize models.
+
+.. automethod:: tortoise.Tortoise.generate_schemas
+    :noindex:
+
+``generate_schema`` generates schema on empty database, you shouldn't run it on every app init, run it just once, maybe out of your main code.
```
AEnterprise: is there any major downside to running it regardless? i currently run it all the time to not have to write an install script for new instances but just have tortoise figure it out during startup

tortoise/backends/base/client.py
```diff
@@ -54,59 +58,146 @@ def __init__(
...
+        :annotation: Type[BaseSchemaGenerator]
+
+        The DDL schema generator
+
+    .. attribute:: capabilities
+        :annotation: Capabilities
+
+        Contains the connection capabilities
+    """
+
```
AEnterprise: might be worth mentioning that none of this should be called directly but that the model methods should be used generally for people who end up here when searching for something like "insert" or "execute query"

docs/databases.rst
```diff
@@ -86,6 +94,8 @@ Parameters
     Duration of inactive connection before assuming that it has gone stale, and force a re-connect.
 ``schema``:
     A specific schema to use by default.
+``ssl``:
+    Either ``True`` or a custom SSL context for self-signed certificates. See :ref:`db_ssl` for more info.
```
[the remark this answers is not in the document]
grigi: Good point. Need to specify required/optional more clearly

docs/models.rst
```diff
@@ -228,10 +228,62 @@ The ``Meta`` class
...
+^^^^^^^^^^^^^^^^^^^^
+
+.. note::
+
+    A ``ForeignKeyField`` is a virtual field, meaning it has no direct DB backing.
+    Instead it has a field (by default called :samp:`{FKNAME}_id` (that is, just an ``_id`` is appended)
+    that is the actual DB-backing field.
+
+    This is an important detail as it would allow one to assign/read the actual value directly,
+    which could be considered an optimization if the entire foreign object isn't needed.
```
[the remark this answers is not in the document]
grigi: Yes.

Right now we don't do compunds PKs. So the answer is easy, but not satisfactory.

docs/schema.rst
grigi: Er, you don'g get a big error if your db lost its contents?

tortoise/backends/base/client.py
grigi: That would unfortunately require making querying SQL directly an official API. So this might be done better as an extra feature?

docs/models.rst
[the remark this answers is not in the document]
AEnterprise: even if it's an easy answer, probably best to mention it for people who are looking for it

docs/schema.rst
AEnterprise: this only is for new instances, like if someone starts it up for the first time then it'll just create the database on startup instead of crashing cause it doesn't exist and requiring them to run a seperate script

just wondering since you mention here you shouldn't do this if there is a downside to this design pattern

tortoise/backends/base/client.py
AEnterprise: i mean for general things, most people will wana do simple selects or inserts but if they just use the search to find "insert", this might come up

grigi:
Thanks for the suggestions @AEnterprise :+1:

Issue #281, opened 2020-01-31
Support foreign key to unique field


lntuition:
Well, I tried to set fk with other unique key. But i cannot find method with tortoise-orm.
Does tortoise support like below option? I cannot find with document.

With Django orm, it can be represented by `to_field` param.
> ForeignKey.to_field
>
> The field on the related object that the relation is to. By default, Django uses the primary key of the related object. If you reference a different field, that field must have unique=True.

grigi:
Oh, you mean to non-primary key, but still unique, key?

No, we just have FK's to the primary key of a table at this stage.

Technically it should work, not sure how much work it will be to actually make it work.

lntuition:
> Oh, you mean to non-primary key, but still unique, key?

@grigi Yeah, that's what i mean. I hope tortoise will support this feature soon. Thank you for comment.

lntuition:
https://github.com/tortoise/tortoise-orm/blob/c28784ce25f4adfe1b634092940dc82cb9b7b55e/tortoise/backends/base/schema_generator.py#L174-L193

@grigi I think foreign key schema reference field name is hardcoded with
` field=field_object.reference.model_class._meta.db_pk_field `
I don't know much about internal of tortoise-orm. Just change field name with passed param to ForeignKeyField can break other side?

grigi:
Yes, you are right. I suspect there are many small hardcoded code for this. There was a naïve assumption after all :-)

I think the ForeignKeyInstance should contain the remote db field as a parameter, and we set it in the constructor, and only validate it on init() (as at the time of defining the FK we may not have the remote model class yet). 

Then to verify it works everywhere, add it to the generated schema so we can test it's DDL generation is correct, add it to some of the test models, and copy the fk test cases, and also test for fails.

So this will be quite a lot of work. Do you want to start on it?

lntuition:
thank you for comment :) Also i want to contribute with this issue, but i'm not sure that i can do this... Because i'm beginner with this domain...

If there are someone who can lead me from begining, i will try to anaylaze internal of tortoise-orm..

grigi:
Sure, if you have time I can help you :+1: 

You already started, the `db_pk_field` is a `str` that contains the DB column name that we are joined to.

Lets start by only adding it to FK, as O2O is implemented almost identically, and will be easy to get working later on.

[...]

PR #287, opened 2020-02-10
Support FK's to unique key.

lntuition:
## Description
Support FK's to unique key.
* [x] Add `to_field` and `to_field_instance` to `Relational_field(...)`
* [x] Add `from_field` to `ReverseRelation`
* [x] Add validation for condition of foreignkey and pass `to_field_instance` at `_init_relations`
* [x] Change SQL schema generator with `to_field_instance`
* [x] Check SQL schema generation work with multiple backend (e.g. sqlite, postgresql, mysql ... ) 
* [x] Change foreignkey and reverse foreign key get/setter with `to_field_instance`
* [x] Change prefetch relations with `to_field_instance`
* [x] Change update foreignkey with `to_field_instance`
* [x] Change join with related field with `to_field_instance`
* [x] Add testcase 
* [x] Add examples
* [x] Renew Document

Fixed : #281

### DIFF
--- a/examples/relations_with_unique.py
+++ b/examples/relations_with_unique.py
@@ -1,0 +1,43 @@
+"""
+This example shows how relations between models work.
+
+Key points in this example are use of ForeignKeyField and ManyToManyField
+to declare relations and use of .prefetch_related() and .fetch_related()
+to get this related objects
+"""
+from tortoise import Tortoise, fields, run_async
+from tortoise.models import Model
+
+
+class Tournament(Model):
+    id = fields.IntField(uk=True)
+    name = fields.CharField(max_length=100, pk=True)
+
+    events: fields.ReverseRelation["Event"]
+
+    def __str__(self):
+        return self.name
+
+
+class Event(Model):
+    id = fields.IntField(pk=True)
+    name = fields.TextField()
+    tournament: fields.ForeignKeyRelation[Tournament] = fields.ForeignKeyField(
+        "models.Tournament", related_name="events", to_field="name",
+    )
+
+    def __str__(self):
+        return self.name
+
+
+async def run():
+    await Tortoise.init(db_url="sqlite://:memory:", modules={"models": ["__main__"]})
+    await Tortoise.generate_schemas()
+
+    tournament = Tournament(id=1, name="New Tournament")
+    await tournament.save()
+    await Event(name="Without participants", tournament_id=tournament.name).save()
+
+
+if __name__ == "__main__":
+    run_async(run())
--- a/examples/relations_with_unique.py
+++ b/examples/relations_with_unique.py
@@ -8,6 +8,7 @@ to get this related objects
 import logging
 
 from tortoise import Tortoise, fields, run_async
+from tortoise.functions import Count
 from tortoise.models import Model
 
 logging.basicConfig(level=logging.DEBUG)
@@ -25,7 +26,7 @@ class Student(Model):
     id = fields.IntField(pk=True)
     name = fields.TextField()
     school: fields.ForeignKeyRelation[School] = fields.ForeignKeyField(
-        "models.School", related_name="students", to_field="id",
+        "models.School", related_name="students", to_field="id"
     )
 
 
@@ -34,10 +35,13 @@ async def run():
     await Tortoise.generate_schemas()
 
     school = await School.create(id=1024, name="School")
-    student = await Student.create(name="Sang-Heon Jeon", school=school)
+    student1 = await Student.create(name="Sang-Heon Jeon1", school=school)
+    student2 = await Student.create(name="Sang-Heon Jeon2", school_id=school.id)
 
-    student = await Student.get(id=student.id)
-    print(await student.school)
+    await School.filter(students__id=student1.id).values("id", "name", student="students__name")
+    await Student.filter(id=student2.id).values("id", "name", school="school__name")
+    queryset = School.all().annotate(count=Count("students"))
+    await queryset.filter(name="School").first()
 
 
 if __name__ == "__main__":
--- a/examples/relations_with_unique.py
+++ b/examples/relations_with_unique.py
@@ -33,7 +33,7 @@ async def run():
 
     company = Company(id=1, name="First Company")
     await company.save()
-    # await Employee(name="Sang-Heon Jeon", company_id=company.uuid).save()
+    await Employee(name="Sang-Heon Jeon", company_id=company.uuid).save()
 
 
 if __name__ == "__main__":
--- a/examples/relations_with_unique.py
+++ b/examples/relations_with_unique.py
@@ -1,17 +1,12 @@
 """
-This example shows how relations between models work.
+This example shows how relations between models especially unique field work.
 
-Key points in this example are use of ForeignKeyField and ManyToManyField
-to declare relations and use of .prefetch_related() and .fetch_related()
-to get this related objects
+Key points in this example are use of ForeignKeyField and OneToOneField has to_field.
+For other basic parts, it is the same as relation exmaple.
 """
-import logging
-
 from tortoise import Tortoise, fields, run_async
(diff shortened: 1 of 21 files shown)


tortoise/backends/base/schema_generator.py
```diff
@@ -179,23 +179,26 @@ def _get_table_sql(self, model: "Type[Model]", safe: bool = True) -> dict:
...
+
+                field_type = reference.to_field.get_for_dialect(self.DIALECT, "SQL_TYPE")
+                field_name = reference.to_field.source_field
+                if not field_name:
+                    field_name = reference.to_field.model_field_name
+
                 field_creation_string = self._create_string(
                     db_field=db_field,
-                    field_type=field_object.get_for_dialect(self.DIALECT, "SQL_TYPE"),
+                    field_type=field_type,
```
grigi: Oh dear, our field type depends on the DB backend...
We would want the field object so that we can get the SQL_TYPE lazily (as a model shouldn't be tightly bound to a DB type)
I didn't think of this?
Maybe we should also store the `to_field_instance`? :thinking:

tortoise/backends/base/schema_generator.py
```diff
@@ -179,23 +179,26 @@ def _get_table_sql(self, model: "Type[Model]", safe: bool = True) -> dict:
                     if reference.description
                     else ""
                 )
+
+                field_type = reference.to_field.get_for_dialect(self.DIALECT, "SQL_TYPE")
```
grigi: Good catch, I didn't think about this :+1:
The only thing I'm concerned about is that `to_field` mutates? It starts of as an `Optional[str]` and then mutates to `Field`.

lntuition: Thank you for review. I think so and also CI catch that point(logs are below). So how can we fix this? I think pass `Field` to schema_generator is inevitable. Can we type case to_field or do we have another workaround?

``` bat
tortoise/backends/base/schema_generator.py:183: error: Item "str" of
"Optional[str]" has no attribute "get_for_dialect"
                    field_type = reference.to_field.get_for_dialect(self.D...
                                 ^
tortoise/backends/base/schema_generator.py:183: error: Item "None" of
"Optional[str]" has no attribute "get_for_dialect"
                    field_type = reference.to_field.get_for_dialect(self.D...
...
```

grigi: I think lets not mutate the type.

Leave the ``Optional[str]`` ``to_field`` in the Field, but only use it as read only.
Add a new field that is for internal use, like https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/fields/relational.py#L280-L285
possibly called ``to_field_instance``?

Consider moving the ``to_field`` to the ``RelationalField`` base class (which is a new common base class to fix some typing in a recent PR) to reduce duplication of code.

lntuition: Oh, I see. That's really good opinion. I will make patch for that. I also check your above checkpoint with PostgreSQL UUIDField and i work well.

grigi: :+1:

grigi:
Cool, could we turn your example into an automated test, to avoid accidental regressions. (and to vet that the generated SQL is correct)
Please add a model with fk-to-unique in here: https://github.com/tortoise/tortoise-orm/blob/develop/tests/schema/models_schema_create.py
And then unfortunately this behemoth: https://github.com/tortoise/tortoise-orm/blob/develop/tests/schema/test_generate_schema.py will need to be updated to vet the exact generated SQL.

You can run `make test` or `green` locally to run tests and ensure that they pass.
(I do expect these to pass)

Then we have the schema generation done. Next up would be resolving the fk correctly.
Basically all these needs to be de-hardcoded:
https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/models.py#L74
https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/base/executor.py#L214-L260
https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/base/executor.py#L356
https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/query_utils.py#L46-L88
(I think that's it)

Just complete your example to query the relation forwards & backwards and do a basic `await 
Just make sure that you follow the relation forwards & backwards and do a basic `await event.tournament` to confirm you get the right object.
Then we can just turn that into tests, and I think this is perfect then :-)

lntuition:
@grigi thank you for your comment. I generate testcase for bad relation and generate schema. I'd appreciate that you check above TC. And i have a question for about generate schema testcase. Does return of `get_schema_sql` has no order? (I don't know why `company` is generated first and then `defaultpk` next, etc )

And for write TC for fk, I want to refer existing TC. Can you tell me where i can find existing `JOIN` testcode?

grigi:
The get_schema_sql is very deterministic. 
Those tests generate the exact same SQL string on every run, as the test validation is an exact string equality
It shuffles the order based on relationships, so the tables dependencies get processed first, and M2M will always be last. We need to ensure that the tables that are referenced by FK's already exist. Hence the dependency tracking shuffling things.

grigi:
:thinking: I see what you mean. Even though it is definitely deterministic, it does feel like it does too much shuffling. If you want to, you can drop a `# TODO: ...` in the schema generator.  That way it might get brought to attention on the next refactor?

Your tests look really good! I :heart_eyes: the negative testing! At least now we know that the user gets a sane error message, which means less support! Yay!

tortoise/backends/base/executor.py
```diff
@@ -344,17 +344,22 @@ def get_update_sql(self, update_fields: Optional[List[str]]) -> str:
     async def _prefetch_direct_relation(
         self, instance_list: "List[Model]", field: str, related_query: "QuerySet"
     ) -> "List[Model]":
-        related_objects_for_fetch = set()
+        related_objects_for_fetch = {}
```
grigi: In  the pydantic branch I have a few fixes for this, basically to preserve ordering. I'll do a release this weekend with a few fixes here. So be prepared for conflicts here.

examples/relations_with_unique.py
```diff
@@ -0,0 +1,57 @@
...
+from tortoise import Tortoise, fields, run_async
+from tortoise.models import Model
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+class School(Model):
+    uuid = fields.UUIDField(pk=True)
+    name = fields.TextField()
+    id = fields.IntField(unique=True)
```
grigi: I like how you confused the issue here on purpose. Makes this a more robust test :-)

grigi:
Also, for some reason I can't even get Travis to try to run the last version in this PR. Maybe all the force pushes confused it?
Note that we will likely squash this PR into one nice commit, so we don't care about making the commit history look good at this stage. (and Force Pushes can lead to the Dark Side)

I see nothing that looks obviously wrong here :-)

examples/relations_with_unique.py
lntuition: Sure. I will add enough testsuite :)

lntuition:
@grigi Okay! after Resolving conflict, then i will stack commit not using force push.

tortoise/backends/base/executor.py
lntuition: Okay I see. I will resolve conflict after pydantic branch merged.

grigi: No, I merged the fixes into master (to keep the already huge pydantic work focused).

lntuition: Ah, I understand. I checked it

lntuition:
@grigi At this point, I think i write all testsuite for this feature (it includes unit and integrity test suite both) Can you review once more? If it is okay, then i will write example and change log and documentation. I hope this works well :smile:

grigi:
Looks good! Thank you, I'm going to merge it now :-)

PR #289, opened 2020-02-10
add `between and` support

long2ice:
## Description

add `between and` syntax support
## Motivation and Context

add `between and` support to replace use `gte,lte`both
## Checklist:

### DIFF
--- a/tortoise/filters.py
+++ b/tortoise/filters.py
@@ -1,6 +1,6 @@
 import operator
 from functools import partial
-from typing import TYPE_CHECKING, Any, Dict, Iterable, Optional
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Optional, Tuple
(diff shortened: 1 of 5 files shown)


tortoise/filters.py
```diff
@@ -53,6 +53,10 @@ def not_in(field: Term, value: Any) -> Criterion:
     return field.notin(value) | field.isnull()
 
 
+def between_and(field: Term, value: Any) -> Criterion:
```
grigi: The definition for this is PyPika is `def between(self, lower, upper):` So, could you change the type of ``value`` to ``Tuple[Any, Any]`` and then expand it to two parameters. This is just to make the code more self explanatory.

grigi:
:tada: 
Thank you for your first PR :-)

PR #294, opened 2020-02-19
add F expression support in queryset.update()

### LINKED ISSUE #245: F Expression for atomic update?
I'm looking for a way to do an atomic update of a field or multiple fields (to increment it)

In Django ORM this is possible using the `F` class, I didn't see anything equivalent in tortoise documentation.

The goal would be to achieve something like this

```
counter = Counter.objects.get_or_create(name = name)
counter.count = F('count') +1
counter.save( update_fields=["count"] )
```

To generate some SQL like

`update counter set count = count + 1`

(In my case I also want to cast in some cases)

`update counter set count = cast(cast(count as integer) + 1) as char)`

Wondering if either of the above are possible yet with tortoise.

long2ice:
## Description

add ``F`` expression support in ``queryset.update()``.
## Motivation and Context

to resolve #245 .
## How Has This Been Tested?

``make test``
## Checklist:

### DIFF
--- a/docs/query.rst
+++ b/docs/query.rst
@@ -282,3 +282,18 @@ You can view full example here:  :ref:`example_prefetching`
 
 .. autoclass:: tortoise.query_utils.Prefetch
     :members:
+
+
+F expression
+================
+
+An F() object represents the value of a model field. It makes it possible to refer to model field values and perform database operations using them without actually having to pull them out of the database into Python memory.
+
+For example to use ``F`` to update user balance atomic:
+
+.. code-block:: python3
+
+    from tortoise.models import F
+    await User.filter(id=1).update(F('balance') - 10)
+
+But ``user.balance = F('balance') - 10`` is not supported now.
--- a/docs/query.rst
+++ b/docs/query.rst
@@ -297,4 +297,4 @@ For example to use ``F`` to update user balance atomic:
     await User.filter(id=1).update(balance = F('balance') - 10)
     await User.filter(id=1).update(balance = F('balance') + F('award'), award = 0)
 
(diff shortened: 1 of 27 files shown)


tortoise/queryset.py
```diff
@@ -658,7 +660,12 @@ def _make_query(self) -> None:
                     db_field = self.model._meta.fields_db_projection[key]
                 except KeyError:
                     raise FieldError(f"Field {key} is virtual and can not be updated")
-                value = executor.column_map[key](value, None)
+                if not isinstance(value, (F, ArithmeticExpression)):
+                    value = executor.column_map[key](value, None)
```
grigi: Ok, so after investigating, I think due to F and ArithmeticExpression baving different interfaces. Both inherit `pypika.Term`, so maybe just check against that?

docs/query.rst
```diff
@@ -282,3 +282,19 @@ You can view full example here:  :ref:`example_prefetching`
...
+
+For example to use ``F`` to update user balance atomic:
+
+.. code-block:: python3
+
+    from tortoise.expressions import F
+    await User.filter(id=1).update(balance = F('balance') - 10)
+    await User.filter(id=1).update(balance = F('balance') + F('award'), award = 0)
+
+But ``user.balance = F('balance') - 10`` is not supported now.
```
grigi: I think this last line may confuse users?
May be better to state we ONLY support `F` expressions as part of a `.update()` queryset parameter for now?

long2ice: ok,I will update docs and fix the error test later.

grigi:
Hi @long2ice Thanks for this work!

PR #300, opened 2020-02-27
Windows install without C compiler

grigi:


grigi:
@AEnterprise Could you see if this branch can install on your locked down windows system?
a simple `pip install https://github.com/tortoise/tortoise-orm/archive/win_no_c.zip` should do it.

AEnterprise:
That one installs perfectly fine, thanks!

Issue #257, opened 2019-12-03
Ability to SUM multiple columns


bbedward:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to use the `Sum` function over multiple columns.

Example for a model

```
class Stats(Model):
    one_column = fields.IntField(default=0)
    two_column = fields.IntField(default=0)
```

I'd like to be able to do run a query like `select sum(one_column+two_column) from stats`

**Describe the solution you'd like**
`Stats.annotate(total_sum=Sum('one_column+two_column'))` (or something similar)

**Additional context**

Pypika can already resolve this as seen in this test:

https://github.com/kayak/pypika/blob/16339f85ed871d937609e9dbe26487b382ca5211/pypika/tests/test_formats.py

`fn.Sum(self.table_abc.fizz + self.table_abc.buzz)`

grigi:
Hmm, This would be a useful feature.
There is a syntax issue that could get in the way (string + string is an example). So we should probably wrap the strings in "field name" objects (and translate strings into one or multiple "field name" objects, to keep simple (and legacy) syntax working.)

Note that Functions/aggregations/Q have to resolve the actual columns lazily, because we map the fields to columns. 

:thinking: this might be rolled together with some of #245

lntuition:
With #245, apply F field to annotation can be workaround. maybe like this?
```python
Stats.annotate(total_sum=F('one_column')+F('two_column'))
```

If you agree, then i want to try this. But I think it can be hard work.@grigi Do you agree with this workaround? And also can you help me once more?

lntuition:
With my current view, I think add variable something like `_annotation_f_expression` into `QuerySet` and modify queryset and query utils looks good. I hope my solution looks ok :smiley:

grigi:
:thinking: It does make sense to merge `F` operators and Aggregations/Functions. There is some code overlap between them already.

What about merging the functionality of F and Functions? 
Then one could write
```py
Stats.annotate(average_sum=(Sum('one_column') + Sum('two_column')) / 2)
```
and it would do what one expects? And your example would work too?

lntuition:
I agree that adding `F` operator functionality to `Functions` is best solution. I need to check with that direction. Wow your example is better than mine 👍

lntuition:
[...]

PR #301, opened 2020-02-28
[WIP] Allow F expression to function argument

lntuition:
## Description
This patch is related #257. Allowing `F` expression to `Function` argument. 

* [x] Allow  `F` expression to `Function` argument
* [x] Fix type casting (with broadcasting also)
* [x] Add testcase (including negative test)
* [x] Add change to CHANGELOG.rst

Current :
Waiting review...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -74,6 +74,7 @@ future
 
 - Fix default type of ``JSONField``
 - Install on Windows does not require a C compiler any more.
+- Allow F expression to ``Function``. (#301)
 
 0.15.17
 -------
--- a/tests/fields/test_decimal.py
+++ b/tests/fields/test_decimal.py
@@ -146,8 +146,7 @@ class TestDecimalFields(test.TestCase):
 
     async def test_aggregate_sum_no_exist_field_with_f_expression(self):
         with self.assertRaisesRegex(
-            FieldError,
-            "Field not_exist is not default field in model, so can not be used with F expression",
+            FieldError, "There is no non-virtual field not_exist on Model DecimalFields",
         ):
             await testmodels.DecimalFields.all().annotate(sum_decimal=Sum(F("not_exist"))).values(
                 "sum_decimal"
--- a/tests/fields/test_decimal.py
+++ b/tests/fields/test_decimal.py
@@ -3,7 +3,7 @@ from decimal import Decimal
 from tests import testmodels
 from tortoise import fields
 from tortoise.contrib import test
-from tortoise.exceptions import ConfigurationError, IntegrityError
+from tortoise.exceptions import ConfigurationError, FieldError, IntegrityError
 from tortoise.expressions import F
 from tortoise.functions import Avg, Max, Sum
 
@@ -108,6 +108,40 @@ class TestDecimalFields(test.TestCase):
             values[0], {"sum_decimal": Decimal("37.26")},
         )
 
+        values = (
+            await testmodels.DecimalFields.all()
+            .annotate(sum_decimal=Sum(F("decimal") + 1))
+            .values("sum_decimal")
+        )
+        self.assertEqual(
+            values[0], {"sum_decimal": Decimal("40.26")},
+        )
+
+        values = (
+            await testmodels.DecimalFields.all()
+            .annotate(sum_decimal=Sum(F("decimal") + F("decimal")))
+            .values("sum_decimal")
+        )
+        self.assertEqual(
+            values[0], {"sum_decimal": Decimal("74.52")},
+        )
+
+    async def test_aggregate_sum_fail_with_f_expression(self):
+        with self.assertRaisesRegex(
+            FieldError,
+            "Field not_exist is not default field in model, so can not be used with F expression",
+        ):
+            await testmodels.DecimalFields.all().annotate(sum_decimal=Sum(F("not_exist"))).values(
+                "sum_decimal"
+            )
+
+        with self.assertRaisesRegex(
+            FieldError, "Cannot use arithmetic expression between different field type"
+        ):
+            await testmodels.DecimalFields.all().annotate(
+                sum_decimal=Sum(F("decimal") + F("id"))
+            ).values("sum_decimal")
+
     async def test_aggregate_avg(self):
         await testmodels.DecimalFields.create(decimal=Decimal("0"), decimal_nodec=1)
         await testmodels.DecimalFields.create(decimal=Decimal("9.99"), decimal_nodec=1)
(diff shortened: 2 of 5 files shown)


tortoise/queryset.py
```diff
@@ -667,9 +667,7 @@ def _make_query(self) -> None:
                 if not isinstance(value, (F, ArithmeticExpression)):
```
grigi: :thinking:  Should we change this to check for instance type `pypika.Term`? That is the common base class of both those?

tortoise/functions.py
```diff
@@ -34,14 +35,15 @@ class Function:
...
+    __slots__ = ("field", "field_is_str", "field_object", "default_values")
 
     database_func = BaseFunction
     # Enable populate_field_object where we want to try and preserve the field type.
     populate_field_object = False
 
-    def __init__(self, field: str, *default_values: Any) -> None:
+    def __init__(self, field: Union[str, F, ArithmeticExpression], *default_values: Any) -> None:
         self.field = field
+        self.field_is_str: bool = isinstance(field, str)
```
grigi: Is this needed? We are not mutating `self.field` that I can see? Are you planning to in the future?

tortoise/expressions.py
```diff
@@ -1,32 +1,67 @@
...
+                    field_object = left_field_object
+
+            (
+                arithmetic_expression_or_field.right,
+                right_field_object,
+            ) = cls.resolver_arithmetic_expression(model, right)
+            if right_field_object:
+                if field_object and type(field_object) != type(right_field_object):
+                    raise FieldError(
+                        "Cannot use arithmetic expression between different field type"
```
grigi: :+1:

tests/fields/test_decimal.py
```diff
@@ -95,6 +95,53 @@ def test_decimal_places_bad(self):
...
+        with self.assertRaisesRegex(
+            FieldError,
+            "Field not_exist is not default field in model, so can not be used with F expression",
+        ):
+            await testmodels.DecimalFields.all().annotate(sum_decimal=Sum(F("not_exist"))).values(
+                "sum_decimal"
+            )
+
+        with self.assertRaisesRegex(
+            FieldError, "Cannot use arithmetic expression between different field type"
```
grigi: Could you split these tests? There is no reason to have them both in the same test?

tortoise/expressions.py
```diff
@@ -1,32 +1,67 @@
...
+                    if func:
+                        arithmetic_expression_or_field = func(
+                            field_object, arithmetic_expression_or_field
+                        )
             except KeyError:
-                raise FieldError(f"Field {name} is virtual and can not be updated")
-            arithmetic_expression_or_field.name = name
+                raise FieldError(
+                    f"Field {name} is not default field in model,"
+                    f" so can not be used with F expression"
```
grigi: The message here is confusing to me? Do you say this only works with fields that have a default set?

tests/fields/test_decimal.py
```diff
@@ -108,6 +155,37 @@ def test_decimal_places_bad(self):
...
+            .annotate(avg_decimal=Avg(F("decimal") + 1))
+            .values("avg_decimal")
+        )
+        self.assertEqual(
+            values[0], {"avg_decimal": Decimal("13.42")},
+        )
+
+        values = (
+            await testmodels.DecimalFields.all()
+            .annotate(avg_decimal=Avg(F("decimal") + F("decimal")))
```
grigi: Could you do maths with one of the other fields? `decimal_nodec` or `decimal_null` instead. So we see that it works with both.

grigi: Also, what is our expected behaviour when doing arithmetic on nulls? We probably want to ensure consistent behaviour at least?

grigi:
@lntuition Looks good! I only have minor comments/questions.

tests/fields/test_decimal.py
lntuition: Sure. I will split those.

tortoise/queryset.py
lntuition: Looks okay. I changed it !

tortoise/functions.py
lntuition: Yeah. We are not mutating `self.field` current stage. hmm.. also can removed.. I changed it !

tortoise/expressions.py
lntuition: Yeah. Maybe this message can confuse user. Can you recommend more clearer message?

tests/fields/test_decimal.py
lntuition: We can do with one of the other fields. i will add testcase to test that scenario.

lntuition: And doing arithmetic on nulls should generate null. But i'm not sure that all database(sqlite, postgres. etc) do equals.

grigi: About nulls, I feel the most correct behaviour would to have the query fail? And then you'd have to explicitly set a defualt-if-null with `coalese()`?

At least it would be good to run the test with all db's to see what they do? And hope they do the same thing... At minumim just to document it?

tortoise/expressions.py
grigi: We'll only get a `KeyError` if the field is not a DB column. So maybe we should explicitly mention that?
`There is no DB-backed field called {name}`?
or
`There is no non-virtual field {name} on Model {model.__name__}`?

tests/fields/test_decimal.py
lntuition: Could you see this [link](https://www.sqlite.org/nulls.html)? This links saids add anything to null is correct behavior. It should be okay just addding testcase with add something and null. 

Also i add test with add something with null. Does CI test it with all supported database? (e.g sqlite, postgres, etc)

grigi: CI tests with all supported DB's yes. That's a handy link. It looks like the supported DB's should behave the same.

tortoise/expressions.py
lntuition: I think below looks better ! I will change the message :+1:

CHANGELOG.rst
```diff
@@ -74,6 +74,7 @@ future
 
 - Fix default type of ``JSONField``
 - Install on Windows does not require a C compiler any more.
+- Allow F expression to ``Function``. (#301)
```
grigi: Could you please expand this a bit. Whilst technically correct, the average user will not see the implication.
Rather say ``Allow usage of F expressions in annotations`` ?

grigi:
Looks good. Please just improve the changelog, and I'll merge. :+1:

CHANGELOG.rst
lntuition: Sure, I change changelog and also resolve conflicts :smile:

grigi:
:+1:

PR #302, opened 2020-02-28
Fix IntegrityError with unique field and get_or_create

### LINKED ISSUE #140: IntegrityError with unique field and get_or_create

grigi:
Fixes #140

Had to introduce an extra lock for nested transactions for DB's that support it to prevent network errors with both PostgreSQL & MySQL.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -74,6 +74,7 @@ future
 
 - Fix default type of ``JSONField``
 - Install on Windows does not require a C compiler any more.
+- Fix ``IntegrityError`` with unique field and ``get_or_create``
 
 0.15.17
 -------
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -29,7 +29,6 @@ class TestConcurrencyIsolated(test.IsolatedTestCase):
         for una in unas:
             self.assertEqual(una[0], unas[0][0])
 
-    @test.expectedFailure
     async def test_concurrent_get_or_create(self):
         unas = await asyncio.gather(*[UniqueName.get_or_create(name="d") for _ in range(10)])
         una_created = [una[1] for una in unas if una[1] is True]
@@ -76,7 +75,6 @@ class TestConcurrencyTransactioned(test.TestCase):
         for una in unas:
             self.assertEqual(una[0], unas[0][0])
 
-    @test.skip("Crashes with MySQL & PostgreSQL")
     async def test_concurrent_get_or_create(self):
         unas = await asyncio.gather(*[UniqueName.get_or_create(name="b") for _ in range(10)])
         una_created = [una[1] for una in unas if una[1] is True]
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -29,6 +29,7 @@ class TestConcurrencyIsolated(test.IsolatedTestCase):
         for una in unas:
             self.assertEqual(una[0], unas[0][0])
 
+    @test.skipIf(sys.version_info < (3, 7), "aiocontextvars backport not handling this well")
     async def test_concurrent_get_or_create(self):
         unas = await asyncio.gather(*[UniqueName.get_or_create(name="d") for _ in range(10)])
         una_created = [una[1] for una in unas if una[1] is True]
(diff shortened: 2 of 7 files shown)


grigi:
@abondar based this on your original attempt to fix this. Please review.

grigi:
Wow, Py3.6 gets so confused with contexts :sob: 
I feel we may just have to remove it from CI altogether...
It has been deprecated for a few months now...

Unfortunately it would mean dropping pypy too.

I'll just conditionally skip the test for now.

grigi:
Another option is to disable all connection pooling on py3.6?

AEnterprise:
wow didn't expect pooling to be such a major pain on 3.6

but personally i'm more in favor for dropping only specific features like pooling then dropping an entire version (prob best to log a warning on startup though so people are aware)

grigi:
Py3.6 uses a contextvar backport, and unfortunately it has shortcomings. py3.7+ has extra state in their asyncio C-structs to keep track better.
I'm busy seeing what would be the lowest-impact change for disabling pooling for py3.6. It seems forcing a pool of 1 doesn't do it.
I suspect I would have to use explicit locking like I do for SQLite to get it working right.

AEnterprise:
i think forcing a shared lock instead of making new locks for concurrent operations should do the trick

that and forcing a poolsize of 1 regardless since it can't use multiple so there is no reason to have any extras open

grigi:
Well, in the meantime I'm getting hillarious errors, like "cannot rollback; the transaction is not yet started" but investigating further the cause is the dreaded "cannot perform operation: another operation is in progress".

And fixing race conditions fixes that.

grigi:
@AEnterprise We are discussing the potential deprecation for py3.6 in #303
This PR fixes a bug, and should go in soon.
I'll merge this to master instead of develop, and cherry-pick the windows fix as well and release that. All the new feature work is going in as v0.16

grigi:
I'll cherry pick the commits instead.

PR #305, opened 2020-03-02
Improve code

lntuition:
This is minor patch to improve code better. This patch is not related to any feature.

* [x] refactor execute in `queryset` to remove duplicate code
* [x] remove duplicate assignment at init relation
* [x] remove duplicate Exception in relational field

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -183,8 +183,7 @@ class QuerySet(AwaitableQuery[MODEL]):
         "_prefetch_map",
         "_prefetch_queries",
         "_single",
-        "_get",
-        "_count",
+        "_raise_does_not_exist",
         "_db",
         "_limit",
         "_offset",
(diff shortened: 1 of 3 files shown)


tortoise/queryset.py
```diff
@@ -479,7 +475,8 @@ def get(self, *args: Q, **kwargs: Any) -> QuerySetSingle[MODEL]:
         """
         queryset = self.filter(*args, **kwargs)
         queryset._limit = 2
-        queryset._get = True
+        queryset._single = True
+        queryset._raise_does_not_exist = True
```
grigi: Should you not also set it for `get_or_none`?

lntuition: I think `get_or_none` do not raise `DoesNotExist`. it will return None

lntuition:
Also I don't know why CI failed with pypy...

tortoise/queryset.py
grigi: Doh, yet, you are right. It should however raise an exception if more than one record gets returned. I got confused about that.

grigi:
The one test is a bit fragile, and timing dependant. Sometimes Travis has a hiccup on that one test. I restarted it, it should just go away soon.

Issue #304, opened 2020-03-01
Need checking limit value is negative


lntuition:
I found behavior of negative value in Limit clause is depends on db. With below links,

- [Mysql](https://stackoverflow.com/questions/9852956/what-does-using-a-negative-value-for-mysql-limit-do) : ERROR. It will not working well.  
- [Sqlite](https://www.sqlite.org/lang_select.html) : It considered that there is no upper bound on the number of rows returned.

So do we need to consider each db's behavior and should we make it behave differently?
Or maybe Just checking that limit value is negative and give warning or error can be useful (and also i think this can be general solution. Because it is hard to know how each db behaves)
What do you think about this? If agree, i will make patch for this feature 😄

grigi:
Negative limit, and negative offset doesn't really make sense. I think it would be correct to raise error in those cases?

lntuition:
@grigi I agree. it can lead to inconsistent behavior. I will raise error with that and add negative test for checking that scenario.

PR #306, opened 2020-03-02
Add validation with limit and offset

lntuition:
Add validation to limit and offset's param. It should be non-negative number. If it is negative number then raise `ParamsError`

Fixed: #304 

* [x] Add validation to limit and offset's param
* [x] Add negative testcase
* [x] Add to CHANGELOG.rst


grigi:
Looks good! Could you please add two one-liners to the changelog?

lntuition:
Oh, i missed it. I updated it !

grigi:
:+1:

Issue #310, opened 2020-03-06
Function with source_field fields does not work well


lntuition:
```python
class TestFields(Model):
    ch = fields.CharField(max_length=100, source_field='source_ch')

async def run():
    await TestFields.create(ch="Apple")
    await TestFields.create(ch="Apple   ")
    values = await TestFields.annotate(a=Trim("ch"), b=Length("ch")).values("a", "b")
    print(values)
```

**To-be**
```console
[{'a': 'Apple', 'b': 5}, {'a': 'Apple', 'b': 8}]
```

**As-is**
``` console
[{'a': 'ch', 'b': 2}, {'a': 'ch', 'b': 2}]
```

Also, It only works at sqlite. postgresql looks like treating generated query as an error.

lntuition:
Also I think this is looks messy bug. I will make patch to hotfix this with below workaround
```diff
--- a/tortoise/functions.py
+++ b/tortoise/functions.py
@@ -59,7 +59,11 @@ class Function:
                 function_joins.append(join)
                 field = related_field_meta.basetable[related_field_meta.db_pk_field]
             else:
-                field = table[field_split[0]]
+                field = model._meta.fields_map.get(field_split[0])
+                if field.source_field:
+                    field = table[field.source_field]
+                else:
+                    field = table[field_split[0]]
                 if self.populate_field_object:
                     self.field_object = model._meta.fields_map.get(field_split[0], None)
                     if self.field_object:  # pragma: nobranch
```

lntuition:
However, I think this type of bug can exist anywhere that creates query (also current workaround looks not beautiful) I think we should look all query generating code and may refactor that?.. Maybe It could be a little difficult :worried: 

How about your thinking? @grigi

grigi:
Woah, that's a bad one. I know to root this out with regular fields I did the whole `StraightFields`/`SourceFields` tests. Basically I have 2 models that are supposed to be identical from the Tortoise model interface, but the one has translation to the DB.

We may have to extend that with a lot more test cases?

grigi:
We still need to identify anything that works with pypika objects directly, (I don't think `Table`, but the others) and do a straight/source test on all of them.

PR #311, opened 2020-03-06
Fix Function with source_field

lntuition:
Current `Function` is not worked well with `source_field` option. After applying this patch `Function` will be work well with `source_field` option

Fixed: #310 

* [x] Fix `Function` with `source_field` option
* [x] Add testcase
* [x] Update CHANGELOG


grigi:
Oh dear, I just had a look at how nice and neat https://github.com/tortoise/tortoise-orm/blob/develop/tests/test_source_field.py is, and then couldn't find the annotation test cases, and here they are: https://github.com/tortoise/tortoise-orm/blob/develop/tests/test_filtering.py#L148-L345

This is a bit messy.

If we can clean up the tests some more, we can try and make the same tests run on both default/sourced fields. Just like how test_source_field.py just subclasses the earlier test and replaces the objects to test against.

lntuition:
@grigi I add testcase for `Functions`. Could you review once more? (I delete comment from coveralls, because it is duplication. I don't know why it works twice)

tortoise/functions.py
```diff
@@ -59,9 +59,13 @@ def _resolve_field_for_model(
...
             else:
-                field = table[field_split[0]]
+                field_object = cast(Field, model._meta.fields_map.get(field_split[0]))
+                if field_object.source_field:
+                    field = table[field_object.source_field]
+                else:
+                    field = table[field_split[0]]
                 if self.populate_field_object:
-                    self.field_object = model._meta.fields_map.get(field_split[0], None)
+                    self.field_object = model._meta.fields_map.get(field_split[0])
```
grigi: If you don't want a default, then isn't using `[]` easier to read? Also, I think, faster. (Mostly because `[]` is builtin, and `.get()` is a method call)

lntuition: Okay :) i will fix it

grigi:
:+1:

grigi:
I backported this fix to the 0.15 series for inclusion in next bugfix release.

Issue #296, opened 2020-02-22
Is there any solution for COUNT(DISTINCT, column) ?


pjongy:
**Is your feature request related to a problem? Please describe.**
I've feel something strange while using tor 'tortoise.functions.Count' that does inherited Aggregate.
For my expectation, Count should be inherit can represent with distinction (ex. DistinctFunction)

For instance, pypika that tortoise is using for query genrerating's Count inherits 'DistinctOptionFunction' and we can use it with distinct builder like this:
```
fn.Count('*').distinct()
```

but tortoise does not admit distinct builder because of structure that follows below:
tortoise.function.Function
```
function_field = self.database_func(field, *default_values)
return {"joins": function_joins, "field": function_field}
```

**Describe the solution you'd like**
Count function can be used with distinct flag like this:
```
class DistincFunction:
    def __init__(..., is_distinct):
    ...
    if is_distinct:
        function_field = self.database_func(field, *default_values).distinct()
...
class Count(DistinctFunction):
    database_func = functions.Count
```

**Describe alternatives you've considered**
.

**Additional context**
.

pjongy:
Now, I made some manual code and using distincted count in my project temporary like this:

```
from pypika import functions
from tortoise import ConfigurationError
from tortoise.functions import Function

class DistinctFunction(Function):
    def __init__(self, field, is_distinct, *default_values) -> None:
        super().__init__(field, *default_values)
        self._is_distinct = is_distinct

    def _resolve_field_for_model(self, model, field: str, *default_values) -> dict:
        field_split = field.split("__")
        if not field_split[1:]:
            function_joins = []
            if field_split[0] in model._meta.fetch_fields:
                related_field = model._meta.fields_map[field_split[0]]
                related_field_meta = related_field.model_class._meta
                join = (model._meta.basetable, field_split[0], related_field)
                function_joins.append(join)
                field = related_field_meta.basetable[related_field_meta.db_pk_field]
            else:
                field = model._meta.basetable[field_split[0]]

                if self.populate_field_object:
                    self.field_object = model._meta.fields_map.get(field_split[0], None)
                    if self.field_object:
                        func = self.field_object.get_for_dialect(
                            model._meta.db.capabilities.dialect, "function_cast"
                        )
                        if func:
                            field = func(self.field_object, field)
            if self._is_distinct:
                function_field = self.database_func(field, *default_values).distinct()
            else:
                function_field = self.database_func(field, *default_values)
[...]

PR #312, opened 2020-03-07
Add is_distinct to Aggregate (#296)

pjongy:
Enhanced Aggregate class for distinct option

## Description

For now, we can use Count function with:
Count('field', is_disitnct=True)
and then it might generate SQL like:
COUNT(DISTINC field)
by pypika:
Count().distinct()

## Motivation and Context

There were no solution for count with distinct about aggregated result

## How Has This Been Tested?

Execute count function with distinct flag in aggregating multiple backward relation situation.

## Checklist:


tortoise/functions.py
```diff
@@ -110,10 +113,32 @@ def resolve(self, model: "Type[Model]", table: Table) -> dict:
...
+    :param is_distinct: Flag for aggregate with distinction
     """
 
     database_func = AggregateFunction
 
+    def __init__(
+        self,
+        field: Union[str, F, ArithmeticExpression],
+        *default_values: Any,
+        is_distinct=False,
```
grigi: @lntuition Now here is a valid use-case to keep `Aggregate` separate from `Function`.
A function like `Length`, `Trim`, etc... doesn't make sense with a `distinct`, but aggregates really do.

grigi: @pjongy I feel an interface of `Count('somecol', distinct=True)` would read/write better.

grigi:
Thank you for the PR @pjongy 
Other than the minor comment re `distinct` vs `is_distinct` the code looks good.
Could you please do a test? e.g. just confirm that the result changes correctly when distinct changes from `False` to `True`.
e.g. add multiple records, do an count() without and with a distinct, and check the result is correct (and different)

Also, please add a short feature description in the `CHANGELOG.rst` under section 0.16.0
and add yourself to the `CONTRIBUTORS.rst`

pjongy:
First of all, thanks for your review

I accept your suggestion(distinct=True) is more reasonable. (FIxed)
I've add some test for COUNT(DISTINCT, column) (Added)

But while I'm working on, I found something going wrong in relation joining.
(Occurs same in CI: https://travis-ci.com/tortoise/tortoise-orm/jobs/295745089)

I think cause of test fails is wrong condition chain flows in while manuplate JOIN ~ ON condition
I expected (and it work in tortoise==0.15.8)
``` SQL
SELECT "post"."name","post"."id",COUNT(DISTINCT "comment"."id") "comments_count",COUNT(DISTINCT "favorite"."id") "favorites_count" FROM "post" LEFT OUTER JOIN "comment" ON "post"."id"="comment"."post_id" LEFT OUTER JOIN "favorite" ON "post"."id"="favorite"."post_id" WHERE "post"."id"=2 GROUP BY "post"."id" LIMIT 1
```
but real work ( == 0.15.9 upper)
``` SQL
SELECT "post"."name","post"."id",COUNT(DISTINCT "comment"."id") "comments_count",COUNT(DISTINCT "favorite"."id") "favorites_count" FROM "post" LEFT OUTER JOIN "comment" ON "post"."id"="comment"."post_id" LEFT OUTER JOIN "favorite" ON "comment"."id"="favorite"."post_id" WHERE "post"."id"=2 GROUP BY "post"."id" LIMIT 1
```
expect: LEFT OUTER JOIN "favorite" ON "post"."id"="favorite"
real: LEFT OUTER JOIN "favorite" ON "comment"."id"="favorite"

I think https://github.com/tortoise/tortoise-orm/commit/55618fa43cb18a9bd3db0a8959a40496a1c6f6f5 commit could affect in this situation
So, can you check this bug(?) for detail? @grigi 

Thank you.

grigi:
I will have a look during the week. Thank you for finding where the regression happened, it will make it easier to fix the issue.

lntuition:
@grigi @pjongy I found error point and fixed it. I'll share information after do my personal work.

grigi:
Wow, that was fast. I was planning to look now.

lntuition:
There was some wrong table assignment to `_resolve_annotate` method in `queryset`
With #316, I remove wrong table assignment and also checked it works well.
So just merge develop branch will remove this issue.

And also i have some reviews :) plz check it. Thank you for your hard work :smile:

CHANGELOG.rst
```diff
@@ -98,6 +98,10 @@ New features:
 - Allow usage of ``F`` expressions to in annotations. (#301)
 - Now negative number with ``limit(...)`` and ``offset(...)`` raise ``ParamsError``. (#306)
 
+* Add distinct to Aggregate
+
+  Functions that inherits Aggregate can use with ``distinct`` flag
+
```
lntuition: This CHANGELOG is also technically correct. but I think average user will not see the implication. 
Rather say `Add ability to supply ``distinct`` flag to Aggregate`? Just one line looks enough.

tests/testmodels.py
```diff
@@ -637,3 +637,26 @@ class Principal(Model):
...
+    post: fields.ForeignKeyRelation[Post] = fields.ForeignKeyField(
+        "models.Post", related_name="comments"
+    )
+
+
+class Favorite(Model):
+    id = fields.IntField(pk=True)
+    post: fields.ForeignKeyRelation[Post] = fields.ForeignKeyField(
+        "models.Post", related_name="favorites"
+    )
```
lntuition: I think adding new models is not essential. Also Current test model is enough huge. How about reusing `DoubleFK` model instead?

tests/test_aggregation_with_distinct.py
```diff
@@ -0,0 +1,40 @@
...
+
+        school_with_count = (
+            await Post.filter(id=post.id)
+            .annotate(comments_count=Count("comments"), favorites_count=Count("favorites"))
+            .first()
+        )
+
+        self.assertEqual(school_with_count.comments_count, 6)
+        self.assertEqual(school_with_count.favorites_count, 6)
+        # comments_count and favorites_count will same if no distinct
```
lntuition: We do not need to test this here. It is part of aggregation not aggregation with distinct. How about removing this? And also I think adding new TestClass is not essential. How about moving this testsuite to `test_aggregation.py` (inside of `TestAggregation`) ?

CHANGELOG.rst
pjongy: Done.
Thx for your suggestion

tests/testmodels.py
pjongy: I think models we need are "multiple backward relation" not "multiple foriegn key relation" that is your suggestion. So, how about using just `Tournament` that has two backward relation?
(And your model redundant model removing opinion is acceptable. - Done)

tests/test_aggregation_with_distinct.py
pjongy: Moving this test suite to test_aggregation is reasonable. - Done.
And my idea for asserting without `distinct` was to show comparison opposites `distinct`, but I think it does not feat for test as your opinion. - Done.

tests/testmodels.py
lntuition: I misunderstand point. Anywhere change looks good :+1:

PR #316, opened 2020-03-10
Improve code

lntuition:
Current `_resolve_annotate` called resolve method twice per element.
Minor patch for remove unnecessary method call for performance.
And also remove unused return value from `_join_table_by_field`

@grigi Just waiting code review and ci status :) and then i will merge it to develop

* [x] Remove unnecessary method call
* [x] Remove unused return value
* [x] Check it works with existing testsuite

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -576,17 +576,19 @@ class QuerySet(AwaitableQuery[MODEL]):
     def _resolve_annotate(self) -> None:
         if not self._annotations:
             return
+
         table = self.model._meta.basetable
-        if any(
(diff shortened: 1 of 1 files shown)


lntuition:
@grigi Yeah, I found that problem so i fixed. Also micro benchmark can be useful to check bottleneck or trash code. But I think we do not need at current level. We still have a lot of things to implement and fix. Thank you for reviewing :smile:

grigi:
Well, I have to say, profilers find surprising things.
Also, from the first benchmark to now, tortoise got a lot faster (over 100x!) And the worst offence was a sync issue where we waited on ourself...
And accidentally not using our prepared statement cache, etc...

PR #317, opened 2020-03-12
Improve code

lntuition:
Current `_resolve_field_for_model` in `Function` logic is unidirectional.
So it does not need to use recursive. This is minor patch for remove
unnecessary method call stack, parameter passing, split string and etc..

@grigi I think this patch is better code. But maybe other doesn't think so. Plz check it :)

- [x] Change logic from recursive to iterative
- [x] Check it works with existing testsuite

### DIFF
--- a/tortoise/functions.py
+++ b/tortoise/functions.py
@@ -54,46 +54,45 @@ class Function:
     def _resolve_field_for_model(
         self, model: "Type[Model]", table: Table, field: str, *default_values: Any
     ) -> dict:
-        field_split = field.split("__")
-        if not field_split[1:]:
-            function_joins = []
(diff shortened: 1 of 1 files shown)


grigi:
It's certainly easier to grok, so, yes, it is better.
Often I would roll this changes up with whatever feature I want it to enable? Or if it is a cleanup, then it is fine.

We need to allow ourselves to do constant small clean ups to minimise tech debt, else it will become a monster soon enough.

I did a quick line count:
Code: 5953 loc
Tests: 8890 loc
Docs: 962(examples) + 1552(docstrings) + 1530(rst) = 4044 loc
Interesting how large docs has gotten, I did not expect that.

PR #318, opened 2020-03-13
Started working on fixes for FastAPI & pydantic

### LINKED ISSUE #188: can you create a example for fastapi？

**Additional context**
Add any other context about the feature request here.

grigi:
Once I started building an "ideal" fastapi sample I started running into many small issues.
This PR is to get them ALL sorted out.

Fixes #188 

* [x] We need a better way of handling List[] of item. 
* [x] Optional auto handling of common `DoesNotExist` error in a generic way (404)
* [x] Optional auto handling of common `IntegrityError` error in a generic way (422)
* [x] For create (and possibly update) we need to exclude read-only fields. https://fastapi.tiangolo.com/tutorial/extra-models/ proposes using inheritance to do that, but now we have to return multiple models, e.g. read-model & write-model. 
* [x] Basic CRUDL example
* [x] Update docs

Moved out:
* Pydantic model should handle basic validation correctly
  * default is non-nullable and non-required.
  * read only fields - auto datetimes?
  * nullable/required — data & FK/O2O fields (They are not the same, but Pydantic treats them too similarily. The issue tracker also doesn't show any urgency on the movement so it's easier falling in line with the simpler definition)
  * basic int limits
  * basic str limits

Removed from scope:
* Support field/computed aliases
* Model names must be globally unique
* Relational model in example


grigi:
@AEnterprise @jet10000 Could you please look at the FastAPI example:
* https://github.com/tortoise/tortoise-orm/pull/318/files#diff-0330a02818bd9feacb3ac2382c45c6f4
* https://github.com/tortoise/tortoise-orm/pull/318/files#diff-4a73601108ada4c0ba8c66bd1ed0c916

I'm calling it done for this simpler example (and changes to make using FastAPI easier). After this I'll have a look at a more comprehensive example.

AEnterprise:
@grigi that looks so much cleaner then the hoops i was jumping through to do the same, thanks a ton!

with this you can do a the basic CRUD endpoints, and in my experience that should cover most basic usecases

jet10000:
If make a auth example, which supports cookie and token authentication, it will attract more users.

grigi:
> @grigi that looks so much cleaner then the hoops i was jumping through to do the same, thanks a ton!
> 
> with this you can do a the basic CRUD endpoints, and in my experience that should cover most basic usecases

I was surprised at how many hoops I had to jump through, and how small changes made it a lot better.


> If make a auth example, which supports cookie and token authentication, it will attract more users.
> 
> https://frankie567.github.io/fastapi-users/

Possibly that should be a library? Or do you see it as part of the fastapi extension?

jet10000:
This link is just for reference, I think tortoise-orm should build such a use case by itself. In fact, any modern application orm is the core.If the tortoise-orm official can design authentication library or use case base on tortoise-orm, it will be more valuable.

grigi:
Well, if it has to be an auth library, it should be `allauth`, right?
Because for it to be useful it needs to support local (securely), federated, and then session/cookie/jwt/bearer auth tokens. That way it can do auth for applications too?

(Sorry, I got carried away, but auth is quite a hot topic)

I feel it is too big a topic to just do an example, but better to do a whole library? I mean, look at how big a topic auth is for Django (https://docs.djangoproject.com/en/3.0/topics/auth/passwords/), and then it gets greatly extended by e.g. `django-allauth` to make it sane for API's/modern systems.

jet10000:
Yes, it is a big topic, maybe a simplified version can be made, which can be improved gradually. I now use fastapi-users for basic registration and authentication. In any case, if there is such a tortoise-orm-based authentication library, it will greatly attract more users.

grigi:
So, what is the concern with fastapi-users? That it is not so easy to set up? Or do you want a auth system that works with other frameworks?

jet10000:
fastapi-users is currently the fastest I can get started, and the only one that supports tortoise-orm. I just think that if tortoise-orm has such a use case example, it can be more convenient and easy to use for customization and getting started.

grigi:
Ah, I get it. Yes, making the example do a full user-management flow would be useful.

Issue #309, opened 2020-03-05
Is there any solution for filter in Aggregate function?


pjongy:
**Is your feature request related to a problem? Please describe.**
In my project, I need filter while using Count function like this
(It is an example code that is not really in project code):
``` python
User.first()
.preferch_ralated(
    'post'
)
.annotate(
    post_count=Count('post')  # We need to filter post.status is NORMAL
)
```

In my mind, It can be achieved with SQL(pseudo):
``` sql
SELECT COUNT(CASE WHEN post.status = 0 THEN post.id ELSE null END)
FROM user
INNER JOIN post ON post.user_id = user.id
```
and pypika:
``` python
Count(
    Case().when(board.status == NORMAL, board.id).else_(None)
)
```

django.db.models has filter argument in Aggregate for filtering while SQL function runs
(https://github.com/django/django/blob/828e3b1335e2614d338a630fd5b5f88d38a6b5d2/django/db/models/aggregates.py#L16)

And usage in django:
(https://docs.djangoproject.com/en/2.2/ref/models/conditional-expressions/#conditional-aggregation)
``` python
>>> Client.objects.aggregate(
...     regular=Count('pk', filter=Q(account_type=Client.REGULAR)),
...     gold=Count('pk', filter=Q(account_type=Client.GOLD)),
...     platinum=Count('pk', filter=Q(account_type=Client.PLATINUM)),
... )
```

**Describe the solution you'd like**
I think functions.Aggregate can recieve parameter 'filter' like this:
``` python
class Aggregate(Function):
    database_func = AggregateFunction
    def __init__(self, field, _filter: Q=None, *default_values) -> None:
        self.field = field
        self.field_object: Any = None
        self.default_values = default_values
        self.filter = _filter
    ...
```

and I expect enhanced sql function usage is:
``` python
User.first()
.preferch_ralated(
    'post'
)
.annotate(
    post_count=Count('post', _filter=Q(status=NORMAL)) 
)
```

**Describe alternatives you've considered**
.

**Additional context**
.

grigi:
The proposed API does make sense. Right now we don't have anything like that, unfortunately.

@lntuition Could you look at this together with #257 or should we do this completely separately?

lntuition:
I think it is different with #257. As is we can't do like above action in tortoise atomically. Current work around it use `filter` with `annotate` multiple times. (but it is also too inefficient).

After doing #257, I should check this, too.

lntuition:
[...]

PR #320, opened 2020-03-16
[WIP] Allow filter to aggregate function

lntuition:
Allow `Q` for filter of aggregate Functions.

Fixed : #309 

* [x] Allow _filter param to aggregate
* [x] Add testcase
* [ ] Update Changelog

### DIFF
--- a/tests/test_aggregation.py
+++ b/tests/test_aggregation.py
@@ -89,8 +89,8 @@ class TestAggregation(test.TestCase):
             .annotate(
                 all=Count("events", _filter=Q(name__in=["New Tournament"])),
                 no=Count("events", _filter=Q(name__not="New Tournament")),
-                event1=Sum("events__event_id", _filter=Q(events__name="Event 1")),
-                event2=Min("events__event_id", _filter=Q(events__name__not="Event 1")),
+                event1=Count("events", _filter=Q(events__name="Event 1")),
+                not_event1=Count("events", _filter=Q(events__name__not="Event 1")),
             )
             .first()
         )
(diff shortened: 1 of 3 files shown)


lntuition:
I Found current solution. but It's not beautiful. I think join to Pypika object sometimes use `_get_joins_for_related_field` and sometimes do not use. I really make confusing. If we can, choosing one solution is more better. looks like current Q object is too oriented only filter option in queryset

grigi:
I see what you mean. Agreed. `Q` is very filter oriented, and now the functions code has a lot of similarities with it. And behaviour isn't entierly consistent.

How do you think we should resolve this? Should we mark it as technical debt with big `# TODO:` statements (as in get to it later, preferably once we have an eagle's eye view) or try and fix it up-front. Note that we have to be careful with breaking backwards compatibility. (Although I think this is an implementation detail, it's not going to have a large fallout)

lntuition:
Ah... I think we have big debt.. we should devide into small units.. and do hard refactor of that.. We need to prune many code. As you said, with pruning can break backwards compatibility. Also marking TODO can be good, too..

grigi:
I vote to put TODOs now, and possibly treat it as a larger project. E.g. like fixing up Fields that was the driver between 0.14 and 0.15

lntuition:
And does `Q` works different with database? I think it works sqlite good. but not working with postgres

grigi:
> And does `Q` works different with database? I think it works sqlite good. but not working with postgres

It should work everywhere... I mean we run the entire test suite for each DB?

lntuition:
hmm... i need to check why ci failed. maybe my testcase is wrong?

lntuition:
I found what i wrong. but i need to think more to support this feature. I left this PR to WIP for a while

Issue #327, opened 2020-03-22
After version 0.16.1 aggregation in aggregate classes works wrong


pjongy:
**Describe the bug**
Post versions of 0.16.1 could cover _ _ aggregateing in annotate method but after 0.16.1 there are no more work as expect (ps. annotate is not root cause it is my example)

**To Reproduce**
Steps to reproduce the behavior, preferaby a small code snippet.

version < 0.16.1
``` python
tournaments = await Tournament.annotate(
    events_participants_count=Count("events__participants")
).filter(id=tournament.id)
print(tournaments.events_participants_count)
```
version \>= 0.16.1
``` python
tournaments = await Tournament.annotate(
    events_participants_count=Count("events__participants")
).filter(id=tournament.id)
print(tournaments.events_participants_count)  # Error
```

exception like this:
```
  File "/storage/tortoise-orm/tortoise/queryset.py", line 604, in _resolve_annotate
    self._join_table_by_field(*join)
  File "/storage/tortoise-orm/tortoise/queryset.py", line 121, in _join_table_by_field
    self.query = self.query.join(join[0], how=JoinType.left_outer).on(join[1])
  File "/root/.pyenv/versions/3.7.6/envs/py37/lib/python3.7/site-packages/pypika/queries.py", line 1391, in on
    self.query.do_join(JoinOn(self.item, self.how, criterion, collate))
  File "/root/.pyenv/versions/3.7.6/envs/py37/lib/python3.7/site-packages/pypika/queries.py", line 969, in do_join
    join.validate(base_tables, self._joins)
  File "/root/.pyenv/versions/3.7.6/envs/py37/lib/python3.7/site-packages/pypika/queries.py", line 1485, in validate
    tables=", ".join(map(str, missing_tables))
pypika.utils.JoinException: Invalid join criterion. One field is required from the joined item and another from the selected table or an existing join.  Found ["event"]
```

**Expected behavior**
version < 0.16.1
``` python
tournaments = await Tournament.annotate(
    events_participants_count=Count("events__participants")
).filter(id=tournament.id)
print(tournaments.events_participants_count)
```

**Additional context**
I think culprit commit is `42d85cb84f6684e7364446fc30ec930f0fd60ac4`
(This is not for pressing commiter's responsibility but for more effective bug fix as information)

Lastly, I recommend you to add below test code
``` python
    async def test_nested_aggregation_in_annotation(self):
        tournament = await Tournament.create(name="0")
[...]

PR #328, opened 2020-03-22
Hotfix for aggregation

lntuition:
We found #317 change `joins` order in aggregation. So removing reversed operation may fix this bug(and fortunately, it leads to better performance) Add more testcase to check aggregation. 

Special thank to detailed bug reporting @pjongy 
Fixed : #327 

- [x] Fix bug in join with aggregation
- [x] Add more testcase to aggregation 
- [x] Add Changelog

### DIFF
--- a/tests/test_aggregation.py
+++ b/tests/test_aggregation.py
@@ -50,6 +50,23 @@ class TestAggregation(test.TestCase):
         with self.assertRaisesRegex(ConfigurationError, "name__id not resolvable"):
             await Event.all().annotate(tournament_test_id=Sum("name__id")).first()
 
+    async def test_nested_aggregation_in_annotation(self):
+        tournament = await Tournament.create(name="0")
+        await Tournament.create(name="1")
+        event = await Event.create(name="2", tournament=tournament)
+
+        team_first = await Team.create(name="First")
+        team_second = await Team.create(name="Second")
+
+        await event.participants.add(team_second)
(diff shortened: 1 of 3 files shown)


lntuition:
@pjongy Could you check that this PR fix your problem?

pjongy:
Thank you for your quick response and hotfix.
It seems bug is fixed and looks good to me! :)

tests/test_aggregation.py
```diff
@@ -50,6 +50,23 @@ class TestAggregation(test.TestCase):
...
+
+        await event.participants.add(team_second)
+        await event.participants.add(team_first)
+
+        tournaments = await Tournament.annotate(
+            events_participants_count=Count("events__participants")
+        ).filter(id=tournament.id)
+        self.assertEqual(len(tournaments), 1)
+        self.assertEqual(tournaments[0].id, tournament.id)
+
```
pjongy: I've copied exsitence test code for skeleton so... we need to check about our requirements
I think check assert for `events_participants_count` realy `2` is reasonable
``` python
self.assertEqual(tournaments[0].events_participants_count, 2)
```

pjongy:
Oh, I've reviewed something minor can you check it?

tests/test_aggregation.py
lntuition: Good point. I fixed it

Issue #326, opened 2020-03-21
order_by nested looks buggy


sm0k:
**Describe the bug**
Order by on nested field seems buggy

**To Reproduce**
```
from tortoise.models import Model
from tortoise import fields

class Tournament(Model):
    id = fields.IntField(pk=True)
    name = fields.TextField()

    def __str__(self):
        return self.name


class Event(Model):
    id = fields.IntField(pk=True)
    name = fields.TextField()
    tournament = fields.ForeignKeyField('models.Tournament', related_name='events')
    participants = fields.ManyToManyField('models.Team', related_name='events', through='event_team')

    def __str__(self):
        return self.name


class Team(Model):
    id = fields.IntField(pk=True)
    teamname = fields.TextField()

    def __str__(self):
        return self.teamname

from tortoise import Tortoise

async def init():
    # Here we connect to a SQLite DB file.
    # also specify the app name of "models"
    # which contain models from "app.models"
    await Tortoise.init(db_url="sqlite://:memory:", modules={"models": ["__main__"]})
    # Generate the schema
    await Tortoise.generate_schemas()

    # Create instance by save
    tournament = Tournament(name='New Tournament')
    await tournament.save()

    # Or by .create()
    await Event.create(name='Without participants', tournament=tournament)
    event = await Event.create(name='Test', tournament=tournament)
    participants = []
    for i in range(2):
        team = await Team.create(teamname='Team {}'.format(i + 1))
        participants.append(team)

    # M2M Relationship management is quite straightforward
    # (also look for methods .remove(...) and .clear())
    await event.participants.add(*participants)

    # You can query related entity just with async for
    async for team in event.participants:
        pass

    # After making related query you can iterate with regular for,
    # which can be extremely convenient for using with other packages,
    # for example some kind of serializers with nested support
    for team in event.participants:
        pass

    # Or you can make preemptive call to fetch related objects
    selected_events = await Event.filter(
        participants=participants[0].id
    ).prefetch_related('participants', 'tournament')

    # Tortoise supports variable depth of prefetching related entities
    # This will fetch all events for team and in those events tournaments will be prefetched
    await Team.all().prefetch_related('events__tournament')

    # You can filter and order by related models too
    await Tournament.filter(
        events__name__in=['Test', 'Prod']
    ).order_by('-events__participants__teamname').distinct()
import asyncio
if __name__ == "__main__":
    asyncio.run(init())

```
[...]

PR #330, opened 2020-03-23
Nested order_by path + unittest

sm0k:
Patch for order_by method on nested field

## Description
`field_object = self.model._meta.fields_map.get(field_name)`
Becomes
`field_object = model._meta.fields_map.get(field_name) if self.model != model else self.model._meta.fields_map.get(field_name)`
in queryset.py

## Motivation and Context
https://github.com/tortoise/tortoise-orm/issues/326

## How Has This Been Tested?
See py.test tests/test_order_by_nested.py

## Checklist:

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -184,10 +184,7 @@ class AwaitableQuery(Generic[MODEL]):
                 annotation_info = annotation.resolve(self.model, table)
                 self.query = self.query.orderby(annotation_info["field"], order=ordering[1])
             else:
-                if self.model != model:
-                    field_object = model._meta.fields_map.get(field_name)
-                else:
-                    field_object = self.model._meta.fields_map.get(field_name)
+                field_object = model._meta.fields_map.get(field_name)
 
                 if not field_object:
                     raise FieldError(f"Unknown field {field_name} for model {self.model.__name__}")
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -187,7 +187,7 @@ class AwaitableQuery(Generic[MODEL]):
                 field_object = model._meta.fields_map.get(field_name)
 
                 if not field_object:
-                    raise FieldError(f"Unknown field {field_name} for model {self.model.__name__}")
+                    raise FieldError(f"Unknown field {field_name} for model {model.__name__}")
                 field_name = field_object.source_field or field_name
                 field = getattr(table, field_name)
 
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -184,7 +184,11 @@ class AwaitableQuery(Generic[MODEL]):
                 annotation_info = annotation.resolve(self.model, table)
                 self.query = self.query.orderby(annotation_info["field"], order=ordering[1])
             else:
-                field_object = self.model._meta.fields_map.get(field_name)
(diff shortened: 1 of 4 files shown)


tortoise/queryset.py
```diff
@@ -184,7 +184,10 @@ def resolve_ordering(
                 annotation_info = annotation.resolve(self.model, table)
                 self.query = self.query.orderby(annotation_info["field"], order=ordering[1])
             else:
-                field_object = self.model._meta.fields_map.get(field_name)
+                if self.model != model:
```
grigi: Would this not mean that when self.model == model the `model` var is still the correct one?
Just do the `field_object = model._meta.fields_map.get(field_name)` instead.

tortoise/queryset.py
```diff
@@ -184,7 +184,10 @@ def resolve_ordering(
...
+                    field_object = model._meta.fields_map.get(field_name)
+                else:
+                    field_object = self.model._meta.fields_map.get(field_name)
                 if not field_object:
                     raise FieldError(f"Unknown field {field_name} for model {self.model.__name__}")
```
grigi: This should be `{model.__name__}`

sm0k:
So I ran the following commands locally (from Makefile):
```
isort -rc tortoise/ examples/ tests/ conftest.py
black -l 100 --py36 tortoise/ examples/ tests/ conftest.py
```
Wich modified many files (see last commit)
Unit tests are all green on my computer
But Travis jobs are all red now :D 
Sorry im not familiar enough with these processes...

tortoise/queryset.py
sm0k: I tried to just replace self.model with model but didn't seems to work

lntuition:
Ah it's generated by auto styling.. I miss checked it ! sorry.

sm0k:
Looks good now I guess?

sm0k:
> Ah it's generated by auto styling.. I miss checked it ! sorry.
> p.s) @smok found that it is generated by wrong black version which should be fixed

Was in fact due to isort 4.3.4 on my box

tortoise/queryset.py
lntuition: I found that just change model is okay with TCs. How about remove unnecessary logic

``` diff
diff --git a/tortoise/queryset.py b/tortoise/queryset.py
index d09ce89..68f6647 100644
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -184,13 +184,9 @@ class AwaitableQuery(Generic[MODEL]):
                 annotation_info = annotation.resolve(self.model, table)
                 self.query = self.query.orderby(annotation_info["field"], order=ordering[1])
             else:
-                if self.model != model:
-                    field_object = model._meta.fields_map.get(field_name)
-                else:
-                    field_object = self.model._meta.fields_map.get(field_name)
-
+                field_object = model._meta.fields_map.get(field_name)
                 if not field_object:
-                    raise FieldError(f"Unknown field {field_name} for model {self.model.__name__}")
+                    raise FieldError(f"Unknown field {field_name} for model {model.__name__}")
                 field_name = field_object.source_field or field_name
                 field = getattr(table, field_name)

```

grigi:
Thanks for the great work!

PR #350, opened 2020-04-11
Partial model .only(...) support 

### LINKED ISSUE #125: Provide an API for only fetching certain database fields
Consider the following:

```
class Employee(Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(max_length=1024)
    ... 100+ other fields

async def get_employee_names():
    employees = await Employee.all()
    return [employee.name for employee in employees]
```

In this case, `get_employee_names` only needs a single field from the `employee` table, which is `name`. If `Employee` has a lot of fields in the table (as often occurs with mature databases), we serialize all those fields into memory even when we only need a single field. That's wasteful.

An improved API might be

```
async def get_employee_names():
    employees = await Employee.only('name').all()
    return [employee.name for employee in employees]
```

Django implements this API as documented here: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#only

grigi:
Fixes #125 

Done:
* [x] Populates model partially
* [x] raises ``AttributeError`` for unpopulated fields
* [x] raises ``IncompleteInstanceError`` when attempting to save field that isn't populated (meaning it works only when both PK and the ``update_fields`` are ALL available.
* [x] Works for ``source_field``'ed fields
* [x] Test for ``.filter()``, and not just ``.get()``
* [x] Document usage of ``.only()`` with caveats
* [x] Update changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,27 @@ Changelog
 
 0.16
 ====
+0.16.6
+------
+* Added support for partial models:
+
+  To create a partial model, one can do a ``.only(<fieldnames-as-strings>)`` as part of the QuerySet.
+  This will create model instances that only have those values fetched.
+
+  Persisting changes on the model is allowed only when:
+
+  * All the fields you want to update is specified in ``<model>.save(update_fields=[...])``
+  * You included the Model primary key in the `.only(...)``
+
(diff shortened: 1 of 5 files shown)


reedjosh:
Wow! You're on it. I love this project!

Issue #273, opened 2020-01-10
Model Signals


mojimi:
Does tortoise saves the old values before a save? So I can compare what has been changed.

The context here is I'm trying to create a "application trigger" module, for example :

```python
#When a task is done, all its child tasks should be automatically done too
@Task.bind('after-save')
async def update_child_tasks(self, old_values):
  if old_values.status != self.status:
    if self.status == 'done'
      await Task.filter(mother_task=self).update(status='done')
```
This is only a small example, there are many other cases where I need triggers like this

Hope it is understood what I'm trying to do, emphasis on trying, I'm accepting ideas on different ways to approach this.

Also it wouldn't work on batch QuerySet methods, like update or batch_create, if anyone has any idea on how to overcome this that'd be great

grigi:
Right now tortoise doesn't keep the old values unfortunately.

For batch update/delete you get a counter of how many records got affected, so if that is >0 something happened. But we don't know what the old value was.

Do you want a global trigger?

mojimi:
@grigi Can I somehow get the IDs of the affected records after a batch operation? That'd solve it.

I'm not sure what you mean by global trigger, I'm just trying to come up with a system similar do Django signals, or any sort of before/after bindings. 

I've managed to do it, but it won't work on batch operations yet

grigi:
@mojimi Unfortunately not.
SQLite reports the no of lines filtered, has no syntax to return other data during an update.
MySQL reports the no of lines updated, has no syntax to return other data during an update.
Postgres eports the no of lines filtered(not updated), and has syntax to return data.

In none of the cases we get the right data (or any data).

The best would be to run two queries in a transaction:
- Filter by parameters to `.filter()` and by NOT parameters to `.update()` → This should get you the object that is going to change in the next query
- Do the update

And depending on the database and the configured isolation level, would possibly raise an exception if any race gets detected.
So the data would be pretty close to trustworthy, but not guaranteed 100% trustworthy.


Django's signals is more-or-less what I was talking about :+1:

mojimi:
I guess for getting "old values" I could just re-query the object and compare changes, that would even be more accurate.

I don't know how PR works and my code is probably not standardized enough to be merged, but I can share what I came up with if that's useful

grigi:
Thank you, it would be useful to even drop it here in the comments :smile:

long2ice:
hey,I try to implement model signals in my own thought:
```python
class BaseModel(Model):
[...]

PR #354, opened 2020-04-17
add model signals support

long2ice:
## Description

add model signals support,current support ``pre_save``,``post_save``,``pre_delete``,``post_delete``.

## Motivation and Context

resolve [https://github.com/tortoise/tortoise-orm/issues/273](https://github.com/tortoise/tortoise-orm/issues/273)
## How Has This Been Tested?

added ``test_signals.py`` and all passed. 
## Checklist:


grigi:
Thanks for this!

I'm concerned about performance impact if not used with the two extra calls for save & delete. Possibly use `asyncio.gather()` to schedule them all at the same time?
Also, I have to test the impact as is. It might be practically non-existent.

Also, we need test cases for this? e.g. create a new model in `tortoise.tests.testmodels` and attach signals to that. Then test how that works?

And finally, docs & README update.

tortoise/models.py
```diff
@@ -724,6 +741,47 @@ def update_from_dict(self, data: dict) -> MODEL:
...
+        await asyncio.gather(*listeners)
+
+    async def _pre_save(
+        self,
+        using_db: Optional[BaseDBAsyncClient] = None,
+        update_fields: Optional[List[str]] = None,
+    ) -> None:
+        listeners = []
+        for listener in self._listeners.get(Signals.pre_save, {}).get(self.__class__, []):
+            listeners.append(listener(self.__class__, self, using_db, update_fields))
```
grigi, line 771: Why do you send the Class and the Instance for each callback?

long2ice: multiple Class can register same signal like ``pre_save(*senders)``,and in listener,we can know which Class send that signal,and get the Class instance,maybe we can know Class with ``instance.__class__`` and without Class pass in?

grigi: Also, all the class methods still exist on the instance.
One can do <instance x>.filter(...)
Although it doesn't read right if you do that.

tortoise/models.py
```diff
@@ -724,6 +741,47 @@ def update_from_dict(self, data: dict) -> MODEL:
         self._set_kwargs(data)
         return self  # type: ignore
 
+    @classmethod
+    def register_listener(cls, signal: Signals, listener: Callable):
+        assert callable(listener), "listener must be callable!"
```
grigi, line 746: Could you change this to an `if` that raises `tortoise.exceptions.ConfigurationError`?

long2ice: yeah,that's better.

tortoise/models.py
```diff
@@ -594,6 +610,7 @@ class Model(metaclass=ModelMeta):
 
     # I don' like this here, but it makes auto completion and static analysis much happier
     _meta = MetaInfo(None)  # type: ignore
+    _listeners: Dict[Signals, Dict[Type[MODEL], List[Callable]]] = {}
```
grigi, line 613: mypy complains about this. I thought about it a bit, and it's hint is just right.
Change the class definition to be:
```py
class Model(Generic[MODEL], metaclass=ModelMeta):
```
This will bound MODEL to the class of Model (not just the instance)

grigi:
It's looking good so far.
Just do the two small changes I requested, and see if `make style ci` runs locally.
Thank you!

long2ice:
why there SQL syntax in Travis CI?

grigi:
It seems to be MySQL specific? Weird.

grigi:
No, I get the same errors when running locally without the injection fixes. So it's not that.

grigi:
Got it. "signal" is a reserved word in MySQL. And the transactioned rollback doesn't quote...

long2ice:
Ok,so I need rename ``Signal`` model.

grigi:
I have a fix, which I'll push to develop. Just have to feed the kids (they have no patience).

grigi:
Please merge develop in, and it should fix that issue.

grigi:
Oh, no. I fixed it the issue. Sorry, we were out of sync.

grigi:
Thank you, I merged it in :+1:

PR #357, opened 2020-04-18
support python 3.9

lntuition:
We solved problem that pypika with python 3.9 version(https://github.com/kayak/pypika/pull/398)
And also pypika check python 3.9 with CI, so I think we can support python 3.9 too

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - 3.7
   - 3.8
+  - 3.9
 env:
   global:
     TORTOISE_TEST_MODULES=tests.testmodels
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 python:
   - 3.7
   - 3.8
-  - 3.9
+  - 3.9-dev
 env:
   global:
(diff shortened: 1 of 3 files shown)


lntuition:
Also i'm not sure that it works with Travis well .. Check and help plz. Thank u :smile:

lntuition:
hmm.. ci failed with mypy. Maybe is it related with https://github.com/python/mypy/issues/8627 ?

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 python:
   - 3.7
   - 3.8
+  - 3.9-dev
```
grigi, line 5: Consider replacing this with:
```yaml
matrix:
  include:
  - python: 3.9-dev
    env: TEST_RUNNER=green
    script: green
    after_success: echo
```
instead so we skip the linting until mypy gets fixed?

grigi: I edited the yaml above, use `make testall` instead, and drop the ENV. Copy Pasta mistake.

lntuition: That's good. I will try it !

grigi:
:tada: !

Issue #358, opened 2020-04-20
Implement conditional aggregation


EmilioCarrion:
**Is your feature request related to a problem? Please describe.**
As an user I want to be able to filter related objects before annotate a function

**Describe the solution you'd like**
A nice solution is the used by the DjangoORM described [here](https://docs.djangoproject.com/en/3.0/ref/models/conditional-expressions/#conditional-aggregation)
e.g `House.all().annotate(minors=Count('habitant', filter=Q(age__lt=18)))` where we count how many `minors` are in each `House`.

**Describe alternatives you've considered**
The example explained above cover my considerations

**Additional context**
N/A

lntuition:
duplicate with #309 Please look it. Current stage we didn't support that feature.

grigi:
Ah, right. Apologies @EmilioCarrion in telling you to open a ticket. Could you please follow #309 and we'll do it there

PR #362, opened 2020-04-21
Allow filter to aggregate function

lntuition:
Allow `Q` for filter of aggregate Functions.

Fixed : #309 

* [x] Allow `_filter` param to aggregate function
* [x] Add testcase
* [x] Add Example
* [x] Update Changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ Changelog
 0.16
 ====
 
+0.16.8
+------
+- Allow `Q` expression to function with `_filter` parameter
+
 0.16.7
 ------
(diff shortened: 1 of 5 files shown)


lntuition:
I close #320 while cleaning unused branch (Oops..) Also this feature is requested once more on #358. I think this solution is not bad in this stage. @grigi plz look once more. thank u :smile:

lntuition:
@grigi Sure. I add more TC with `source_field` option. And It also seems work well :smile:

grigi:
Cool, all looks good! I'm merging and doing releases :grin:

Issue #347, opened 2020-04-09
Explicit group by


mojimi:
Is it currently possible to do a explicit group by? I searched the docs but couldn't find anything about this.

The use I'm looking for is to get the count of each distinct value in a column, like the following query :

```sql
SELECT
  title,
  COUNT(status) AS `status_count`
FROM
  tasks
GROUP BY
  status
```

This would count the number of tasks of each status

grigi:
No, we don't have explicit group by at this stage, what syntax do you propose?

Part of the problem with an explicit group by is that we can't guarantee to get models, as the data may be skewed. So it could only work with `.values()` or `.values_list()`.

```py
await Tasks.all().annotate(status_count=Count('status')).values('title', 'status_count').group_by('status')
```

I think  a syntax like that makes sense to me? Would that work fine for you with the limitations specified?

mojimi:
> 
> 
> No, we don't have explicit group by at this stage, what syntax do you propose?
> 
> Part of the problem with an explicit group by is that we can't guarantee to get models, as the data may be skewed. So it could only work with `.values()` or `.values_list()`.
> 
> ```python
> await Tasks.all().annotate(status_count=Count('status')).values('title', 'status_count').group_by('status')
> ```
> 
> I think a syntax like that makes sense to me? Would that work fine for you with the limitations specified?

Well in my case it would definitely work just fine as I'm only looking for the count, but maybe its worth exploring more cases where this would be useful.

The syntax you proposed seems pretty good, is what I'd expect

PR #363, opened 2020-04-22
add group by support

long2ice:
## Description

add group by support
## Motivation and Context

resolve https://github.com/tortoise/tortoise-orm/issues/347.
## How Has This Been Tested?

add ``test_group_by.py`` to test.
## Checklist:


CHANGELOG.rst
```diff
@@ -13,6 +13,10 @@ Changelog
 ------
 - Fixed regression where ``GROUP BY`` class is missing for an aggregate with a specified order.
 
+0.16.8
```
grigi, line 16: Minor issue: There is already a 0.16.8 header, could you merge the two sections?

docs/examples/basic.rst
```diff
@@ -53,6 +53,15 @@ Functions
 
 .. _example_schema_create:
 
+Aggregate
+=========
+.. literalinclude::  ../../examples/aggregations.py
```
grigi, line 58: Is it right to call the example "aggregate"? It mostly is about adding explicit group_by statements?

tortoise/expressions.py
```diff
@@ -12,9 +12,7 @@
 class F(Field):  # type: ignore
     @classmethod
     def resolver_arithmetic_expression(
-        cls,
-        model: "Type[Model]",
-        arithmetic_expression_or_field: Union[ArithmeticExpression, Field],
+        cls, model: "Type[Model]", arithmetic_expression_or_field: Term,
     ) -> Tuple[Union[ArithmeticExpression, Field], Optional[Field]]:
```
grigi, line 16: Should the response type not also change to `Tuple[Term, Optional[Field]]`?

CHANGELOG.rst
long2ice: OK

docs/examples/basic.rst
long2ice: What about ``Group By``?

tortoise/expressions.py
long2ice: Yeah,my negligence

tortoise/queryset.py
```diff
@@ -664,6 +679,9 @@ def _make_query(self) -> None:
             self.query._offset = self._offset
         if self._distinct:
             self.query._distinct = True
+        self.resolve_ordering(
```
grigi, line 682: Note that I moved `self.resolve_ordering` to above `self.resolve_filters` to fix a bug in the previous commit. This is now specified twice.

Also, please have a look at `resolve_filters()` as it conditionally adds a default `order_by` at about line 100:
```py
        if has_aggregate and (self._joined_tables or having_criterion or self.query._orderbys):
            self.query = self.query.groupby(
                self.model._meta.basetable[self.model._meta.db_pk_field]
            )
```
Please ensure that when one specifies an group_by manually, that we don't create duplicate group bys.

docs/examples/basic.rst
grigi: I think that's fine :smile:

tortoise/queryset.py
long2ice: I'm not so clear about that,why there group by ``pk``?

tortoise/queryset.py
```diff
@@ -679,9 +679,7 @@ def _make_query(self) -> None:
             self.query._offset = self._offset
         if self._distinct:
             self.query._distinct = True
-        self.resolve_ordering(
-            self.model, self.model._meta.basetable, self._orderings, self._annotations
-        )
+        print(str(self.query))
```
grigi, line 682: Print debugging for the Win!

tortoise/queryset.py
grigi: It's for when an aggregation runs on a child model (e.g. counting no of children) it needs a group_by on the PK.
Or if you filter on it, it needs to be grouped else "having" won't see it.
The order_by doesn't make sense to me, but if you don' group by on PostgreSQL it fails with syntax error.

I would be happy to not automatically group_by PK at all if a manual group_by is specified, because then you are taking control from us, right?

tortoise/queryset.py
long2ice: Sorry,that's my time to get off work and lose checking...😅

grigi: We all do it, don't worry about it. :+1:

grigi:
Thank you for your good work! :-)

Issue #365, opened 2020-04-24
Filtering a database using __icontains fails for JSONB fields


odimko:
**Describe the bug**

In the previous version of the library (`tortoise-orm==0.16.4` namely), filtering of a database using `__icontains` on a JSONB field of a model would convert the field to `VARCHAR` and then search for the requested substring. However, for `tortoise-orm==0.16.8` that does not work and fails with an error:
```shell
function upper(jsonb) does not exist
HINT:  No function matches the given name and argument types.
You might need to add explicit type casts.
```

**To Reproduce**

The code to execute the query:
```python
phrases = ["a red apple", "a book", "the Sun", "a big store"]
query = Q()
for phrase in phrases:
    query = query & Q(poem__icontains=phrase)
    collection = collection.filter(query)
total_filtered = await collection.count()
```
where `collection` is a table that has the `poem` field, which in turn is of a JSONB type.
For `tortoise-orm==0.16.8` the raw query is (and it fails with the error above):
```sql
SELECT COUNT(*) FROM "collection" WHERE UPPER("phrases") LIKE UPPER(\'%a red apple%\') ESCAPE \'\\\
```

**Expected behavior**

```sql
SELECT COUNT(*) FROM "collection" WHERE UPPER(CAST("phrases" AS VARCHAR)) LIKE UPPER('%a red apple%')
```

**Additional context**

PostgreSQL is used for database handling.

lntuition:
Ah, I found `CAST` was removed at 733eaf3. I think we should revert it. Thank you for your report !

odimko:
> Ah, I found `CAST` was removed at [42adca3](https://github.com/tortoise/tortoise-orm/commit/42adca39dbfec65cd9d6cff9edb5c80651b3690d). I think we should revert it. Thank you for your report !

great, looking forward to the updated library :+1:

grigi:
Hi, I released 0.16.9 and it should have the fix for this issue.

Could you please test and confirm?

odimko:
yes, works great.
thank you both @lntuition and @grigi for the fast replies and action!

yoann9344:
This leads to a problem. I have postgres field "CHAR" (not variable) on which I have an index "bpchar_pattern_ops". And by casting to "VARCHAR", postgres doesn't consider the index anymore.

It would be nice to revert this casting. I've not found any solution yet with the current state of tortoise !

PR #366, opened 2020-04-24
Revert SQL casts for string comparisions

lntuition:
We removed SQL cast for string comparisions and it leads issue with Postgres.
So revive `CAST` to filter (So, this should be revert of 42adca3)

Fixed: #365

### DIFF
--- a/tortoise/filters.py
+++ b/tortoise/filters.py
@@ -3,7 +3,8 @@ from functools import partial
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Optional, Tuple
 
 from pypika import Table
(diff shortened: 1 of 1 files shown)


lntuition:
I'm not sure that this is best way to solve #365. This can be just hotfix for it. @grigi If you have more clever solution, could you explain ?

grigi:
I'm going to merge this and add the changelog entry so we can get the fix out with the next release. Thanks @lntuition :+1:

lntuition:
Ah, you already merge it. My Internet was wack for two days. Thank you for your working @grigi :+1:

PR #368, opened 2020-04-25
make IntEnumField accept valid int value and make CharEnumField accept valid str value

long2ice:
## Description

make ``IntEnumField`` accept valid int value and make ``CharEnumField`` accept valid str value

## How Has This Been Tested?

test create,update with ``int`` and ``str`` value in ``CharEnumField`` and ``IntEnumField``
## Checklist:


long2ice:
> Ah, that's an interesting convenience use case. 👍
> 
> Please add a changelog entry, and mention it's usage in the docs (probably fine to mention this in the docstrings) 😃

OK

grigi:
Thanks for the good work!

PR #375, opened 2020-04-30
fix: "sqlite://:memory:" in Windows

dongfangtianyu:
## Description
Catch OSError and OSError.errno ==22

## Motivation and Context
When use `"sqlite://:memory:" ` in Windows, 
```
OSError: [WinError 123] The filename, directory name, or volume label syntax is 
incorrect
```

## How Has This Been Tested?
I  tested it on my own computer
Only Linux in Travis Ci, which may not be automatically verified. 
But it's easy to test on all Windows .

## Checklist:

- [ x ] My code follows the code style of this project.
- [ x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,10 @@ Changelog
 
 0.16
 ====
+0.16.11
+-------
+- fix: ``sqlite://:memory:`` in Windows thrown ``OSError: [WinError 123]``
+
 0.16.10
 -------
 - Fix bad import of ``basestring``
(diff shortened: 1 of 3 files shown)


grigi:
So this just bypasses OSError nr 22 on windows?

Could you please add yourself to the `CONTRIBUTORS.rst` and a new `CHANGELOG.rst` entry under 0.16.11 with a short one-liner describing this fix.

OT: We need to add Windows to CI, I know travis has some windows support...

dongfangtianyu:
Yes, I think Windows first check filename syntax , so not thrown FileNotFoundError

lntuition:
@grigi Currently, we use `python` support in travis. and it is not available on the macOS and Windows env(Really too sad). For support various os env, we need to change travis to setup various dist(Linux, windows, macOS, etc..). I'll look into it if I have time :smile:

Issue #377, opened 2020-05-01
Escaping JSON data is incorrect for UPDATEs


gdmoore:
Tortoise 0.16.10, Postgresql 12, JSONB column created with `fields.JSONField()`.

When inserting data initially with `model.create`, the prepared statement passes the data correctly:
```
INSERT INTO "usermodel" (...) VALUES (...): 
...
 '{"attributes":{"cn":["William Bush"],"createTimestamp":["2020-04-30 16:20:02+00:00"],"creatorsName":["cn=admin,dc=example,dc=org"],"description":["Lt. William Bush"],
...
```
This inserts a dict with a bunch of data, as expected.

Updating the data generates incorrectly escaped SQL whether using `QuerySet.update` or `model.save`.

Queryset:
```
UPDATE "usermodel" SET "ldap"='"{\n    \"attributes\": {\n        \"cn\": [\n            \"William Bush\"\n        ],\n        \"createTimestamp\": [\n            \"2020-04-30 16:20:02+00:00\"\n        ],\n   

...
```

Model:
```
UPDATE "usermodel" SET "ldap"=$1 WHERE "ident"=$2: ['"{\\n    \\"attributes\\": {\\n        \\"cn\\": [\\n            \\"William Bush\\"\\n        ],\\n        \\"createTimestamp\\": [\\n            \\"2020-04-30 16:20:02+00:00\\"\\n        ],\\n        \\"creatorsName\\": [\\n            \\"cn=admin,dc=example,dc=org\\"\\n        ],

...
```

As you can see both update statements are wrong: instead of inserting a JSON document, they end up inserting a string. Since a string _is_ a valid JSON document, the update is successful but when attempting to read the data later it's trash.

In the Queryset case since it's not using a prepared statement I guess it has to attempt to escape the string. Is it possible to change it to use prepared statements?

In the model case since it's using a prepared statement there is no reason to quote the data; it should be passed untouched like INSERT does.

long2ice:
Yes,it's duplicated encode and will be fixed later.

grigi:
Just an update, The attempt at #378 would not resolve your issue, and as per our discussion on gitter, it's not urgent right now as if you read the value back via the ORM, it auto-decodes the string, which is technically also a bug.

So we have a bug masking a different bug.

I'll address this with the planned field refactor for 0.17.

PR #378, opened 2020-05-02
make JSONField accept valid json str value

long2ice:
## Description

make JSONField accept valid json str value
## Motivation and Context

resolve https://github.com/tortoise/tortoise-orm/issues/377
## How Has This Been Tested?

add test cases
## Checklist:


grigi:
I don't think this is going to fix the root issue in #377 
The issue is that JSONB is a native JSON support, and because JSON is any of `str, int, float, bool, None, list, dict` a `str` is a valid option. Reading it back is correct, just the update is doing it wrong.

Thanks for your attempt, but I don't feel comfortable having intentional type-morphing as a feature (where we would normally try and avoid such behaviour).

long2ice:
Is it not a good choice to update ``JSONField`` with json str value? Maybe we need it like ``CharEnumField`` or ``IntEnumField``,in db it's always ``Text``.

grigi:
For SQLite and MySQL, yes. But not for PostgreSQL. It's of type `JSONB` there. So an actual JSON blob.

I did quick look, and the issue is that passing as a reference it expects a JSON string, but a read returns a JSON blob.
When passing it as part of the SQL then it expects JSON content, but not quoted as a string.

Tortoise fixes this by automatically parsing the JSON (because it expects it to be a string) but that still populates the data with incorrect data.

grigi:
e.g. when inserted via a `.save()` the value is read back as:
```py
{'some': ['text', 3]}
```
But after an `.update()` it's:
```py
'{"some": ["text", 3]}'
```

If anything we need to be consistent. This is a PostgreSQL specific bug.

long2ice:
well,I'm not so familiar with PostgreSQL,so I close this pr?

grigi:
Ok, I'll work on this issue.

PR #383, opened 2020-05-05
docs: add test code to FastAPI Examples

dongfangtianyu:
## Description

Create an coroutine in the synchronization code ，and give it to the event loop of FastAPI app to run. 
In this way, API, DB and TestCase will use the same event loop to avoid many problems

## Motivation and Context

This may not be the best option, but it's worth documenting so others can continue to improve : )

## How Has This Been Tested?
```bash
cd examples/fastapi
pip install tortoise-orm fastapi  pytest requests asynctest
pytest tests.py
```

## Checklist:

### DIFF
--- a/docs/examples/fastapi.rst
+++ b/docs/examples/fastapi.rst
@@ -22,6 +22,9 @@ models.py
 ---------
 .. literalinclude::  ../../examples/fastapi/models.py
 
+tests.py
+-------
+.. literalinclude::  ../../examples/fastapi/_tests.py
 main.py
 -------
 .. literalinclude::  ../../examples/fastapi/main.py
--- a/examples/fastapi/_tests.py
+++ b/examples/fastapi/_tests.py
@@ -1,0 +1,40 @@
+# mypy: no-disallow-untyped-decorators
+# pylint: disable=E0611
+import asyncio
+from typing import Generator
+
+import pytest
+from fastapi.testclient import TestClient
+
+from main import app
+from models import Users
+from tortoise.contrib.test import finalizer, initializer
+
+
+@pytest.fixture(scope="module")
+def client() -> Generator:
+    initializer(["models"])
+    with TestClient(app) as c:
+        yield c
+    finalizer()
+
+
+@pytest.fixture(scope="module")
+def event_loop(client: TestClient) -> Generator:
+    yield client.task.get_loop()
+
+
+def test_create_user(client: TestClient, event_loop: asyncio.AbstractEventLoop):  # nosec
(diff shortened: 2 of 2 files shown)


examples/fastapi/tests.py
```diff
@@ -0,0 +1,35 @@
...
+    data = response.json()
+    assert data["username"] == "admin"
+    assert "id" in data
+    user_id = data["id"]
+
+    async def get_user_by_db():
+        user = await Users.get(id=user_id)
+        return user
+
+    event_loop = client.task.get_loop()
```
grigi, line 33: Ah, I see. This is how we get the event loop.

Just wish we could have the top-level async, and the event loop not being messed up, so we don't have to do boilerplate heavy code like in here.

grigi:
Thanks for getting started with this! :+1:
Could you please resolve the flake8 linter issues with this? https://travis-ci.com/github/tortoise/tortoise-orm/jobs/328165608

examples/fastapi/tests.py
dongfangtianyu: Yes, maybe we can try pytest-asyncio ？

dongfangtianyu:
checks failed again 😥
Could  you tell me why examples/fastapi/main.py not have import error ?
```python
from models import User_Pydantic, UserIn_Pydantic, Users
```

in examples/fastapi/tests.py ，look the same, but have import error  [https://travis-ci.com/github/tortoise/tortoise-orm/jobs/328474531](https://travis-ci.com/github/tortoise/tortoise-orm/jobs/328474531)
```python
from models import Users
```

examples/fastapi/tests.py
```diff
@@ -0,0 +1,39 @@
+# mypy: no-disallow-untyped-decorators
```
lntuition, line 1: Add `# pylint: disable=E0611` here can resolve lint error with CI

dongfangtianyu: yes, it is worked, thinks :)

examples/fastapi/tests.py
grigi: I did try that a while ago, and it destroyed and recreated the event loop on every test, and that caused stale network connections to databases.
And I'm unsure how it will work with the starlette test client, as it's designed to be run with the event loop paused. So it may also just break.
My experience with py.test has been that it tries to do too much, so breaks asyncio a lot. e.g. pytest 5.4 has a regression that breaks asyncio contexts, which we depend upon.

grigi:
Thank you for this, looks good to start with :+1:

Issue #387, opened 2020-05-07
Missing ids in query that bulk_create produces.


madnesspie:
**Describe the bug**
Hi! `Model.bulk_create` excludes object ids from query that it produces. This leads the ids to an ascending order, even if otherwise specified. 

**To Reproduce**
```python
# models.py
class SomeModel(models.Model):
    id = fields.IntField(pk=True)
   field = fields.IntField()

# test_bulk_create.py
import models

async def test_bulk_create_ids():
    await models.SomeModel.bulk_create(
        [
            models.SomeModel(id=100, field=1),
            models.SomeModel(id=200, field=1)
        ]
    )
    objs = await models.SomeModel.all()
    assert objs[0].id == 1
    assert objs[1].id == 2
```

**Expected behavior**
The bulk_create creates the ids according to objects that it gets.

lntuition:
@madnesspie Thank you for reporting! I'll make PR to support with specified `pk`

grigi:
Hold on. That was by design, even documented?
Bulk create is a create and forget system.

grigi:
The point of bulk_create is to ensure that the models are in the DB in the fastest way possible for a specific use case: https://tortoise-orm.readthedocs.io/en/latest/models.html?highlight=bulk#tortoise.models.Model.bulk_create

If you fetch the objects later on again they will be correct.

But db-generated PK's are not fetched.

Part of this comes from difficulty in getting results back for a db-generated pk.
PostgreSQL has syntax to return a value, so we could possibly get it working right there with minimal overhead (I don't know if `execute_many` will return any data), but getting it working right with SQLite or MySQL may require not doing an `execute_many` as we only get the rowid as a property of the cursor. Which means we will have to issue it one at a time and then read. Which won't be any faster than doing a transaction with inserts.

We might be able to get it working performant for SQLite as only a single writer is allowed (for bulk create we ensure we get a write lock) so we could possibly make the assumption that the rowid is sequential. :thinking: 

MySQL may require an exclusive table lock for that assumption to work, which would impact performance.

So, we can support this, it will likely come at a performance penalty, so I think we should only do so if explicitly instructed.

Possibly we have two different bulk create methods?

grigi:
Oh, hold on. I missed the issue here. Blame lack of sleep. Apologies.

grigi:
This should be fixed in the next release. You can test by doing a
```
pip install https://github.com/tortoise/tortoise-orm/archive/develop.zip
``

PR #388, opened 2020-05-07
Support specified pk to bulk_create

lntuition:
Support specified `pk` to `bulk_create`

fixed : #387 

* [x] Support specified `pk` to `bulk_create`
* [x] Add testcase
* [x] Update Changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Changelog
 0.16.11
 -------
 - fix: ``sqlite://:memory:`` in Windows thrown ``OSError: [WinError 123]``
+- Support specified ``pk`` to ``bulk_create()``
 
 0.16.10
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ Changelog
 0.16.11
 -------
 - fix: ``sqlite://:memory:`` in Windows thrown ``OSError: [WinError 123]``
-- Support specified ``pk`` to ``bulk_create()``
+- Support ``bulk_create()`` insertion of records with overridden primary key when the primary key is DB-generated
 
 0.16.10
 -------
--- a/tests/test_bulk.py
+++ b/tests/test_bulk.py
@@ -20,13 +20,13 @@ class TestBulk(test.TruncationTestCase):
 
     async def test_bulk_create_mix_specified(self):
         await UniqueName.bulk_create(
-            [UniqueName(id=id_) for id_ in range(2000, 3000)] + [UniqueName() for _ in range(1000)]
+            [UniqueName(id=id_) for id_ in range(10000, 11000)] + [UniqueName() for _ in range(1000)]
         )
 
         all_ = await UniqueName.all().values("id", "name")
         self.assertEqual(len(all_), 2000)
 
-        self.assertEqual(all_[:1000], [{"id": id_, "name": None} for id_ in range(2000, 3000)])
+        self.assertEqual(all_[:1000], [{"id": id_, "name": None} for id_ in range(10000, 11000)])
         inc = all_[1000]["id"]
         self.assertEqual(all_[1000:], [{"id": val + inc, "name": None} for val in range(1000)])
 
--- a/tests/test_bulk.py
+++ b/tests/test_bulk.py
@@ -13,6 +13,21 @@ class TestBulk(test.TruncationTestCase):
         inc = all_[0]["id"]
         self.assertEqual(all_, [{"id": val + inc, "name": None} for val in range(1000)])
 
+    async def test_bulk_create_with_specified(self):
+        await UniqueName.bulk_create([UniqueName(id=id_) for id_ in range(1000)])
+        all_ = await UniqueName.all().values("id", "name")
+        self.assertEqual(all_, [{"id": id_, "name": None} for id_ in range(1000)])
+
+    async def test_bulk_create_mix_specified(self):
+        await UniqueName.bulk_create(
+            [UniqueName(id=id_) for id_ in range(1000)] + [UniqueName() for _ in range(1000)]
+        )
+
+        all_ = await UniqueName.all().values("id", "name")
+        self.assertEqual(all_[:1000], [{"id": id_, "name": None} for id_ in range(1000)])
+        inc = all_[1000]["id"]
+        self.assertEqual(all_[1000:], [{"id": val + inc, "name": None} for val in range(1000)])
+
     async def test_bulk_create_uuidpk(self):
         await UUIDPkModel.bulk_create([UUIDPkModel() for _ in range(1000)])
         res = await UUIDPkModel.all().values_list("id", flat=True)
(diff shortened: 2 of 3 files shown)


lntuition:
Hmm.. I need to check why postgres failed.. Maybe it works differently with sqlite?...

grigi:
It seems that it failed here:
```
File "/home/travis/build/tortoise/tortoise-orm/tests/test_bulk.py", line 27, in test_bulk_create_mix_specified

    self.assertEqual(all_[:1000], [{"id": id_, "name": None} for id_ in range(1000)])
```
Whilst trying to generate a diff to show a nice error message. It ran out of stack space and aborted?

So it obviously didn't do the right thing. Try testing with a smaller range (e.g. 10) to debug it with?

grigi:
Also, PostgreSQL overrides the default `_prepare_insert_statement`: https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/asyncpg/executor.py#L18-L28

tortoise/backends/base/executor.py
```diff
@@ -168,14 +168,28 @@ async def execute_insert(self, instance: "Model") -> None:
...
+                values_lists_all.append(
+                    [
+                        self.column_map[field_name](getattr(instance, field_name), instance)
+                        for field_name in self.regular_columns_all
+                    ]
+                )
+
+        if values_lists:
+            await self.db.execute_many(self.insert_query, values_lists)
+        if values_lists_all:
```
grigi, line 191: I would do the `_all` variant first, as it uses pre-generated id's and will fail if the auto-generated id's generate a value that overlaps.

CHANGELOG.rst
```diff
@@ -12,6 +12,7 @@ Changelog
 0.16.11
 -------
 - fix: ``sqlite://:memory:`` in Windows thrown ``OSError: [WinError 123]``
+- Support specified ``pk`` to ``bulk_create()``
```
grigi, line 15: This is not specific enough. Try something like:

Support ``bulk_create()`` insertion of records with overridden primary key when the primary key is DB-generated

tests/test_bulk.py
```diff
@@ -13,6 +13,21 @@ async def test_bulk_create(self):
         inc = all_[0]["id"]
         self.assertEqual(all_, [{"id": val + inc, "name": None} for val in range(1000)])
 
+    async def test_bulk_create_with_specified(self):
+        await UniqueName.bulk_create([UniqueName(id=id_) for id_ in range(1000)])
```
grigi, line 17: I think I have an idea re PostgreSQL. The id you generate is from 0 to 999
I don't think 0 is a safe serial number. It should start from 1 onwards?

tests/test_bulk.py
```diff
@@ -13,6 +13,21 @@ async def test_bulk_create(self):
...
+        all_ = await UniqueName.all().values("id", "name")
+        self.assertEqual(all_, [{"id": id_, "name": None} for id_ in range(1000)])
+
+    async def test_bulk_create_mix_specified(self):
+        await UniqueName.bulk_create(
+            [UniqueName(id=id_) for id_ in range(1000)] + [UniqueName() for _ in range(1000)]
+        )
+
+        all_ = await UniqueName.all().values("id", "name")
```
grigi, line 26: Please assert that the size of all_ is 2000
Just to ensure that ID-overlap isn't happening?

tortoise/backends/base/executor.py
lntuition: That a good point! I think it will prevent `pk` overlapping.

CHANGELOG.rst
lntuition: Oh I see, really thank you for correction. :+1:

tests/test_bulk.py
lntuition: Good. I'll Change range value and will retry it

tests/test_bulk.py
lntuition: As you said, I think Changing insertion sequence will prevent ID overlapping. But we can beat the stone bridge before you cross. I'll add it soon.

lntuition:
@grigi :+1: CI works well. I think changing the insertion sequence is valid. But just one more question? Do we need to test more bigger case with `bulk_create`? Current size is just hundreds.

grigi:
The tests are to validate that things are working. So doing bulk inserts with a thousand records is OK for testing. I think?

Good work, btw!

grigi:
Oh, I see, you reduced the amount so you could see what was wrong. You're welcome to bump it back to 1000 if you so wish :-)

lntuition:
Hmm.. rollback bulk object size make wrong with postgres :(

grigi:
Wow, that's unexpected?

The `_prepare_insert_statement` override is to return the generated PK if required with an extra `RETURNS` statement.

I had a look at the error, and it's totally different now:
```
asyncpg.exceptions.UniqueViolationError: duplicate key value violates unique constraint "uniquename_pkey"

DETAIL:  Key (id)=(2001) already exists.
```

I also see that the error happens when inserting `values_lists` AFTER it already inserted `values_lists_all`
Which makes no sense, as the DB is supposed to be in charge of generating those values?

grigi:
Oh, postgres has an external seril incrementer. I think if you manually specify a PK it doesn't get updated?

lntuition:
> Oh, postgres has an external seril incrementer. I think if you manually specify a PK it doesn't get updated?

@grigi Ah.. I didn't catched it. So i'm wondering Why small object size didn't hurt external seril incrementer?... It is weird...

And Can we update external seril incrementer? and is that right solution?

grigi:
The generated sequence is probably somwhere in the 1000-2000 range, and the specified range is lower.
We do clear the table of data, but the sequence doesn't reset. So changing it to:
```
UniqueName(id=id_) for id_ in range(10000, 11000)] + [UniqueName() for _ in range(1000)]
```
or 
```
UniqueName(id=id_) for id_ in range(1, 1001)] + [UniqueName() for _ in range(1000)]
```

A proper fix would be to emit something like this to the DB:
```sql
ALTER SEQUENCE uniquename_id_seq RESTART;
```
Or we could make it auto submit that when we truncate tables for postgresql, so that test ID's are consistent?
(Which is a lot of work, so lets not do that now).

grigi:
I'm pretty certain the issue is NOT with the code, just the test/how we test.

lntuition:
@grigi Thank you for your kind explanation. I understand it. I will change the range of TC :smile:

grigi:
Oh no!
Now we have this error:
```
black --check -l 100 -t py37 tortoise/ examples/ tests/ conftest.py || (echo "Please run 'make style' to auto-fix style issues" && false)
would reformat /home/travis/build/tortoise/tortoise-orm/tests/test_bulk.py
Oh no! 💥 💔 💥
1 file would be reformatted, 178 files would be left unchanged.
Please run 'make style' to auto-fix style issues
```
:sob:

lntuition:
Oh, I was in such a hurry ! I'll fix it quickly

PR #389, opened 2020-05-07
change pypika version to equal 0.36.5

bmys:
HOTFIX: change required version of pypika
## Description

Pypika has removed basestring from from pypika.terms in new version, requirements.txt is using the newest version and that cause fail build on Heroku. 
"""
ImportError: cannot import name 'basestring' from 'pypika.terms' (/app/.heroku/python/lib/python3.8/site-packages/pypika/terms.py)
"""

## Checklist:

- [x ] My code follows the code style of this project.
- [x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pypika>=0.36.5
+pypika==0.36.5
 ciso8601>=2.1.2; sys_platform != "win32" and implementation_name == "cpython"
 iso8601>=0.1.12; sys_platform == "win32" or implementation_name != "cpython"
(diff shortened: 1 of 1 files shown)


lntuition:
We already remove basestring on 0.16.10 and also we need to follow pypika. what version of tortoise do you use? and why do we dix pypika version? i don't think so....

grigi:
Agree, we already fixed this. Please confirm by upgrading to ``tortoise-orm==0.16.10``?

Issue #385, opened 2020-05-07
Add exists method


Nenrikido:
**Describe the solution you'd like**
I'd like a `Table.exists(field=value)` method that would return `True` or `False`

**Describe alternatives you've considered**
For now the best alternative is `get_or_none` which can be useful with Python 3.8's walrus operator, but i feel like it's only a workaround.

long2ice:
I think this is useful when we only want to know whether exists in db, rather than select real record and convent to python object. What do you think? @grigi

grigi:
Sounds like a good requirement. :+1:

PR #390, opened 2020-05-08
Add  queryset.exists()

long2ice:
## Description

Add  ``queryset.exists()``,``Model.exists()`` method
## Motivation and Context

resolve https://github.com/tortoise/tortoise-orm/issues/385
## How Has This Been Tested?

add ``test_exists()``
## Checklist:


tortoise/queryset.py
```diff
@@ -536,6 +536,18 @@ def count(self) -> "CountQuery":
             offset=self._offset,
         )
 
+    def exists(self) -> "ExistsQuery":
```
grigi, line 539: Could you add `, *args: Q, **kwargs: Any`
and then call `queryset = self.filter(*args, **kwargs)` and use that queryset instead?

That should allow for a shorter form of `await Desks.exists(owned_by=jon)` whilst still allowing the longer form that you implemented to continue working.

grigi:
And could you also implement an `exists` on Model that just returns `Queryset(cls).exists(*args, **kwargs)` so we can support the short form as well?

Or, leave the exists in the queryset as you did it, but make this do .filter().exists() ?
That would be fine too.

tortoise/queryset.py
long2ice: I think do filtering in ``exists`` is not comfortable,it should like other methods like ``count()``,``first()`` etc,right?

long2ice:
> And could you also implement an `exists` on Model that just returns `Queryset(cls).exists(*args, **kwargs)` so we can support the short form as well?
> 
> Or, leave the exists in the queryset as you did it, but make this do .filter().exists() ?
> That would be fine too.

This is OK

PR #397, opened 2020-05-11
Make Field.default effect on db level when generate table, current only support MySQL

long2ice:
## Description

Make ``Field.default`` effect on db level when generate table, current only support ``MySQL``
## Motivation and Context

It's better to make default in db level.
## How Has This Been Tested?

update current ``test_early_init.py``,``test_generate_schema.py`` to make it rigit.
## Checklist:


grigi:
Hi, could you please answer a few questions:

What is the aim of this PR?
What issue are you trying to solve?

I'm asking because having multiple sources of data is a sure recipe for weird and troublesome bugs. So we either generate as much data on the DB or as little as possible.

MySQL doesn't have the syntax supported that when we create a field, it can return the data without an extra query. Once you do extra queries atomicity becomes a problem, so we run the risk of introducing subtle runtime bugs.

And then there is the concern that `ON UPDATE CURRENT_TIMESTAMP` with MySQL will overwrite the value we provide with its own.
It's fairly common to run DB's and applications on different timezones, so we may en up with different offsets. To make it worse MySQL is not very good at handling timezones, and neither is Tortoise. (to compound the problem)

I'm just very concerned about operational complexity for this?

long2ice:
Well,I just want db do more things for us.You are right that ``ON UPDATE CURRENT_TIMESTAMP`` will overwrite value we provide which will cause timezone question, sorry that I don't clear before. And I have i question that why we use ``DATETIME(6)`` instead of ``DATETIME``?

grigi:
`DATETIME(6)` is for a higher-precision datetime than regular `DATETIME` in MySQL.

I'm still jittery on the `ON UPDATE`, as the docs basically indicate it will ignore any value write. And timezone implementation on MySQL is so... unpredictable, this change still terrifies me. Even worse is that different versions/configurations will behave differently.

Hence me asking: What is this supposed to fix?

long2ice:
No,``ON UPDATE`` don't ignore input value,I tested on ``MySQL``,It only auto update when you not explicit update that column,if you specified,it will update to your give value.

grigi:
Ok. Could you please add a test that does an update query on such a field, so we can be sure that is the case?

I think doing a `.update()` should auto-generate the update field, and we need to ensure that it is exactly the same when we re-fetch the object from the DB?

long2ice:
> Ok. Could you please add a test that does an update query on such a field, so we can be sure that is the case?
> 
> I think doing a `.update()` should auto-generate the update field, and we need to ensure that it is exactly the same when we re-fetch the object from the DB?

OK, could you give a review?

grigi:
:+1:
Thank you, it looks good.

Just fix the isort issue and move the changelog entry under 0.16.12

grigi:
As always, your work is of high quality :+1:

long2ice:
Thanks!😄

PR #399, opened 2020-05-12
[WIP] Support windows env to travis

lntuition:
With #375, we found that different development environment can lead some unexpected errors. So we need to check many envs automatically. This is first step that support Windows to travis CI.

(Sadly, I'm not good at travis. Maybe we need to cross several bridges to reach destination. It can be hard work. Any advices welcome! :smile:)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ matrix:
       - choco install python --version 3.7.4
       - python -m pip install --upgrade pip setuptools wheel
       - choco install make
-    env: PATH=/c/Python37:/c/Python37/Scripts:$PATH TEST_RUNNER=green
-    script: green -vv
+    env: PATH=/c/Python37:/c/Python37/Scripts:$PATH TEST_RUNNER=nose2
+    script: "nose2 --plugin tortoise.contrib.test.nose2 --db-module tests.testmodels --db-url sqlite://:memory:"
     after_success: echo
   - python: 3.7
     env: TEST_RUNNER=green
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ matrix:
       - choco install python --version 3.7.4
       - python -m pip install --upgrade pip setuptools wheel
       - choco install make
-    env: PATH=/c/Python37:/c/Python37/Scripts:$PATH
-    script: make test
+    env: PATH=/c/Python37:/c/Python37/Scripts:$PATH TEST_RUNNER=green
+    script: green
     after_success: echo
   - python: 3.7
     env: TEST_RUNNER=green
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
       - python -m pip install --upgrade pip setuptools wheel
       - choco install make
     env: PATH=/c/Python37:/c/Python37/Scripts:$PATH TEST_RUNNER=green
-    script: green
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,6 +7,16 @@ env:
...
   include:
+  - name: Python 3.7 / Windows / Partial
+    os: windows
+    language: shell
+    before_install:
+      - choco install python --version 3.7.4
+      - python -m pip install --upgrade pip
+      - choco install make
+    env: PATH=/c/Python37:/c/Python37/Scripts:$PATH
+    script: make testall
```
grigi, line 18: This will try and test all the db variants? I suspect we may not have them all available right now.
It's fine to start with just `make test` to only test SQLite.

grigi:
Possibly we should install less, we don't need the linters, etc... ?
Then a simpler `pip install asynctest pydantic "pytest<5.4" green -e .` could also do fine?
And then we just run `py.test` or `green`?

.travis.yml
lntuition: Yeah. I'll try with the smaller one.

grigi:
Oh dear, it froze up somewhere :'(

`green -vv` would at least tell us exactly WHICH test caused a hang?

lntuition:
OMG, Output is broken..

grigi:
I think that terminal is still unreadable :-( 
Possibly Windows and coloured terminal = trouble ?
Lets try disabling any fancy terminal stuff and concurrency whilst we are at it.

`-s1` → Single process.
`-T` → Force terminal colors off.
`green -s1 -T -vv` ?

grigi:
or possibly nose2 works better on windows?
`nose2 --plugin tortoise.contrib.test.nose2 --db-module tests.testmodels --db-url sqlite://:memory: -v` ?

grigi:
Oops, you forgot the `-v` so we can see the failed progress, or where it is getting stuck.
Because it looks like it is getting stuck again :-( →Line 672

Does a newer Python run better on Windows?
I have no real idea where to start as I don't have a Windows system, else I would try and run it locally and see what is going on.

lntuition:
Oh. I see.. I also don't know why failed. I'll try different ways (need time) and also try to setup windows env to my local computer :smile

lntuition:
Find point! `test_run_async_raised` stucked. We need to check why it only stuck on Windows.

grigi:
I see the `test_update_from_dict (tests.test_model_methods.TestModelMethods) ... FAIL` fail as well. 

We could add a skip to `test_run_async_raised` to just skip it for now and continue testing? Possibly add a conditional windows `@skipif(os.name == 'nt')` skip?

grigi:
Looks like you're playing whack-a-mole :grin:

lntuition:
Hmm.. I don't now why `test_run_async_raised` not passed 😢 ..

p.s) I'm really sorry because i'm struggled with my personal work so i do not have time for working about this and etc... (Unfortunately, I think it will not be easily resolved..)

grigi:
No worries. We are all doing this in our extra time.

Whack-a-mole is a game where you whack one thing down, and then another pops up, and you whack it down.
Typically not a good thing re software.

grigi:
Thank you for starting with this :grinning: . I got it working in #467 and will merge that and close this.

PR #400, opened 2020-05-13
make default effect in db level worked for sqlite and postgres

long2ice:
## Description

make default effect in db level worked for sqlite and postgres.
## Motivation and Context

It's better to make default in db level.
## How Has This Been Tested?

update test_generate_schema.py to make it work.
## Checklist:


tests/schema/test_generate_schema.py
```diff
@@ -186,13 +186,13 @@ async def test_schema(self):
 CREATE TABLE "tournament" (
     "tid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
     "name" VARCHAR(100) NOT NULL  /* Tournament name */,
-    "created" TIMESTAMP NOT NULL  /* Created *\\/'`\\/* datetime */
+    "created" TIMESTAMP NOT NULL  /* Created *\\/'`\\/* datetime */ DEFAULT CURRENT_TIMESTAMP
```
grigi, line 189: Could we move the default to before the comment? It currently reads rather badly?

tortoise/backends/sqlite/schema_generator.py
```diff
@@ -19,3 +21,21 @@ def _table_comment_generator(self, table: str, comment: str) -> str:
...
+    def _column_default_generator(
+        self,
+        table: str,
+        column: str,
+        default: Any,
+        auto_now_add: bool = False,
+        auto_now: bool = False,
+    ) -> str:
+        default_str = " DEFAULT"
+        if not (auto_now or auto_now_add):
```
grigi, line 34: I think `auto_now` implies `auto_now_add`?

grigi: Yep: https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/fields/data.py#L295
You can simplify this?

tests/schema/test_generate_schema.py
long2ice: The ``FIELD_TEMPLATE = '"{name}" {type} {nullable} {unique}{primary}{comment}{default}``, so...

tortoise/backends/sqlite/schema_generator.py
long2ice: Yeah, that's right.

tests/schema/test_generate_schema.py
grigi: You can override it on a per-dialect level. add a `FIELD_TEMPLATE` in `SqliteSchemaGenerator` and it will use that instead.

grigi: Also, the comment is always a blank string for Postgresql as it emits it in a later instruction. And it is overridden for MySQL. So you can happily change the default?

grigi: Hold on, did you not add that in your previous pr?

long2ice: What? You mean move all dialect  ``default`` previous ``comment``?

grigi: I meant you added it in your previous PR about this:
https://github.com/tortoise/tortoise-orm/commit/72f84f0848dc68041157f03e60cd1c92b0ee5137#diff-738f1a7c384904401ec302972c0c4c21R21

So you can change the one in `base/schema_generator.py` and it will fix SQLite, and have no change for PostgreSQL (since PostgreSQL does comments outside of the `create table` statement)

PR #406, opened 2020-05-15
[WIP] Apps refactorting. Migrations ground work in progress 

abondar:
So this is what I've been working on lately, but it is far from being finished, and considering https://github.com/long2ice/aerich we have to make sure if it is really needed to be finished.

So what are key features here? 
First of all it is apps refactoring. With this, Tortoise becomes a little bit less globalstated, and apps takes some functionality from `Tortoise` class. It is needed to make apps not just container, but something more that can handle logic of registering and deregistering models.

All of that is needed to make it possible to have several apps at the moment, which is required for migrations, when you want to write your data migrations using code of historical models that can be constructed in memory from previous migrations.

So, regarding that PR, what's done and what is there to be done
- [X] Apps refactoring
- [X] State of app module, that holds info about all models at given point of "time"
- [X] Operations base, that allows to regenerate model in it's historical state
- [X] State editor base, that should be called by `Operations` to translate operation to SQL
- [ ] Autodetect changes based on two state classes
- [ ] Write code in operations that calls state editor
- [ ] Write migrations class that runs operations
- [ ] Code generate migration files
- [ ] CLI to work with all of that
- [ ] Operation models for indexex (right now it is done mostly for fields and models)
- [ ] Concrete implementations for state editors of different db backends
- [ ] Do a major clean up
- [ ] Bring branch up to date with master

On how it should work in general:
Each migration consists of list of operations. There is virtual app state that describes current state of the app (all model, fields and etc)

Running migration forward calls each operation in order. Before each operation we clone current state and then operation mutates it's to match state that is desired. Also, each operation execute set of SQL actions, predescribed for given operation. 

On the other hand, we can get current state of the app from `Tortoise.app` and use autodetector to generate operations based on diff between them, and create migration with given operations.

Using state mechanism we can write data migrations using models regenerated for given point in migrations graph


grigi:
Personally I think the apps refactor alone is worth it. I'll review in detail when I get a chance.

I originally envisioned building models off the 'model.describe()' spec, so migrations could be be done with a lot less infrastructure.

sunshineinwater:
It looks like will take a long time......

Issue #95, opened 2019-01-30
Allow to retrieve the raw SQL of a queryset


florimondmanca:
**Is your feature request related to a problem? Please describe.**
There's at the moment no way to inspect the actual SQL query that a QuerySet would execute when evaluated. This can however be very useful for debugging.

**Describe the solution you'd like**
Add an `.sql` property to `QuerySet` that could be used like this:

```python
>>> print(Book.all().sql)
SELECT * FROM "books";
```

Implementation-wise, I believe `return self._make_query().get_sql()` should work.

**Describe alternatives you've considered**
We could have named this `.query`, but it would be unclear whether this returns a plain string (as this issue suggests) or the actual `Query` object from pypika. I don't think the second option is a good idea, because using pypika should an implementation detail to our users.

**Additional context**
Follow-up on @abondar 's suggestion in #91

Priestch:
@florimondmanca This feature was added in version 0.16.12, you can give it a try!

PR #410, opened 2020-05-19
(WIP) Add .sql() (#95)

Priestch:
## Description
Add .sql() method for instance of QuerySet to get the actual SQL be executed.

## Motivation and Context
Sometimes we just want to make sure that the SQL generated by the ORM is correct, or for debugging usage as described in #95 

## How Has This Been Tested?
I'm not sure how to add related test code, maybe current test is enough?

## Checklist:

### DIFF
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -317,4 +317,4 @@ class TestQueryset(test.TestCase):
 
     async def test_get_raw_sql(self):
         sql = IntFields.all().sql()
-        self.assertRegex(sql, r'^SELECT.+FROM.+')
+        self.assertRegex(sql, r"^SELECT.+FROM.+")
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -317,4 +317,4 @@ class TestQueryset(test.TestCase):
 
     async def test_get_raw_sql(self):
         sql = IntFields.all().sql()
-        self.assertGreater(len(sql), 10)
+        self.assertRegex(sql, r'^SELECT.+FROM.+')
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -314,3 +314,7 @@ class TestQueryset(test.TestCase):
     async def test_order_by_bad_value(self):
         with self.assertRaisesRegex(FieldError, "Unknown field badid for model IntFields"):
             await IntFields.all().order_by("badid").values_list()
+
+    async def test_get_raw_sql(self):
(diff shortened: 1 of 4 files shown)


grigi:
Hi, thanks for this.

Could you please add yourself to `CONTRIBUTORS.rst` and a short description of the change in `CHANGELOG.rst`.

Also, could you please do a test for this?
A simple `sql = Foo.all().sql()` would do.

Priestch:
Add a simple test and update the `CHANGELOG.rst`.

tests/test_queryset.py
```diff
@@ -314,3 +314,7 @@ async def test_all_values(self):
     async def test_order_by_bad_value(self):
         with self.assertRaisesRegex(FieldError, "Unknown field badid for model IntFields"):
             await IntFields.all().order_by("badid").values_list()
+
+    async def test_get_raw_sql(self):
+        sql = IntFields.all().sql()
+        self.assertGreater(len(sql), 10)
```
lntuition, line 320: This testcase don't looks nice. It is unable to check that if a valid query has been created. We should compare that REAL concrete query was generated. Also need to check more query (e.g : with join, using function, etc ...)

Priestch: > Hi, thanks for this.
> 
> Could you please add yourself to `CONTRIBUTORS.rst` and a short description of the change in `CHANGELOG.rst`.
> 
> Also, could you please do a test for this?
> A simple `sql = Foo.all().sql()` would do.

I'm not sure about how to add test code for .sql(), so I ask for help, maybe just make a detail assert is enough?
```python
assert sql == 'SELECT .. FROM'
```

I think maybe we don't need to check more query, because we already test the function in other tests of queryset.

lntuition: As you said, I think `assertEqual(sql, 'SELECT ... FROM')` is better. Also I want to check that new method is strong in many cases. I think we need more opinion. @grigi What do you think about this one?

grigi: We definitely want to test that it's a query.
A simple = would be ideal, with the caveat of different syntaxes.

Note that the SQL generated is going to be different between different dialects, so you may just test that it starts with `SELECT ` and has ` FROM ` etc...

Priestch: @grigi @lntuition I update the test, if we need more tests, I will do.

By the way, thank all your work for the awesome ORM.

grigi:
:+1: Thank you for this!

PR #411, opened 2020-05-20
Fix postgres BooleanField default value convent

long2ice:
## Description

Fix postgres BooleanField default value convent.
## Checklist:


grigi:
Could you add a test case that tests for this edge case? I just want to ensure that we test against regressions happening later on. (They sadly often do :sob:)

long2ice:
Yes, lazy is devil, more tests is good.

grigi:
Think of it this way → More tests buys you future laziness :grin:

PR #415, opened 2020-05-22
Fix JSONField typed in pydantic_model_creator

long2ice:
## Description

Fix ``JSONField`` typed in ``pydantic_model_creator``
## Motivation and Context

Before, it is wrong convert with ``JSONField`` in ``pydantic_model_creator``, such as ``[{"a": 1, "b": 2}]`` will convert to ``{'a': 'b'}``
## How Has This Been Tested?

add ``test_json_field`` in ``test_pydantic.py`` to test it.
## Checklist:


tortoise/contrib/pydantic/creator.py
```diff
@@ -344,7 +344,9 @@ def get_submodel(_model: "Type[Model]") -> Optional[Type[PydanticModel]]:
             comment = _cleandoc(func)
             if annotation is not None:
                 pannotations[fname] = annotation
-
+        # Json fields
+        elif field_type is fields.JSONField:
+            pannotations[fname] = Any  # type: ignore
```
grigi, line 349: :thinking: Yes, this sort-of makes sense ito JSON, as it can only support what it can support.

grigi:
:+1: Thank you for this fix. I'll do a release soon.

PR #433, opened 2020-06-10
Support config of loop and connection class to asyncpg

blakewatters:
Allows configuration of the asyncpg connection class and asyncio event loop used by `AsyncpgDBClient` by passing them through the `config` option.

## Motivation and Context

Needed to enable SQL logging of the models.

## How Has This Been Tested?

My test suite passes and the app runs but I did not spin up on the Tortoise tests because these changes are pretty surgical and I am short on time.

## Checklist:

### DIFF
--- a/tortoise/backends/asyncpg/client.py
+++ b/tortoise/backends/asyncpg/client.py
@@ -50,6 +50,8 @@ class AsyncpgDBClient(BaseDBAsyncClient):
     executor_class = AsyncpgExecutor
     schema_generator = AsyncpgSchemaGenerator
     capabilities = Capabilities("postgres")
+    connection_class = asyncpg.connection.Connection
+    loop = None
 
     def __init__(
         self, user: str, password: str, database: str, host: str, port: SupportsInt, **kwargs: Any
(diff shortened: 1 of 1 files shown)


grigi:
Hi Blake, thank you for your contribution.

It seems the only test fail is `TestConnectionParams.test_postres_connection_params` and that is just the two extra parameters you have added.
Adding the default two parameters here: https://github.com/tortoise/tortoise-orm/blob/develop/tests/backends/test_connection_params.py#L63-L74 should fix it.

Also, could you please add yourself to the `CONTRIBUTORS.rst` file and a simple one-liner describing the change to the 0.16.14 (or top) of `CHANGELOG.rst`.

grigi:
Hi, I manually merged your work + test fixes and updates to changelog/contributors.

iburadempa:
I don't see the point of this commit. AsyncpgDBClient.init() accepts arbitrary kwargs anyways, they get copied to self.extra and **self.extra gets passed on to asyncpg.create_pool(). Why treat the kwargs 'pool' and 'connection_class' differently than, say, 'setup' or 'init'?

The documentation could however be improved by telling users that they can pass kwargs to create_pool() by using the 'config' alternative for Tortoise.init() and adding the kwargs to config['connections'][conn_name]['credentials']. Admittedly, 'credentials' could be renamed or merged with a to-be-created config['connections'][conn_name]['client_init'].

PR #437, opened 2020-06-19
passing using_db in get_or_create

jiyeonseo:
`using_db` has not been passed when it tried to `create` in `get_or_create` 

## Description
It causes the issue that creates a record in a different DB
I think the argument has been removed in [this commit](https://github.com/tortoise/tortoise-orm/commit/a33d90d783d83dfbe1811c08d0885c994e109430) 

## How Has This Been Tested?
I test this on my code which is using the secondary DB as a "default"(read-only) and the primary DB as "master"(read and write). 
```py

async with in_transaction("master") as connection:
   service_emp = await ServicesEmployees.get_or_create(service_id=service_id, emp_no=emp_no, using_db=connection)

```
Errors occurred in above code. I fixed it just like I did in this PR, and it works well.  

## Checklist:

### DIFF
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -951,7 +951,7 @@ class Model(metaclass=ModelMeta):
             if instance:
                 return instance, False
             try:
-                return await cls.create(**defaults, **kwargs), True
+                return await cls.create(**defaults, **kwargs, using_db=using_db), True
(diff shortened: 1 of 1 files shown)


grigi:
Manually merged with a test case and changelog/contributors updates.

Issue #18, opened 2018-07-02
Generic JSONField for all db drivers


Zeliboba5:
Currently there is only implementation for postgres using `jsonb`. It will be good thing to implement generic field that could be used for basic storing of JSON

grigi:
I would love a feature emulation framework, where a JSONField by default would be emulated using a TextField, but if using PostgreSQL it can be "upgraded" to a faster more searchable variant?

We are already doing it for, e.g. date/time fields and SQLite. it is actually a varchar in the DB, and when you read you get a string, not a date/time object back.

Zeliboba5:
I think it shouldn't be a problem to patch real field we are using in the `Tortoise.init()` - it should be just a matter of making few dicts with mapping of `ClientClass => Field` and storing this dicts inside the field itself. I think it will be quite extendable for future uses and may be even for custom fields.

I will look into it when will get to this issue

grigi:
I have started porting this over, and the PG schema_generator injects the type difference.
That should get it "working", but not all the extra features, like partial searching.

In concert with this I'm planning to do more testing on the model fields themselves.

grigi:
Close this please? Or give me permissions to close it?

PR #451, opened 2020-07-04
Let JSONField use JSON field type with MySQL

nikolaik:
## Description
To allow querying nested data inside a JSONField value with MySQL, the field type for `JSONField` is changed to `JSON`. This is available starting with MySQL version 5.7. The MariaDB and MySQL implementation of the JSON type [differs a bit](https://mariadb.com/kb/en/json-data-type/#differences-between-mysql-json-strings-and-mariadb-json-strings).

## Motivation and Context

Instead of opening an issue I created a minimal PR to get some feedback on how to implement this. 

Should the docs say something about the minimum version of MySQL required?

Somewhat related to work in #18 

## How Has This Been Tested?

Updated existing test

## Checklist:


grigi:
Thank you for your PR.
Hi, it seems that this is only supported on MySQL 5.7+ and MariaDB 10.2+
We still have MySQL 5.5 instances at work, so I'm uncertain if we can just bump the type to JSON universally.

This gives me pause, as we don't have a nice clean way of supporting alternate implementation if not supported.
e.g. SQLite 3.9+ supports JSON but only if built with the right extension. So "minimal" builds don't support it.

Do you have any proposal on this?

SakuraSound:
I just ran into this today while evaluating tortoise for a project. It would be good if we could pass a parameter to signify the overriding type. 

```python
. . .

class User(Model):
    user_id = fields.UUIDFIeld(pk=True)
    preferences = fields.JSONField(native=True)

. . .
```

droserasprout:
Another approach is to provide a separate class with only `_db_mysql` overwritten in some place like `tortoise.fields.compat` module. Or maybe specify target MySQL version in connections config dictionary and adjust field types according to it? So there's a lot of options. @grigi, what do you think about? It would be nice to have full native JSON field support in tortoise.

PR #462, opened 2020-07-21
add parse datetime from timestamp

TrDex:
Added parsing a `datetime` from timestamp.

## Description
-

## Motivation and Context
`DatetimeField.to_python_value()` method accepts Any value. If the value type is `datetime` it's returned unchanged, otherwise the value is passes to `iso8601.iso8601.parse_date(datestring=value)` method, which requires `datestring` to be an instance of `_basestring` (a.k.a. `str` for Python 3+). I added a check if the value is an instance of int, if it is, it's parsed as timestamp.

## How Has This Been Tested?
- 3 testes failed on my machine. After the change 3 tests still fail with the same errors.

## Checklist:
- [?] My code follows the code style of this project.
- [?] My change requires a change to the documentation.
- [-] I have updated the documentation accordingly.
- [+] I have read the **CONTRIBUTING** document.
- [-] I have added tests to cover my changes.
- [-] All new and existing tests passed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Changelog
 0.16.14
 -------
 - Make ``F`` expression work with ``QuerySet.filter()``.
+- Added ``datetime`` parsing from ``int`` for ``fields.DatetimeField``.
 
 0.16.13
 -------
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -31,6 +31,7 @@ Contributors
 * Lev Gorodetskiy ``@droserasprout``
 * Hao Gong  ``@dongfangtianyu``
 * Peng Gao ``@Priestch``
+* Mykola Solodukha ``@TrDex``
 
 Special Thanks
 ==============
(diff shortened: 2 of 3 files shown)


TrDex:
Done!

By the way, I noticed an error in `CHANGELOG.rst`. Here is a screenshot
image

I don't know how to fix this, because I'm new to this type of files, but I thought I should aware you of that.

grigi:
restructured text is like markdown, but supports "directives" and those are customizable.
That directive is to make the section collapsible in the generated docs. Other directives could generate docs off an external source, like code or whatever.

So that's likely a directive that wasn't setup in your editor.

grigi:
No sorry, that directive is to render subsections with a different style. Specific to the cloud sphinx theme.

TrDex:
Thanks for explanation!


> restructured text is like markdown, but supports "directives" and those are customizable.
> That directive is to make the section collapsible in the generated docs. Other directives could generate docs off an external source, like code or whatever.
> 
> So that's likely a directive that wasn't setup in your editor.

PR #480, opened 2020-08-07
Check default for not None in pydantic model creation, propagate default to pydantic

Yolley:
## Description
I propose two changes to `pydantic_model_creator`:
1. Check `default` of a field with `is not None`
2. Propagate default values to a pydantic model

## Motivation and Context
1. As of now if a model has a BooleanField with `default=False` or Integer field with `0` in default, `pydantic_model_creator` won't make these fields `Optional`, thus leading to wrong pydantic schema with these fields marked as required.
2. As of now default values are not propagated to a pydantic model, therefore when we generate a schema and add it to Swagger, fox example, it won't be clear to a user, what is the default value of an Optional field. Also I added callable default to `default_factory` just for consistency (to propagate all default fields), but I guess that it is more controversial, because pydantic will always try to generate field value with factory and if we create a model with `exclude_unset=True` param, we would call default function again, but now on Tortoise side, so it may be changed (may be we could add some boolean flag  to `pydantic_model_creator` arguments that either allow or not allow default values propagation)

## How Has This Been Tested?
The existing tests actually don't have models with Boolean or Integer field with `0` in default value. I could add, but I guess it would to require to change all tests that use these models, so I am not sure if it needs to be done, because well my changes certainly don't break the existing logic.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,11 @@ Changelog
 
 0.16
 ====
+0.16.15
+-------
+- check ``default`` for not ``None`` on pydantic model creation
+- propagate default to pydantic model
+
 0.16.14
 -------
 - Make ``F`` expression work with ``QuerySet.filter()``.
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -38,6 +38,7 @@ Contributors
 * Rodrigo Oliveira ``@allrod5``
 * ``@SnkSynthesis``
 * Alex Sinichkin ``@AlwxSin``
+* ``@Yolley``
 
 Special Thanks
 ==============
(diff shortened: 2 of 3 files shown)


grigi (approved):
This looks good :+1:
Could you please add a one-liner description of this fix in CHANGELOG.rst & add yourself to CONTRIBUTORS.rst

Yolley:
> This looks good 👍
> Could you please add a one-liner description of this fix in CHANGELOG.rst & add yourself to CONTRIBUTORS.rst

I added, but it seems that tests are failing now, and I don't think that this happens cause of my changes.

long2ice:
May should update that test case?
image

grigi:
I don't think that test case has anything to do with the code-changes. 
I restarted the test-run, and now it passes. 
I suppose that test isn't entirely deterministic. it seems to occasionally change order. Will have to investigate.

long2ice:
Yes,it maybe random order

PR #482, opened 2020-08-14
replace List with Iterable for some typings

AlwxSin:
## Description
Change some `List` typing with `Iterable` or `Sequence`

## Motivation and Context
I have a tuple or set with raw model objects and can't just pass them to bulk_create, because mypy will throw an error.
`.bulk_create` explicitly receive only a list of models even though it only iterates over them.
```python
set_of_raw_models = {MyModel(...) for smth in other_smth}
MyModel.bulk_create(set_of_raw_models)  #  <--- mypy error
```

Somewhere I read "A rule of thumb. Arguments should be as general as possible, return values should be as specific as possible".

## Checklist:


grigi (approved):
:+1:

PR #511, opened 2020-09-25
Auto resolve Enum values when creating Pydantic models

lqmanh:
## Description
Auto resolve `Enum` values when creating Pydantic models via `pydantic_model_creator`. 

## Motivation and Context
Current behavior:
```py
from enum import Enum
from tortoise.fields import CharEnumField

class Bar(Enum):
    X = "x"
    Y = "y"

class Foo(Model):
    bar = CharEnumField(Bar)

FooPydantic = pydantic_model_creator(Foo)
one = Foo(bar="x")  # OK
another = Foo(bar=Bar.X)  # Error: str type expected
```

New behavior:
```py
one = Foo(bar="x")  # OK
another = Foo(bar=Bar.X)  # OK too
```

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ Changelog
 ====
 0.16.17
 -------
-- Auto resolve `Enum` values.
+- Auto resolve `Enum` values when creating Pydantic models via `pydantic_model_creator`.
 
 0.16.16
 -------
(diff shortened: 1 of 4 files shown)


lqmanh:
It seems that there was something wrong with `make check`.
I'm not sure. Maybe it's because `black --check -l 100 -t py37 tortoise/ examples/ tests/ conftest.py` failed. However, previous builds succeeded without any problems.

Issue #532, opened 2020-10-23
AttributeError: 'Loop' object has no attribute '_selector' on tortoise.contrib.test


cuducos:
**Bug description**

I am using Python 3.9.0 and in Pytest's `conftest.py` I use the `tortoise.contrib.test.initializer` as suggested [in the docs](https://tortoise-orm.readthedocs.io/en/latest/contrib/unittest.html#py-test), but I get an `AttributeError` because it looks like the loop might not have the attribute `_selector`:

<details>

```
______________________________________________________________________________________ ERROR at setup of test_get_returns_405 _______________________________________________________________________________________

    @fixture(scope="session", autouse=True)
    def db():
>       initializer(["api.models"], db_url=POSTGRES_TESTS_URI, app_label="api")

tests/conftest.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

modules = ['api.models'], db_url = 'sqlite://:memory:', app_label = 'api', loop = <uvloop.Loop running=False closed=False debug=False>

    def initializer(
        modules: List[str],
        db_url: Optional[str] = None,
        app_label: str = "models",
        loop: Optional[AbstractEventLoop] = None,
    ) -> None:
        """
        Sets up the DB for testing. Must be called as part of test environment setup.
    
        :param modules: List of modules to look for models in.
        :param db_url: The db_url, defaults to ``sqlite://:memory``.
        :param app_label: The name of the APP to initialise the modules in, defaults to "models"
        :param loop: Optional event loop.
        """
        # pylint: disable=W0603
        global _CONFIG
        global _CONNECTIONS
        global _SELECTOR
        global _LOOP
        global _TORTOISE_TEST_DB
        global _MODULES
        global _CONN_MAP
        _MODULES = modules
        if db_url is not None:  # pragma: nobranch
            _TORTOISE_TEST_DB = db_url
        _CONFIG = getDBConfig(app_label=app_label, modules=_MODULES)
    
        loop = loop or asyncio.get_event_loop()
[...]

PR #533, opened 2020-10-23
Makes loop's _selector attribute optional

cuducos:
## Description

It looks like that we cannot take for granted that aloop will have a `_selector` attribute, so this PR is a hotfixz to `tortoise.contrib.test` to threat this attribte as optional.

## Motivation and Context

As described in #532, a loop without a `_selector` attribute break the test initializer function. This PR is a hotfix for situations like that.

## How Has This Been Tested?

I ran the context described in that issue with the code from this PR and it works. Also, I ran the test suite.

## Checklist:

- [ ] My change requires a change to the documentation. **(no, it just changes the internals, not the API)**
- [ ] I have added tests to cover my changes. **(no, as there was no test for the `initializer` I thought it might be a design decision not to test testing tools)**

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,7 +8,7 @@ about: Create a report to help us improve
 A clear and concise description of what the bug is.
 
 **To Reproduce**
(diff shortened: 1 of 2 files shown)


abondar:
Not actual after https://github.com/tortoise/tortoise-orm/pull/1636

PR #544, opened 2020-11-06
Timezone support added.

cijo7:
Now a DatetimeField can be created with timezone support.

Eg:

 
    datetime_tz = fields.DatetimeField(tz=timezone(timedelta(hours=7, minutes=30)), db_tz=timezone(timedelta(hours=0)))

## Description
After setting both ``tz`` and ``db_tz`` you can pass datetime aware objects in any timezone and
    it will get automatically converted to ``db_tz`` and stored in database and attribute will be
    accessible with timezone ``tz``. If you pass native datetime objects, the timezone of datetime
    is assumed to be in ``db_tz``.

## How Has This Been Tested?
Tested with mysql, sqlite. Please test in postgresql

### DIFF
--- a/tests/fields/test_time.py
+++ b/tests/fields/test_time.py
@@ -1,4 +1,4 @@
-from datetime import date, datetime, timedelta
+from datetime import date, datetime, timedelta, timezone
 from time import sleep
 
 from iso8601 import ParseError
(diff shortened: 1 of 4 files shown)


long2ice:
I closeed this since we need a complete solution, see https://github.com/tortoise/tortoise-orm/pull/557 and feel free to give suggestion.

PR #551, opened 2020-11-13
Make Tortoise's __init__ docs more clear

jakub-balinski:
Renames example database name and key name in example config dict. Starts from [here](https://github.com/jakub-balinski/tortoise-orm/blob/develop/tortoise/__init__.py#L498).

## Description
I find it confusing to set a key name to "models" in the example config dictionary, when an entry name in it is required to be the same, so I renamed it to "my_app". Also I changed database name in example DB_URL string to "test" instead of "events", which I find more consistent with the example above it (dict format).

## Motivation and Context
Make docs more clear to new users.

## Checklist:

### DIFF
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -495,10 +495,10 @@ class Tortoise:
                                 }
                             },
                             # Using a DB_URL string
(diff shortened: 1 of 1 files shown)


long2ice:
Thanks for that!

PR #561, opened 2020-11-18
Allow passing modules to Tortoise.init_models()

lqmanh:
## Description

Allow passing module objects to `models_paths` param of `Tortoise.init_models()`.

## Motivation and Context

Now `Tortoise.init_models()` only accepts `models_paths` as a list of strings, which may cause `ConfigurationError(f'Module "{models_path}" not found')` later.
With this PR, passing module objects directly instead of their paths can prevent those potential errors. Most editors/IDEs will alert developers immediatelly if there is something wrong.

**Example:**

Instead of:

```py
from tortoise import Tortoise

async def main():
    await Tortoise.init(
        db_url='sqlite://db.sqlite3',
        models={'models': ['app.models']}
    )

...
```

Do this:

```py
from tortoise import Tortoise
import app.models

async def main():
    await Tortoise.init(
        db_url='sqlite://db.sqlite3',
        models={'models': [app.models]}
    )

...
```

## Checklist:
- [ ] ~My change requires a change to the documentation.~
- [ ] ~I have updated the documentation accordingly.~

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Changelog
 0.16.19
 -------
 - Replace set `TZ` environment variable to `TIMEZONE` to avoid affecting global timezone.
+- Allow passing module objects to `models_paths` param of `Tortoise.init_models()`. (#561)
 
 0.16.18
 -------
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -43,6 +43,7 @@ Contributors
 * Bogdan Evstratenko ``@evstratbg``
 * Mike Ryan ``@devsetgo``
 * Eugene Dubovskoy ``@drjackild``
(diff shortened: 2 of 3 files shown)


abondar (approved):
Thanks you! Looks good to me

Could you please add description to the top of changelog file?

lqmanh:
@abondar I've updated the changelog and added myself to contributor list.

abondar:
@lqmanh released as 0.16.19

bnaoki:
Hello, should the types have been updated for `init()` and `generate_config()` as well?

lqmanh:
@bnaoki No I hadn't updated those functions. I'll create another PR when I have time.

lqmanh:
Hey @bnaoki,
Seems like @long2ice has already helped. Many thanks!

PR #599, opened 2020-12-29
Add a default value for `Tortoise.get_connection` connection_name argument

mkniewallner:
## Description
This PR intends to set `default` as a default value for `connection_name` argument in Tortoise `get_connection` method.

## Motivation and Context
Passing `db_url` to Tortoise `init` method  currently automatically instantiates the connection as `default`, and most projects may only ever define a `default` connection when using a dict-based configuration through `config` or `config_file` arguments.

The proposed change would remove the need to explictly pass `default` to the method, which would be useful for projects that only use a single default connection.

## How Has This Been Tested?
I've updated the tests so that they don't specify the parameter anymore, but I've added a specific test to showcase that explicitly passing `default` value still works as expected. I also took the occasion to add a test showcasing a `KeyError` when a connection doesn't exist.

## Checklist:

### DIFF
--- a/examples/manual_sql.py
+++ b/examples/manual_sql.py
@@ -16,8 +16,8 @@ async def run():
     await Tortoise.init(db_url="sqlite://:memory:", modules={"models": ["__main__"]})
     await Tortoise.generate_schemas()
 
-    # Need to get a connection. Unless explicitly specified, the name should be 'default'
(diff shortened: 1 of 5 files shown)


lqmanh:
If this PR ever gets merged, you may want to add that default value to `in_transaction` context manager and `atomic` decorator as well.

PR #623, opened 2021-01-25
Validate JSON after decode

### LINKED ISSUE #621: JSONField calls validate before decode
**Describe the bug**

Tortoise v0.16.20 included custom validations. However, if you use one of these on a `JSONField` then when retrieving data from the database then the validation is run against the encoded string version of the JSON rather than the decoded value. This means the validation function for JSON isn't symmetric as it runs against the decoded value when setting it.

**To Reproduce**

```python
import uuid

from tortoise import fields
from tortoise.exceptions import ValidationError

def must_have_three_items(value: List[str]) -> None:
    if not len(value) == 3:
        raise ValidationError("List must have three items.")

class TestModel(Model):
    id = fields.UUIDField(pk=True, default=uuid.uuid4)
    data = fields.JSONField(validators=[must_have_three_items])

async def repro() -> None:
    m1 = await TestModel.create(data=["foo", "bar", "baz"]
    m2 = await TestModel.get(pk=m1.id)  # boom! value = '["foo", "bar", "baz"]'
```

**Expected behavior**

It ought to be possible to write the validation for JSON fields as a symmetric function, i.e. if it validates the Python value on the way into the database then it ought to validate the Python value on the way out, rather than the literal JSON string.

**Additional context**

I was using Postgres so this was a JSONB column, but I don't think this should matter.

The fix looks to be quite straightforward, moving [this line](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/fields/data.py#L470) after the conditional decode. I'm happy to put a PR together if you confirm the issue is valid.

gregbeech:
Validates JSON retrieved from the database after decode rather than before.

## Description

Changes the order of operations so JSON is decoded before validation is run, along with a test for each of the three supported input types for the field.

## Motivation and Context

Custom JSON validations are currently run before the value stored in the database is converted to the Python type. This creates an asymmetry in the validation function as it runs against the Python type on create/update, but against the raw JSON string on fetch.

Closes #621

## How Has This Been Tested?

All unit tests run.

## Checklist:


tests/testmodels.py
```diff
@@ -257,10 +258,16 @@ class JSONFields(Model):
...
+    @staticmethod
+    def dict_or_list(value: Union[dict, list]):
+        if not isinstance(value, (dict, list)):
+            raise ValidationError("Value must be a dict or list.")
+
     id = fields.IntField(pk=True)
     data = fields.JSONField()
     data_null = fields.JSONField(null=True)
     data_default = fields.JSONField(default={"a": 1})
+    data_validate = fields.JSONField(null=True, validators=[lambda v: JSONFields.dict_or_list(v)])
```
gregbeech, line 270: Putting the field on this model means the 'describe' tests need to be updated, but it feels more appropriate than putting it on the validation model as this is testing the `JSONField` itself, not packaged validations. I've used a static method on the object rather than a module method as the validator to keep things self-contained, but happy to change if you find it too clumsy.

long2ice:
Looks good, could you update changelog and version? After CI pass, I will merge it.

long2ice (commented):
Good!

long2ice (approved):

PR #630, opened 2021-01-31
fix save with F expression and field with source_field

vaal-:
Hi
I'm not 100% sure, but there seems to be a bug in the case of using F expression for a field with source_field. 
If I update the record via update call - no problem.
But if I update via save call - TypeError: not enough arguments for format string.


long2ice:
Could you add unittest to confirm that?

vaal-:
> Could you add unittest to confirm that?

Added a couple of tests

long2ice (approved):
Good!

long2ice:
Thanks for contribution!

PR #653, opened 2021-03-03
Do not propagate default_factory to pydantic

Yolley:
## Description
Pydantic never really allowed usage of `default` and `default_factory` at the same time, but in the latest release they added this validation https://github.com/samuelcolvin/pydantic/commit/b742c6f527f32fd0c6eb96f0a35f9e632a3af6bd#diff-bfaaac2ae0a71d9686a272982d29d0bed82bfaacf8b5b30e2b88ed2543e74a3bR124, that causes ValueError when we try to create a pydantic model for Tortoise Model with fields with callable default values. 

## Motivation and Context
This is a fix for my previous pull request https://github.com/tortoise/tortoise-orm/pull/480, I am not sure why with `default_factory` I've also added a `default`
```python
                if callable(field_default):
                    fconfig["default_factory"] = field_default
                    properties[fname] = None
```
May be because I was thinking of removing `default_factory`, because there was a chance of calling this factory function twice (once by pydantic, then by tortoise). So, I guess it would be better to remove `default_factory`, then there is also no need to keep `default` with `None` value. Also these fields will still be optional in json schema, so that end user that views this schema could understand that these fields are not required.

## How Has This Been Tested?

I've run only simple test with `make test` on python3.7 (these changes do not depend on python version), test ran successfully:
```
950 passed, 15 skipped, 4 xfailed in 31.63s
```

## Checklist:

### DIFF
--- a/tortoise/contrib/pydantic/creator.py
+++ b/tortoise/contrib/pydantic/creator.py
@@ -382,9 +382,8 @@ def pydantic_model_creator(
             description = comment or _br_it(fdesc.get("docstring") or fdesc["description"] or "")
             fconfig["description"] = description
             fconfig["title"] = fname.replace("_", " ").title()
-            if field_default is not None:
-                if not callable(field_default):
-                    properties[fname] = field_default
+            if field_default is not None and not callable(field_default):
+                properties[fname] = field_default
             pconfig.fields[fname] = fconfig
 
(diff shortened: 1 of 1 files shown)


long2ice (approved):

long2ice:
Thanks!

PR #654, opened 2021-03-03
Do not pass `Null` value to `generated=True` fields

### LINKED ISSUE #644: 🐛 `generated=True` doesn't work for custom fields when explicitly passing `pk` value on `Model` creation
**Describe the bug**
`query` parameter in [`client.py`](https://github.com/tortoise/tortoise-orm/blob/ef31c19a9da68aed1a7c813e8e9980727ed1490a/tortoise/backends/base/client.py#L146) takes all the SQL parameters (from `Model.regular_columns_all`) on an `INSERT INTO` request when `pk` field is being explicitly passed, so when there are any non-pk (but still generated) fields, it fails to insert `None` into them and gives this error:
```
ERROR: null value in column "index_number" of relation "user" violates not-null constraint
```

**To Reproduce**
I have a model `User`, for example

```python
class User(models.Model):
    name = fields.TextField(null=True)
    index_number = fields.SmallIntField(generated=True, index=True, unique=True)
```

When I create a new model like this
```python
>>> user = User(name='John')
```
it becomes clear that 
```python
>>> user._custom_generated_pk
False
```

But when I do
```python
>>> user = User(id=45, name='John')
>>> user._custom_generated_pk
True
```

Basically, it means that all the fields and their current (ORM- or user-defined) values will be used for an `INSERT INTO` request. The reason is the `execute_insert` method implementation in [`executor.py`](https://github.com/tortoise/tortoise-orm/blob/ef31c19a9da68aed1a7c813e8e9980727ed1490a/tortoise/backends/base/executor.py#L197)

**Expected behavior**
There has to be a way to filter `pk` fields and other generated ones.

**Additional context**
Nothing here yet.

TrDex:
Fixes #644

## Description
Add check if the field is `pk`, so we don't set `self._custom_generated_pk=True` when there is actually no custom generated `pk`

## Motivation and Context
Described in #644.

## How Has This Been Tested?
Run existing tests, all passed.

## Checklist:

### DIFF
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -663,7 +663,7 @@ class Model(metaclass=ModelMeta):
                 passed_fields.add(meta.fields_map[key].source_field)
             elif key in meta.fields_db_projection:
                 field_object = meta.fields_map[key]
-                if field_object.generated:
+                if field_object.pk and field_object.generated:
                     self._custom_generated_pk = True
                 if value is None and not field_object.null:
                     raise ValueError(f"{key} is non nullable field, but null was passed")


long2ice (approved):

long2ice:
Thanks for contribution!

PR #659, opened 2021-03-06
Remove obsolete loop._selector

wankata:
## Description
Remove obsolete loop._selector from contrib/test

## Motivation and Context
The loop._selector was introduced in commit 40f4945 with the ability to provide loop to the initializer. At that point, the loop was BaseSelectorEventLoop, which has a private _selector property.

In 743795d the BaseSelectorEventLoop was replaced by AbstractEventLoop, but the _selector artefact remains.

The problem can be easily identified, when you use uvloop, which implements AbstractEventLoop and is not a SelectorEventLoop.

Resolves: #532
See also: #533

## How Has This Been Tested?
I've tested it manually with an uvloop.Loop and it seems to work properly.

## Checklist:

### DIFF
--- a/tortoise/contrib/test/__init__.py
+++ b/tortoise/contrib/test/__init__.py
@@ -43,7 +43,6 @@ On success it will be marked as unexpected success.
 
 _CONFIG: dict = {}
 _CONNECTIONS: dict = {}
-_SELECTOR = None
 _LOOP: AbstractEventLoop = None  # type: ignore
 _MODULES: List[str] = []
 _CONN_MAP: dict = {}
@@ -100,7 +99,6 @@ def initializer(
     # pylint: disable=W0603
     global _CONFIG
     global _CONNECTIONS
-    global _SELECTOR
     global _LOOP
     global _TORTOISE_TEST_DB
     global _MODULES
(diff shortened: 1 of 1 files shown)


wankata:
Could someone with admin rights on the repo, rerun the checks? The CI stalled for some reason, but I am pretty sure, that they should all pass.

long2ice:
OK

wankata:
Thanks. This is strange, it is hanging again...

long2ice:
Did you run `make style ci` in local?

wankata:
I was just running the make commands as you answered.

It fails on master, develop and of course on my branch with the same issue:
mypy tortoise/ examples/ tests/ conftest.py
Found 19 errors in 7 files (checked 196 source files)
Most of the errors are like:
```
examples/fastapi/main.py:43: error: Untyped decorator makes function "delete_user" untyped
    @app.delete("/user/{user_id}", response_model=Status, responses={404: {"model": HTTPNotFoundError}})
```
But this should emit an error in the ci and not make it hang.

long2ice:
You can merge develop branch first

wankata:
> You can merge develop branch first

What do you mean? I've ran the tests on the develop branch directly. They fail there the exact same way. I've updated my develop from the upstream repo to it's latest state too, and nothing changed.

I would tell, that something in my environment is odd, if you cannot reproduce it, but why the CI is failing although then?
Another interesting question is why the CI just hangs without any errors, but I am failing with typing errors...

And all this to just remove 4 obsolete lines of code :)

stuaxo:
I've tested this patch locally and it means I can use `from tortoise.contrib.test import finalizer, initializer` without hitting an exception because `_selector` doesn't exist any more.

abondar:
Duplicated these changes in https://github.com/tortoise/tortoise-orm/pull/1636

PR #699, opened 2021-04-05
add partitions for postgresql

evstratbg:
I've tried to add PARTITION BY statement to postgresql

Sure, that is not the best PR, so, i'm waiting for comments,  so I can fix it

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -177,6 +177,19 @@ The ``Meta`` class
 
 .. autoclass:: tortoise.models.Model.Meta
 
+    .. attribute:: partition_by
+        :annotation: = None
+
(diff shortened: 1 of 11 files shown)


long2ice:
Thanks! Firstly, you can run `make style ci` in local to make ci pass, secondly, as I konw, `MySQL` also support that, could you support that? And there are different partition types should be support, maybe you can reference `index` design.

evstratbg:
> Thanks! Firstly, you can run `make style ci` in local to make ci pass, secondly, as I konw, `MySQL` also support that, could you support that? And there are different partition types should be support, maybe you can reference `index` design.

ye, I havent use mysql, need to dive in it. Postgres and style will be fixed

Issue #529, opened 2020-10-21
Uvicorn error with the fastapi example provided in the docs


M-e-r-c-u-r-y:
**Describe the bug**
uvicorn error when you run the fastapi example in the docs.
```
WARNING:  Detected file change in 'main.py'. Reloading...
INFO:     Shutting down
INFO:uvicorn.error:Shutting down
INFO:     Waiting for application shutdown.
INFO:uvicorn.error:Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:uvicorn.error:Application shutdown complete.
INFO:     Finished server process [141]
INFO:uvicorn.error:Finished server process [141]
INFO:     Started server process [145]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:uvicorn.error:Application startup complete.
```

**To Reproduce**
Code taken from the fastapi example in the docs.
[models.py](https://tortoise-orm.readthedocs.io/en/latest/examples/fastapi.html#models-py)
[main.py](https://tortoise-orm.readthedocs.io/en/latest/examples/fastapi.html#main-py)
server started with docker using
```
CMD ["uvicorn", "main:app", "--reload", "--workers", "1", "--host", "0.0.0.0", "--port", "8000"]
```

**Expected behavior**
No uvicorn.error during start/shutdown/restart phase

**Additional context**
I suspect starting up, getting and closing the connections to the db is causing the issue.
Possible issue would be how tortoise handles the on startup and shutdown events of the app.

long2ice:
There is no error and log come from uvicorn not tortoise.

M-e-r-c-u-r-y:
@long2ice oh okay. I'll make a thread there then. Thanks for the help :)

mizhgun:
While it is technically not an error, it is more the issue of the Tortoise's FastAPI integration: 

https://github.com/tortoise/tortoise-orm/blob/767cd5d563a36e7795f0331082f6ee51297b661d/tortoise/contrib/fastapi/__init__.py#L93-L96

It uses root logger which by default creates handler with NOSET level. Neither FastAPI nor uvicorn set up the  root logger,  but uvicorn logger propagates log records to root one by default  repeating the messages.

The workaround is either disable propagation for uvicorn loggers or (quick and dirty) monkey patch Tortoise's FastAPI contrib module with custom logger like

```
import logging
from tortoise.contrib import fastapi
fastapi.logging = logging.getLogger('uvicorn')
```

[...]

PR #722, opened 2021-04-19
Avoid using root logger in contrib.fastapi

idmitrievsky:
This relates to #529 and implements a suggested fix.

## Motivation and Context
Refer to #529.


idmitrievsky:
Closed in favour of #752.

Issue #721, opened 2021-04-19
update_or_create: Lock wait timeout exceeded


Yolley:
**Describe the bug**
Since 0.17.2 `update_or_create` stopped working, on mysql (InnoDB 8.0.23, MariaDB 10) it leads to an error
```
  File "/usr/local/lib/python3.8/site-packages/tortoise/models.py", line 1070, in update_or_create
    await instance.update_from_dict(defaults).save(using_db=db)  # type:ignore
  File "/usr/local/lib/python3.8/site-packages/tortoise/models.py", line 930, in save
    await executor.execute_update(self, update_fields)
  File "/usr/local/lib/python3.8/site-packages/tortoise/backends/base/executor.py", line 302, in execute_update
    await self.db.execute_query(
  File "/usr/local/lib/python3.8/site-packages/tortoise/backends/mysql/client.py", line 52, in translate_exceptions_
    raise OperationalError(exc)
tortoise.exceptions.OperationalError: (1205, 'Lock wait timeout exceeded; try restarting transaction')
```
On sqlite it just gets stuck.

**To Reproduce**
In `models.py` we have
```python
from tortoise import fields, Model


class Tournament(Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(unique=True, max_length=64)
```
In `test_mysql.py` we have (db schema already created)
```python
from models import Tournament
from tortoise import Tortoise, run_async


async def init():
    await Tortoise.init(
        db_url='mysql://root:@localhost/test',
        modules={'models': ['models']}
    )
    # Generate the schema
    await Tortoise.generate_schemas()
    await Tournament.all().delete()
    await Tournament.create(id=1, name="test")
    await Tournament.update_or_create(name="test")


if __name__ == "__main__":
    run_async(init())
```
To reproduce we can just try to run `test_mysql.py`, it should run for around a minute (depending on configured lock timeout) and eventually exit with an error
```
tortoise.exceptions.OperationalError: (1205, 'Lock wait timeout exceeded; try restarting transaction')
```

I've also tired to reproduce the bug on sqlite with this
```python
from models import Tournament
from tortoise import Tortoise, run_async


async def init():
    # Here we create a SQLite DB using file "db.sqlite3"
    #  also specify the app name of "models"
    #  which contain models from "app.models"
    await Tortoise.init(
        db_url='sqlite://db.sqlite3',
        modules={'models': ['models']}
    )
    # Generate the schema
    await Tortoise.generate_schemas()
[...]

PR #733, opened 2021-04-23
pass transaction wrapper to using_db in get_or_create/update_or_create

Yolley:
Pass wrapper created by `in_transaction` as `using_db` value to ensure that all actions in get_or_create/update_or_create are done inside created transaction.

## Description
The problem is that `db = using_db or cls._choose_db(True)` returns client that was created outside of transaction and does not have access to lock that is created inside `update_or_create` by calling `select_for_update`, what leads to errors described in attached issue. To ensure that we execute everything inside of our transaction we need to pass `TransactionWrapper` that is created by `in_transaction` in `using_db` value.
Also, as I see, even without this fix all tests run successfully. My guess is that it happens because we already have a transaction created by `test.TestCase`, but in a real environment `update_or_create` just does not work without this fix.

## Motivation and Context
To fix https://github.com/tortoise/tortoise-orm/issues/721

## How Has This Been Tested?
I've run tests on mysql and sqlite.
On sqlite there are 3 tests that are not related to this PR (they are related to delete with limit) that fail.
```
3 failed, 968 passed, 18 skipped, 4 xfailed in 36.73s
```
On mysql all tests ran successfully
```
977 passed, 13 skipped, 3 xfailed, 1 warning in 38.75s
```

## Checklist:

### DIFF
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -1022,12 +1022,12 @@ class Model(metaclass=ModelMeta):
         if not defaults:
             defaults = {}
         db = using_db or cls._choose_db(True)
-        async with in_transaction(connection_name=db.connection_name):
+        async with in_transaction(connection_name=db.connection_name) as conn:
             instance = await cls.filter(**kwargs).first()
             if instance:
                 return instance, False
             try:
-                return await cls.create(**defaults, **kwargs, using_db=using_db), True
+                return await cls.create(**defaults, **kwargs, using_db=conn), True
             except IntegrityError:
                 try:
                     return await cls.get(**kwargs), False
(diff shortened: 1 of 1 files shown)


tortoise/models.py
```diff
@@ -1022,12 +1022,12 @@ async def get_or_create(
...
             defaults = {}
         db = using_db or cls._choose_db(True)
-        async with in_transaction(connection_name=db.connection_name):
+        async with in_transaction(connection_name=db.connection_name) as conn:
             instance = await cls.filter(**kwargs).first()
             if instance:
                 return instance, False
             try:
-                return await cls.create(**defaults, **kwargs, using_db=using_db), True
+                return await cls.create(**defaults, **kwargs, using_db=conn), True
```
long2ice, line 1030: Just update to ` return await cls.create(**defaults, **kwargs, using_db=db), True`

tortoise/models.py
```diff
@@ -1064,10 +1064,10 @@ async def update_or_create(
         if not defaults:
             defaults = {}
         db = using_db or cls._choose_db(True)
-        async with in_transaction(connection_name=db.connection_name):
+        async with in_transaction(connection_name=db.connection_name) as conn:
             instance = await cls.select_for_update().get_or_none(**kwargs)
```
long2ice, line 1068: `instance = await cls.select_for_update().using_db(db).get_or_none(**kwargs)`

long2ice (approved):

long2ice:
Thanks! Please reference my suggests and update changelog

PR #737, opened 2021-04-25
Add more logs for 'Can't create schema due to cyclic fk references' e…

enjoy-binbin:
when i run get_create_schema_sql function for test, i got `Can't create schema due to cyclic fk references` error.

This error does not contain any details. So very confusing because in my project, it does not have any fk table or something

So i take a look on source code, and added some log to debug

At last i find out there are two different modes that defind the same table name in my project ...  Very careless

So i think i would be helpful if the error message contain more detail

Feel free to close this pr if un-needed

### DIFF
--- a/tortoise/backends/base/schema_generator.py
+++ b/tortoise/backends/base/schema_generator.py
@@ -419,27 +419,36 @@ class BaseSchemaGenerator:
         self._get_models_to_create(models_to_create)
 
         tables_to_create = []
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
by the way, any plan with rollback `generate_schemas` function

enjoy-binbin:
ping @long2ice what do you think about this one?

Issue #741, opened 2021-04-27
Provide more context for DoesNotExist Exception


spacemanspiff2007:
With FastAPI it's possible to register a central exception handler for certain errors.
It's elegant to register a listener for ``DoesNotExist``, because then the error handling doesn't have to be created in every route.
However I'd like to give a proper error message what kind of Model did not exist and with what primary key it was queried.
Currently that is not possible because ``DoesNotExist`` doesn't provide the required context.

Ideal implementation:

````python
from tortoise.exceptions import DoesNotExist

 try:
    obj = await MyModel.get(id=1)
except DoesNotExist as e:
    assert e.model is MyModel
    assert e.pk_name == 'id'
    assert e.pk_value == 1
````

PS: It would be nice if we could rename ``DoesNotExist`` to ``DoesNotExistError``

**Comment by marcoaaguiar** [comment]
Hi, spacemanspiff2007, I agree with you. More information is always better than less information (at least for errors) :)
Just keep in mind that `Model.get` works not only queries by pk fields but for any query that returns a single result in the query set.

**Comment by spacemanspiff2007** [comment]
> Just keep in mind that ``Model.get`` works not only queries by pk fields but for any query that returns a single result in the query set.

You are absolutely right! Maybe we can find a way to attach this information to the error or maybe we can define a shortcut to query a Model by primary key, e.g.
```python
obj = await MyModel[1]
```

marcoaaguiar:
Regarding your first point, do you mean attaching the query information? ({"id":1},)

spacemanspiff2007:
yes, exactly

marcoaaguiar:
Regarding your second point, it should be implemented already (haven't checked tho) #386

PR #759, opened 2021-05-06
Added error for lookup by primary key

spacemanspiff2007:
This PR allows to generate nice error messages with a central exception handler (e.g. in FastAPI) - see #741 for initial idea.
I've changed the exception to ```ObjectDoesNotExistError``` since the user explicitly specifies a primary key which makes this a little different than the filter function.
If this will get merged, I'll add something to the docs on how to use it.

```python
async def run():
    try:
        await MyModel[99]
    except ObjectDoesNotExistError as e:
        print(e.model)
        print(e.pk_name)
        print(e.pk_val)
        print(e)
```

Output:
```
<class '__main__.MyModel'>
id
99
models.MyModel has no object with id=99
```


long2ice:
There is `DoesNotExist` already, why new one?

spacemanspiff2007:
This one is only thrown when an item gets accessed by primary key.
If no item is found the error contains the primary key name and value.

That way it is possible to register a central error handler (e.g. in FastAPI) and 
return a http response 404 which can contain the Model and pk.

e.g.
```json
{
    "item": "MyModel",
    "id": 99,
    "msg": "MyModel with id 99 does not exist"
}
```
which is way better than just
```json
{
    "msg": "Object not found"
}
```

Additionally it is not necessary any more to handle the ObjectNotFoundError in every route.
This leads to simpler and easier to read code.

Imho it makes sense to raise different errors for accessing an item by primary key and for filtering and searching an item.
For the first one I am expecting it to exist (otherwise I wouldn't pass a pk) while with the other one I am searching for an occurrence.

spacemanspiff2007:
I have a FastApi application where the user interacts with the db.
With this PR I don't have to check every time I get something passed in:

```python
@router.get(
    path="/filter",
    response_model=List[MyModelApiOut],
    status_code=status.HTTP_200_OK,
)
async def get_models_of_type(type: int, ecu: str):
    type = await MyModelType[type]

    return [MyModelApiOut.from_orm(k) for k in await MyModel.filter(type=type)]
```

Without this PR I have to do this:
```python
@router.get(
    path="/filter",
    response_model=List[MyModelApiOut],
    status_code=status.HTTP_200_OK,
)
async def get_models_of_type(type: int, ecu: str):
    type = await MyModelType.get_or_none(pk=type)
    if type is None:
        return Response404(pk_name='id', pk_value=type, msg=f'MyModelType with id {pk_value} does not exist')

    return [MyModelApiOut.from_orm(k) for k in await MyModel.filter(type=type)]
```

I have to explicitly check for every item that get's passed in by the user and issue a corresponding error message.
This may not be much for one endpoint but it quickly adds up and makes the code unnecessarily complex, especially if the user passes in multiple primary keys in the route.

I really think differentiating between accessing and item with the primary key and searching for an item with the filter function is a smart design choice and I would really like to see tortoise-orm go that way.

spacemanspiff2007:
@long2ice 
It would really simply things if we can get the pk name, pk value and model on the Exception.
What do I have to do to get this or something similar merged?
I created a new exception because in this context we expect a single object that is queried by pk in contrast to the query by filter values.

abondar:
@spacemanspiff2007 would you like to update this PR? 
I am okay with merging it after update

spacemanspiff2007:
@abondar like this?

abondar (approved):

Issue #642, opened 2021-02-17
Add 'optional' argument to 'pydantic_model_creator'.


synchronizing:
While the `pydantic_model_creator` feature is useful, the `exclude` and `include` options are not as versatile as they need to be for the function to be as flexible as it likely should. Consider the following:

```python
from tortoise import models, fields

class Model(models.Model):
    uid: int
    first_name: str
    last_name: str
    number: int
```

Not to get into specific of _which_ framework, say we have the following routes:

* `GET /model/{uid}`
* `POST /model/filter`

For the `GET` above, it's easy to see how 

```python
from tortoise.contrib.pydantic import pydantic_model_creator
from . import model

ModelIn = pydantic_model_creator(model.Model, include=('uid',))
```

Is useful and handy. However, consider the case where one may filter via `first_name`, `last_name`, *or* `number`. This would required a Pydantic schema that is:

```python
from pydantic import BaseModel

class ModelFilter(BaseModel):
    first_name: Optiona[str]
    last_name: Optional[str]
    number: Optional[int]
```

It would be handy if `pydantic_model_creator` took another argument, `optional`, to include such options. Therefore allowing us to do:

```python
ModelFilter = pydantic_model_creator(model.Model, exclude=('uid',), optional=('first_name', 'last_name', 'number'))
```

This relates to #571 as well.

lsabi:
In order to avoid listing all the fields, I'd rather either pass the list of field names (as proposed by this issue) or `True` to the parameter `optional`. In the latter case, it will work as if all fields were passed in the `optional` parameter.

This way, if the data model changes, one does not have to change all the pydantic declarations for that model (in a database upgrade, there may be several).

Issue #571, opened 2020-11-26
[QUESTION] Is there a way to specify fields to be optional in pydantic_model_creator?


ihsanmohamad:
As the title says, how do I make some fields optional using the pydantic_model_creator for my fastapi project? If there is no way, i would go with manual pydantic model. Thank you in advance.

timwford:
If you're after what I think you're after, `pydantic_model_creator` checks to see if your Tortoise database model has the `null=` set. By default, this is False (and thus the item is not optional) but it can be set to True and the resulting Pydantic model will be labeled as optional! 

An example:
```
class OptionalExample(Model):
    optional = fields.IntField(description="An example of an optional field")
    not_optional = fields.IntField(description="An example of a non-optional field", null=True)
```

Which results in the following schema in FastAPI:

<img width="468" alt="Screen Shot 2020-12-12 at 5 03 08 PM" src="https://user-images.githubusercontent.com/48221412/101999600-ebe23500-3c9b-11eb-9698-9bfe0ae1f5d8.png">

malikoth:
I would like the same for a PATCH method I'm writing.  I want to be able to make all fields optional, even though none of the columns might be nullable, because the data from the Pydantic model gets applied on top of an existing valid model.

rsprudencio:
> If you're after what I think you're after, `pydantic_model_creator` checks to see if your Tortoise database model has the `null=` set. By default, this is False (and thus the item is not optional) but it can be set to True and the resulting Pydantic model will be labeled as optional!
> 
> An example:
> 
> ```
> class OptionalExample(Model):
>     optional = fields.IntField(description="An example of an optional field")
>     not_optional = fields.IntField(description="An example of a non-optional field", null=True)
> ```
> 
> Which results in the following schema in FastAPI:
> 
> <img alt="Screen Shot 2020-12-12 at 5 03 08 PM" width="468" src="https://user-images.githubusercontent.com/48221412/101999600-ebe23500-3c9b-11eb-9698-9bfe0ae1f5d8.png">

Isn't your example the other way around? In your code, you mention that adding `null=True` makes it non-optional field.

PR #770, opened 2021-05-12
Add `optional` parameter to `pydantic_model_creator`

Daste745:
## Description
This addition allows specifying optional model parameters without modifying the database schema.

## Motivation and Context
This is particularly useful when declaring models for updating or querying data. This was explained in more detail and with an example in #642 and requested in #571.
My use case is creating a PATCH endpoint which allows users to update their username, password, bio etc. For this I would need to write a new model by hand or do some tricks when declaring it, which I don't want to do

## How Has This Been Tested?
I tested this with my FastAPI application. I created separate Pydantic models for reading, creating and updating user data: 
The read-only model excludes the "password" field (for obvious reasons).
The create model excludes readonly fields.
The update model excludes readonly + makes "username", "mail" and "password" optional, which allows users to update single fields at a time.

I wrote two test cases which check the differences between a model with and without "optional" fields.

## Checklist:


Daste745:
I added a "User" test model and wrote two passing test cases.
Please let me know if there needs to be any more work done on this feature.

AhmedSoli:
Do we have an update here?

Daste745:
I believe this PR is ready for a merge, although I don't understand the conflict in `tests/testmodels.py`. I've been using these changes in one of my projects (not active at the moment) and have not seen any issues with it.

GonerGenesis:
any news on this?

long2ice:
OK, just need resolve the conflicts

Daste745:
Sorry this took so long, I forgot about this PR. Everything should be merging now, although I can see errors in CI on my fork, which I'm not sure how to fix.
```
mypy tortoise/ examples/ tests/ conftest.py
tests/contrib/test_pydantic.py: note: In member "setUp" of class "TestPydanticUpdate":
tests/contrib/test_pydantic.py:1298: error: Return type
"Coroutine[Any, Any, None]" of "setUp" incompatible with return type "None" in
supertype "TestCase"
        async def setUp(self) -> None:
        ^
Found 1 error in 1 file (checked 220 source files)
```

Daste745:
Mypy isn't throwing the error anymore, should be good to go. Tests are passing locally (tested with sqlite only).

Daste745:
Hi @long2ice, could you please take a look at this? I think it would be a nice feature to have for some people.

long2ice (approved):

long2ice:
Thanks!

PR #786, opened 2021-06-04
Handled application_name in asyncpg client arguments

vinayinvicible:
Handled application_name and explicit server_settings arguments for asyncpg


long2ice (approved):

long2ice:
Thanks!

Issue #631, opened 2021-02-01
 Can't set timezone-naive datetimes


Dnomin:
**Describe the bug**
Can't set timezone-naive datetimes

**To Reproduce**
1. Model with two `DatetimeField`s and type `sqlalchemy.DateTime(timezone=False)`
2. Save timezone-naive datetime to first field (All ok)
3. Save timezone-naive datetime to second field (Error, can't compare aware and naive dts)

`values` from `backends/base/executor.py:execute_update`:
* from first execute: `[datetime.datetime(2021, 2, 2, 0, 0), None]`
* from second: `[datetime.datetime(2021, 2, 2, 0, 0, tzinfo=<UTC>), datetime.datetime(2021, 3, 17, 14, 38, 11)]`

So, as we can see, on second execute, first field value have UTC timezone

**Expected behavior**
Timezone-naive fields

**Additional context**
PostgreSQL : latest
Tortoise-orm : latest
Alembic with SQLAlchemy (for migrations) : latest

lsabi:
Can you be more explicit? It seems strange, but maybe because I'm not understanding the context.

In particular, do you mind posting the model's structure? And explain what you mean with `second execute`? Is it an update query of the same record or a new record?

Thanks

xSandie:
We meet the same strange problem. So I have to roll back to 0.15.18，there is no problem with this version.

xSandie:
I have an old project using the version 0.15.18,  when I upgrade to the latest version 0.16.18, all code with `tortoise.fields.DatetimeField(auto_now_add=True)` have the error `tortoise.exceptions.OperationalError: invalid input for query argument $1: datetime.datetime(2020, 6, 6, 7, 12, 30,... (can't subtract offset-naive and offset-aware datetimes)`

### Possible causes
When I read your code, I found that the class `tortoise.fields.DatetimeField` now using `SQL_TYPE = "TIMESTAMPTZ"` in postgresql [source code,](https://github.com/tortoise/tortoise-orm/blob/dee4de436e8f83e894833bd366d30e0cec400d0c/tortoise/fields/data.py#L29) it will cause some problem with the old project. I will appreciate that if your team could mention this in the changelog.

lsabi:
You have to change the configuration during initialization of tortoise.

Here you see a sample config, though it's without the timezone related configuration, but we'll get there in a moment.

[...]

PR #787, opened 2021-06-07
when USE_TZ=False, datetimes must be timezone naive

rafalstapinski:
## Description

This changes the USE_TZ setting to  control whether Tortoise is dealing with timezone aware or timezone naive datetimes. 
When USE_TZ=False, all datetimes should be timezone naive.
When USE_TZ=True, all datetimes should be timezone aware, with the timezone being specified by the TIMEZONE setting. 

## Motivation and Context
Currently, USE_TZ is a misnomer. It suggests you can either use TZ aware datetimes, or you can use TZ naive datetimes. It's implementation does something different - all datetimes are aware, but in USE_TZ=False they are UTC. 

This makes the USE_TZ setting useless, as `USE_TZ=False` == `USE_TZ=True, TIMEZONE=UTC`.
https://github.com/tortoise/tortoise-orm/issues/631

## Checklist:

### DIFF
--- a/tortoise/backends/sqlite/executor.py
+++ b/tortoise/backends/sqlite/executor.py
@@ -43,10 +43,7 @@ def to_db_datetime(
         self.auto_now
         or (self.auto_now_add and getattr(instance, self.model_field_name, None) is None)
     ):
-        if timezone.get_use_tz():
-            value = datetime.datetime.now(tz=pytz.utc)
-        else:
-            value = datetime.datetime.now(tz=timezone.get_default_timezone())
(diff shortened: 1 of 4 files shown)


rafalstapinski:
@long2ice I'm hoping to fix some of the inconsistencies in the `USE_TZ` setting. I've been working through postgres changes, but would like access to running the testing pipeline to speed up validation. Any chance you could approve me running tests against this?

rafalstapinski:
@long2ice bump on the above, any help would be appreciated

PR #802, opened 2021-06-23
Tiny doc change: Added info about sort order

spacemanspiff2007:
Added information about sort order because currently it's missing in the docs.


long2ice:
Thanks!

Issue #776, opened 2021-05-17
annotate + select_related - IndexError


kdelinx:
**Cathed bug annotate + select_related**
My model has there are fields:
```python
class ModelName(BaseDBModel):
    title = fields.CharField(255)
    description = fields.TextField()
    settings = fields.JSONField()

    logo = fields.ForeignKeyField(
        'modules.Image', related_name='logotype',
        on_delete=fields.CASCADE, null=True,
        description='logotype of model'
    )
```

I have simple queryset, like this:
```python
establishment_queryset = (
       ModelName
       .all()
       .select_related('logo')
       .prefetch_related('images')
       .annotate(score_rating=Max('ratings__score'))
       .order_by('-score_rating')
       .limit(20)
)
```
after calling pydantic_queryset_creator with `from_queryset(queryset)`, i catch error with pair from issue.

In traceback I have:
```
  File "./handlers/establishments_view.py", line 28, in get
    list_instances = await schema_.from_queryset(establishment_queryset)
  File "./.venv/lib/python3.8/site-packages/tortoise/contrib/pydantic/base.py", line 139, in from_queryset
    __root__=[submodel.from_orm(e) for e in await queryset.prefetch_related(*fetch_fields)]
  File "./.venv/lib/python3.8/site-packages/tortoise/queryset.py", line 871, in _execute
    instance_list = await self._db.executor_class(
  File "./.venv/lib/python3.8/site-packages/tortoise/backends/base/executor.py", line 138, in execute_select
    **dict(
  File "./.venv/lib/python3.8/site-packages/tortoise/backends/base/executor.py", line 141, in <lambda>
    lambda x: x.split(".")[1],
IndexError: list index out of range
```
breakepoint in code - `keys[current_idx : current_idx + index]`
example:
`['score_rating', 'model_name__logo.id', 'model_name__logo.lg', 'model_name__logo.url']` 
for mapping there are fields lamda tying to extract field names after dot in string path, and first arg `score_rating` shouldn't be here, as i understand correct logic case of `.select_related(*args)`

long2ice:
Could you show full code?

PR #823, opened 2021-07-15
Include annotations length in select columns count to fix IndexError

Yolley:
## Description
Currently when we try to use annotate and select_related in one query, there will be an error
```
    lambda x: x.split(".")[1],
IndexError: list index out of range
```
It happens, because when we calculate columns to be selected from a model we include only number of columns that are presented in the model https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/queryset.py#L836. When we use `select_related`, in executor we try to map columns by their indexes from `select_related_idx ` https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/base/executor.py#L137 . So, because we don't include number of annotation columns in our calculation, we get an error, because executor tries to fetch annotation column and split it by `.`, expecting a column from the related model.
Also I've added a test for this issue, without my fix this test would fail.

## Motivation and Context
This is a fix for this issue https://github.com/tortoise/tortoise-orm/issues/776

## How Has This Been Tested?
I've run `make test`, there are 4 failed tests
```
4 failed, 975 passed, 25 skipped, 4 xfailed in 27.72s
```
But they failed even without my changes.
Also I've run `test_mysql`, it completed successfully:
```
985 passed, 20 skipped, 3 xfailed, 1 warning in 42.44s
```

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Changelog
 0.17.6
 ------
 - Add `RawSQL` expression.
+- Fix columns count with annotations in `_make_query`
 
 0.17.5
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ Changelog
 0.17.6
 ------
 - Add `RawSQL` expression.
-- Fix columns count with annotations in `_make_query`
+- Fix columns count with annotations in `_make_query`. (#776)
 
 0.17.5
 ------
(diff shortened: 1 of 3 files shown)


Yolley:
@long2ice please take a look, this is a small code change, and I can't think of any edge cases that would not work because of it. But the bug needs to be fixed, currently in my environment I need to use `prefetch_related` because of it (which leads to additional db queries).

long2ice (approved):

long2ice:
Thanks! Could you update changelog?

Yolley:
> Thanks! Could you update changelog?

Updated changelog and rebased the branch.

long2ice:
Great!

Issue #825, opened 2021-07-17
select_related on one2one field returns objects with unfilled related field


urm8:
**Describe the bug**
the example models look like this
```
class B(Model):
      id = fields.UUIDField(pk=True)
class A(Model):
    the_b = fields.OneToOneField(
        "models.B", related_name="the_a", on_delete=fields.SET_NULL, null=True
    )

then i get it like:
a = A.all().select_related('the_b').get(id=<some_uuid>)
and got:
a.the_b is None 
```
is this intended behaviour and do you have any plans to fix it ?

urm8:
the most stupid solution i came up with is in https://github.com/tortoise/tortoise-orm/pull/826

PR #826, opened 2021-07-17
fix bug with select_related not yielding null #up

urm8:
Fixed (a bug ???) when select_related yielded unfilled instances of related objects instead of just nulls.
## Description
Ran into this unexpected behaviour with something like this:
```
class B(Model):
      id = fields.UUIDField(pk=True)
class A(Model):
    the_b = fields.OneToOneField(
        "models.B", related_name="the_a", on_delete=fields.SET_NULL, null=True
    )
```
then i got it like:
```
a1 = await A.create(the_b=await B.create())
a2 = await A.create(the_b=None)
a = A.all().select_related('the_b').get(id=a2.id)
# and for some reason the next checks failed
assert a.the_b is None
assert a.the_b == a2.the_b
```
## Motivation and Context

Looks like this packages follows the ideas of django-orm and duplicates its best features in most cases, so i thought, like at least select_related behaviour should be same as it in django-orm. 
linked to this issue https://github.com/tortoise/tortoise-orm/issues/825.
## How Has This Been Tested?
tested this with code like the one above. + as import dep in my pet proj, just assured that it don't gives me issues anymore.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Changelog
 - Fix columns count with annotations in `_make_query`. (#776)
 - Make functions nested. (#828)
 - Add `db_constraint` in field describe.
+- Fix `select_related` behaviour for forward relation. (#825)
 
 0.17.5
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,13 +9,16 @@ Changelog
 
 0.17
 ====
+0.17.7
+------
+- Fix `select_related` behaviour for forward relation. (#825)
+
 0.17.6
 ------
 - Add `RawSQL` expression.
 - Fix columns count with annotations in `_make_query`. (#776)
 - Make functions nested. (#828)
 - Add `db_constraint` in field describe.
-- Fix `select_related` behaviour for forward relation. (#825)
 
 0.17.5
 ------
--- a/tests/test_relations.py
+++ b/tests/test_relations.py
@@ -5,9 +5,13 @@ from tests.testmodels import (
     DoubleFK,
     Employee,
     Event,
+    Extra,
+    Pair,
     Reporter,
+    Single,
     Team,
     Tournament,
+    UUIDFkRelatedNullModel,
 )
 from tortoise.contrib import test
 from tortoise.exceptions import FieldError, NoValuesFetched
@@ -300,6 +304,58 @@ class TestRelations(test.TestCase):
         self.assertEqual(event.reporter, reporter)
         self.assertTrue(hasattr(event, "tournament_name"))
         self.assertEqual(event.tournament_name, tournament.name)
+
+    async def test_select_related_sets_null_for_null_fk(self):
+        """Test that select related yields null for fields with nulled fk cols."""
+        related_dude = await UUIDFkRelatedNullModel.create(name="Some model")
+        await related_dude.fetch_related("parent")  # that is strange :)
+        related_dude_fresh = (
+            await UUIDFkRelatedNullModel.all().select_related("parent").get(id=related_dude.id)
+        )
+        self.assertIsNone(related_dude_fresh.parent)
+        self.assertEqual(related_dude_fresh.parent, related_dude.parent)
+
+    async def test_select_related_sets_valid_nulls(self) -> None:
+        """When we select related objects, the data we get from db should be set to corresponding attribute."""
+        left_2nd_lvl = await DoubleFK.create(name="second leaf")
+        left_1st_lvl = await DoubleFK.create(name="1st", left=left_2nd_lvl)
+        root = await DoubleFK.create(name="root", left=left_1st_lvl)
+
+        retrieved_root = (
+            await DoubleFK.all()
+            .select_related("left__left__left", "right")
+            .get(id=getattr(root, "id"))  # noqa
+        )
+        self.assertIsNone(retrieved_root.right)  # ignore
+        self.assertEqual(retrieved_root.left, left_1st_lvl)  # ignore
+        self.assertEqual(retrieved_root.left.left, left_2nd_lvl)  # type: ignore
+
+    async def test_no_ambiguous_fk_relations_set(self):
+        """Basic select_related test cases provided by @https://github.com/Terrance.
+
+        The idea was that on the moment of writing this feature, there were no way to correctly set attributes for
+        select_related fields attributes.
+        src: https://github.com/tortoise/tortoise-orm/pull/826#issuecomment-883341557
+        """
+
+        extra = await Extra.create()
+        single = await Single.create(extra=extra)
+        await Pair.create(right=single)
+        pair = (
+            await Pair.filter(id=1)
+            .select_related("left", "left__extra", "right", "right__extra")
+            .get()
+        )
+        self.assertIsNone(pair.left)
+        self.assertEqual(pair.right.extra, extra)
+        single = await Single.create()
+        await Pair.create(right=single)
+        pair = (
+            await Pair.filter(id=2)
+            .select_related("left", "left__extra", "right", "right__extra")
+            .get()
+        )
+        self.assertIsNone(pair.right.extra)  # should be None
 
 
 class TestDoubleFK(test.TestCase):
--- a/tests/testmodels.py
+++ b/tests/testmodels.py
@@ -305,6 +305,7 @@ class NoID(Model):
 
 
 class UniqueName(Model):
+    id = fields.IntField(pk=True)
     name = fields.CharField(max_length=20, null=True, unique=True)
 
 
@@ -743,3 +744,32 @@ class ManagerModel(Model):
 
     class Meta:
         manager = StatusManager()
+
+
+class Extra(Model):
+    """Dumb model, has no fk.
+    src: https://github.com/tortoise/tortoise-orm/pull/826#issuecomment-883341557
+    """
+
+    id = fields.IntField(pk=True)
+    # currently, tortoise don't save models with single pk field for some reason \_0_/
+    some_name = fields.CharField(default=lambda: str(uuid.uuid4()), max_length=64)
+
+
+class Single(Model):
+    """Dumb model, having single fk
+    src: https://github.com/tortoise/tortoise-orm/pull/826#issuecomment-883341557
+    """
+
+    id = fields.IntField(pk=True)
+    extra = fields.ForeignKeyField("models.Extra", related_name="singles", null=True)
+
+
+class Pair(Model):
+    """Dumb model, having double fk
+    src: https://github.com/tortoise/tortoise-orm/pull/826#issuecomment-883341557
+    """
+
+    id = fields.IntField(pk=True)
+    left = fields.ForeignKeyField("models.Single", related_name="lefts", null=True)
+    right = fields.ForeignKeyField("models.Single", related_name="rights", null=True)
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -56,7 +56,9 @@ class BaseExecutor:
         db: "BaseDBAsyncClient",
         prefetch_map: "Optional[Dict[str, Set[Union[str, Prefetch]]]]" = None,
         prefetch_queries: Optional[Dict[str, List[Tuple[Optional[str], "QuerySet"]]]] = None,
-        select_related_idx: Optional[List[Tuple["Type[Model]", int, str, "Type[Model]"]]] = None,
+        select_related_idx: Optional[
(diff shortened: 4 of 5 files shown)


long2ice (commented):
Thanks! You should merge develop from origin because some file is changed, and change like following maybe better, what do you think? And I think `gh-pages.yml` is no need to change.
```py
for model, index, model_name, parent_model in self.select_related_idx[1:]:
    related_values = values[current_idx : current_idx + index]
    if not any(related_values):
        obj = None
    else:
        obj = model._init_from_db(
            **dict(
                zip(
                    map(
                        lambda x: x.split(".")[1],
                        keys[current_idx : current_idx + index],  # noqa
                    ),
                    related_values,  # noqa
                )
            )
        )
    for ins in instances:
        if isinstance(ins, parent_model):
            setattr(ins, f"_{model_name}", obj)
    instances.append(obj)
    current_idx += index
```
And finally, you need update changelog, thanks!

urm8:
how is that possible ? 
```
┌─[mm@m8] - [~/workspace/github.com/urm8/tortoise-orm] - [Пн июл 19, 20:21]
└─[$] <git:(feature/select_related_on_null_yields_none)> grep -A 5 '0.17.6' CHANGELOG.rst
0.17.6
------
- Add `RawSQL` expression.

0.17.5
------
┌─[mm@m8] - [~/workspace/github.com/urm8/tortoise-orm] - [Пн июл 19, 20:21]
└─[$] <git:(feature/select_related_on_null_yields_none)> poetry version -s 
0.17.5
┌─[mm@m8] - [~/workspace/github.com/urm8/tortoise-orm] - [Пн июл 19, 20:22]
└─[$] <git:(feature/select_related_on_null_yields_none)> 
```
Anyways rebased, added changelog, also like your approach much more, thx

> Thanks! You should merge develop from origin because some file is changed, and change like following maybe better, what do you think? And I think `gh-pages.yml` is no need to change.
> 
> ```python
> for model, index, model_name, parent_model in self.select_related_idx[1:]:
>     related_values = values[current_idx : current_idx + index]
>     if not any(related_values):
>         obj = None
>     else:
>         obj = model._init_from_db(
>             **dict(
>                 zip(
>                     map(
>                         lambda x: x.split(".")[1],
>                         keys[current_idx : current_idx + index],  # noqa
>                     ),
>                     related_values,  # noqa
>                 )
>             )
>         )
>     for ins in instances:
>         if isinstance(ins, parent_model):
>             setattr(ins, f"_{model_name}", obj)
>     instances.append(obj)
>     current_idx += index
> ```
> 
> And finally, you need update changelog, thanks!

like your idea, but it will introduce a bug. When there is more than one o2o field that is also declared in select_related, fields will overwrite each other. I could go with solution alike you provided above, but it will require really strange break statement in this for block:
```
                    for ins in instances:
                        if isinstance(ins, parent_model):
                            setattr(ins, f"_{model_name}", obj)
                        if obj is None:
                            break
``` 
which will break readability, imho.

urm8:
btw, is it ok that my diff contains updated lock file ? I just used ```make up``` thing locally :(

Terrance:
So I also encountered this a few days ago, in the process of migrating from `prefetch_related()` to `select_related()` where applicable.  I had been playing with a patch of my own, which can currently be found at 7ba269b56acf.

I believe this to have the desired behaviour, and also fixes what looks like a potential issue with a model that has two foreign keys for the same related model by explicitly looking for the parent model.  The downside is it relies on the format of the column names, which is probably not desirable.

I'm happy to open a separate PR if it's worth pursuing this route, though it feels like this isn't the right place to be parsing the structure like this.

long2ice (approved):

long2ice:
Could you merge develop? Something is wrong

urm8:
> Could you merge develop? Something is wrong

what exactly ? this branch is already on top of develop 
```
┌─[mm@m8] - [~/workspace/github.com/urm8/tortoise-orm] - [Вт июл 20, 10:52]
└─[$] <git:(feature/select_related_on_null_yields_none)> git remote -v 
origin  git@github.com:urm8/tortoise-orm.git (fetch)
origin  git@github.com:urm8/tortoise-orm.git (push)
tort    git@github.com:tortoise/tortoise-orm.git (fetch)
tort    git@github.com:tortoise/tortoise-orm.git (push)
┌─[mm@m8] - [~/workspace/github.com/urm8/tortoise-orm] - [Вт июл 20, 10:52]
└─[$] <git:(feature/select_related_on_null_yields_none)> git merge tort/develop 
Already up to date.
┌─[mm@m8] - [~/workspace/github.com/urm8/tortoise-orm] - [Вт июл 20, 10:52]
└─[$] <git:(feature/select_related_on_null_yields_none)> gpf!
Everything up-to-date

```

long2ice:
Wondering `ci.yml` did not run

long2ice:
What result is run `make style ci` in your local?

urm8:
> What result is run `make style ci` in your local?

oh, style check failed for test

Terrance:
The changes here as they stand don't consistently handle models with two references to the same related model, if also fetching a relation from each of those.  The proposed fix only assigns `None` to the first instance in `instances` (which is just `instance` anyway?) which doesn't seem quite right, but neither does assigning all instances of the same type as is done currently for non-null relations.  As an example:

```python
class Extra(Model):
    id = IntField(pk=True)

class Single(Model):
    id = IntField(pk=True)
    extra = ForeignKeyField("db.Extra", related_name="singles", null=True)

class Pair(Model):
    id = IntField(pk=True)
    left = ForeignKeyField("db.Single", related_name="lefts", null=True)
    right = ForeignKeyField("db.Single", related_name="rights", null=True)
```

```python
>>> extra = await Extra.create()
>>> single = await Single.create(extra=extra)
>>> await Pair.create(right=single)
<Pair: 1>
>>> pair = await Pair.filter(id=1).select_related("left", "left__extra", "right", "right__extra").get()
>>> pair.left is None
True
>>> pair.right.extra  # should be <Extra: 1>
<tortoise.queryset.QuerySet object at ...>

>>> single = await Single.create()
>>> await Pair.create(right=single)
<Pair: 2>
>>> pair = await Pair.filter(id=2).select_related("left", "left__extra", "right", "right__extra").get()
>>> pair.right.extra  # should be None
<tortoise.fields.relational._NoneAwaitable object at ...>
```

<details>
<summary>More test cases</summary>

```python
class Extra(Model):
    id = IntField(pk=True)

class Single(Model):
    id = IntField(pk=True)
    first = ForeignKeyField("db.Extra", related_name="firsts", null=True)
    second = ForeignKeyField("db.Extra", related_name="seconds", null=True)

class Pair(Model):
    id = IntField(pk=True)
    left = ForeignKeyField("db.Single", related_name="lefts", null=True)
    right = ForeignKeyField("db.Single", related_name="rights", null=True)
```

```python
extra = await Extra.create()
singles = (
    await Single.create(first=extra, second=extra),
    await Single.create(first=extra),
    await Single.create(second=extra),
    await Single.create(),
)
for single in singles:
    await Pair.create(left=single, right=single)
    await Pair.create(left=single)
    await Pair.create(right=single)
await Pair.create()
pairs = await Pair.all().select_related(
    "left", "left__first", "left__second", "right", "right__first", "right__second",
)
```

In here I'm seeing a lot of `NoneAwaitable` instances that aren't being replaced with `None`, but also a couple of query sets that are missing their values:

* `pairs[0]` is a full `single = Single(first=extra, second=extra); pair = Pair(left=single, right=single)`, but `pair.left.first` is left as a `QuerySet` when fetched via `select_related()` (unlike `pair.right.first`)
* `pairs[-2]` is effectively `Pair(right=Single())`, but both `pair.right.first` and `pair.right.second` are `NoneAwaitable`

Another possible case of identifying the wrong parent instance would be a chain of relations like `A->B->A->B` -- assigning the inner-most `B` instance would end up finding the root `A` instance instead of its actual parent.
</details>

urm8:
> The changes here as they stand don't consistently handle models with two references to the same related model, if also fetching a relation from each of those. The proposed fix only assigns `None` to the first instance in `instances` (which is just `instance` anyway?) which doesn't seem quite right, but neither does assigning all instances of the same type as is done currently for non-null relations. As an example:
> 
> ```python
> class Extra(Model):
>     id = IntField(pk=True)
> 
> class Single(Model):
>     id = IntField(pk=True)
>     extra = ForeignKeyField("db.Extra", related_name="singles", null=True)
> 
> class Pair(Model):
>     id = IntField(pk=True)
>     left = ForeignKeyField("db.Single", related_name="lefts", null=True)
>     right = ForeignKeyField("db.Single", related_name="rights", null=True)
> ```
> 
> ```python
> >>> extra = await Extra.create()
> >>> single = await Single.create(extra=extra)
> >>> await Pair.create(right=single)
> <Pair: 1>
> >>> pair = await Pair.filter(id=1).select_related("left", "left__extra", "right", "right__extra").get()
> >>> pair.left is None
> True
> >>> pair.right.extra  # should be <Extra: 1>
> <tortoise.queryset.QuerySet object at ...>
> 
> >>> single = await Single.create()
> >>> await Pair.create(right=single)
> <Pair: 2>
> >>> pair = await Pair.filter(id=2).select_related("left", "left__extra", "right", "right__extra").get()
> >>> pair.right.extra  # should be None
> <tortoise.fields.relational._NoneAwaitable object at ...>
> ```
> 
> More test cases

updated pr a bit
I see what you talking about, but I think that there is not enough info provided in select_related_idx in first place :) . And I'm not sure that this dumb function is responsible for finding/restoring paths as you did in your [commit](https://github.com/tortoise/tortoise-orm/commit/7ba269b56acf7e260df8534d27784c29eda7c8f8), but this is definitely an issue :(

long2ice:
So what about add those test cases?

urm8:
> So what about add those test cases?

going to add them

Terrance:
> I think that there is not enough info provided in select_related_idx in first place

Indeed, given the `Extra`/`Single`/`Pair` example from earlier, the only way to tell the difference between `pair.left.extra` and `pair.right.extra` is with the `index` -- they both have the same model, model name and parent name.  Checking the key list with that index tells you the path is different because the database column names have different table aliases.

> I'm not sure that this dumb function is responsible for finding/restoring paths

I agree -- I'm mainly raising this here for the benefit of the maintainers.  🙂

I think there's a bigger issue here around the ambiguity of relations in that relation list -- I'd like there to be a well-formed attribute path tuple in there, indepedent of the database aliases, such that e.g. `("left", "extra")` would allow you to walk from the root `Pair` instance to where the current related object should slot in.

urm8:
> So I also encountered this a few days ago, in the process of migrating from `prefetch_related()` to `select_related()` where applicable. I had been playing with a patch of my own, which can currently be found at [7ba269b](https://github.com/tortoise/tortoise-orm/commit/7ba269b56acf7e260df8534d27784c29eda7c8f8).
> 
> I believe this to have the desired behaviour, and also fixes what looks like a potential issue with a model that has two foreign keys for the same related model by explicitly looking for the parent model. The downside is it relies on the format of the column names, which is probably not desirable.
> 
> I'm happy to open a separate PR if it's worth pursuing this route, though it feels like this isn't the right place to be parsing the structure like this.

Not sure, what your test case did, but anyway added it to relation testcases, also fixed that stuff with select_related.
@long2ice there is a bug, you can't save models only having pk declared :(

Terrance:
Ah, you beat me to it -- [I had another go at refactoring it with the attribute path](https://github.com/Terrance-forks/tortoise-orm/compare/develop...select-related-null), but it looks like the change here works about the same (and the test case I added also passes with your change). 🙂

urm8:
@long2ice can you merge it pls ?

long2ice:
Hello, since version 0.17.6 is released, could you merge that first and update version to 0.17.7, then move it to section 0.17.7 in changelog? And are all tests passed? Don't know why this pr no ci performed

CHANGELOG.rst
```diff
@@ -15,6 +15,7 @@ Changelog
 - Fix columns count with annotations in `_make_query`. (#776)
 - Make functions nested. (#828)
 - Add `db_constraint` in field describe.
+- Fix `select_related` behaviour for forward relation. (#825)
```
long2ice, line 18: Should move this to 0.17.7

long2ice (approved):

CHANGELOG.rst
urm8: done

urm8:
> Hello, since version 0.17.6 is released, could you merge that first and update version to 0.17.7, then move it to section 0.17.7 in changelog? And are all tests passed? Don't know why this pr no ci performed


Hi ! Updated chlog. I ran tests locally, testall passed, style passed, ci passed

long2ice:
Thanks!

PR #830, opened 2021-07-21
WIP: Add value type hints to built-in fields

Terrance:
## Description

This change makes `Field` a generic class, and implements stub `__get__` and `__set__` methods such that the field type will match the value it's replaced with, when accessed via an instance.

## Motivation and Context

As the `Field` class isn't used as a descriptor (values are applied directly to instances of models), type checkers see attributes typed as the field instances themselves, rather than the types they represent.

## How Has This Been Tested?

Functional tests continue to pass (no change to runtime usage is expected).

Static typing tests need some work -- currently there are plenty of type errors.  The types of attributes on models receiving a relation object itself can be set to the actual return type `ForeignKeyFieldInstance[TModel]` rather than `ForeignKeyRelation[TModel]` (which only needs declaring in order to set the generic's type; by default it'd get `ForeignKeyRelation[Any]`).  `Optional[TModel]` is invalid for nullable fields though which may need some extra thought.

Any suggestions on how to approach the necessary test work would be appreciated.

## Checklist

### DIFF
--- a/tortoise/fields/base.py
+++ b/tortoise/fields/base.py
@@ -1,5 +1,18 @@
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Generic,
+    List,
+    Optional,
+    Tuple,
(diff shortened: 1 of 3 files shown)


waketzheng:
Good job! Now `Field` is generic class as your description.

PR #853, opened 2021-08-04
fix(tortoise): AttributeError: 'dict' object has no attribute '_meta'

SaidBySolo:
## Description
This PR allows us to get the correct model when we run the _reset_apps function.

## Motivation and Context

I found an unknown error while testing with [module](https://github.com/sanic-org/sanic-testing/tree/main/sanic_testing) ``AttributeError: 'dict' object has no attribute '_meta'``
image

So, as a result of checking the print statement, it was confirmed that an unknown value was entered.
image

The picture is the output value of the model.

As you can see, you can see that the unknown is included.

The first time the test is performed normally, but after that, it seems that the normal value does not enter.

To fix this, I used the isinstance function.

## How Has This Been Tested?
OS: Windows 10
Driver: aiomysql==0.0.21
Version: tortoise-orm==0.17.6

## Checklist:

### DIFF
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -642,7 +642,7 @@ class Tortoise:
     async def _reset_apps(cls) -> None:
         for app in cls.apps.values():
             for model in app.values():
-                if not isinstance(model, ModelMeta):
+                if isinstance(model, ModelMeta):
                     model._meta.default_connection = None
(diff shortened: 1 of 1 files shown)


long2ice (approved):

long2ice:
Thanks!

Issue #828, opened 2021-07-21
Functions can't be nested


Terrance:
## Describe the bug

Functions can't be applied to the result of another function.

## To reproduce

```python
from pypika import functions
from tortoise.expressions import F
from tortoise.fields import IntField
from tortoise.functions import Function
from tortoise.models import Model

class Abs(Function):
    database_func = functions.Abs

class Sqrt(Function):
    database_func = functions.Sqrt

class Test(Model):
    value = IntField()
```

```python
await Test.create(value=-4)
...
inst = await Test.get().annotate(abs_sqrt=Sqrt(Abs(F("value"))))
```

## Expected behavior

We should get `inst.abs_sqrt == 2`, but instead:

```
DEBUG:db_client:SELECT "id","value",SQRT(<__main__.Abs object at ...>) "abs_sqrt" FROM "test" LIMIT 2: None
...
tortoise.exceptions.OperationalError: near "<": syntax error
```

In theory a nested function can just be converted to SQL and embedded directly as its wrapper's argument.

PR #873, opened 2021-08-22
Add `Concat` function , also support nested

elisong:
Add `Concat` function , also support nested
## Description

Add `Concat` function, and slightly modified `tortoise.functions.Function` for better support nested Tortoise function in `Concat`.

Now you can:

```python
from tortoise.functions import Max, Concat, Coalesce, Trim

Book.all().annotate(
  long_info=Max(Concat("name", "(", Coalesce(Trim("subject"), "others"), ")"))
  ).values(
    "long_info"
  ).sql()

# SELECT MAX(CONCAT(`name`,'(',COALESCE(TRIM(`subject`),'others'),')')) `long_info` FROM `book`
#----if not modify  tortoise.functions.Function, result will be ----#
# SELECT MAX(CONCAT(`name`,'(',<tortoise.functions.Coalesce object at 0x10edf87d0>,')')) `long_info` FROM `book`
```
**Limitation**: `Concat` is subclass of `Tortoise.functions.Function`, like `Max`, `Upper`,.., it requires `Field` or str of Field name as first input argument, So `Concat("constant-str", F("field"), ...)` won't work.

## Motivation and Context

`Concat` is a common requirement, but not seen in Tortoise. I guess that because the api is different, it can feed in more than one `Field `. I'm not sure whether this pr will do harm to this rule. 

If this pr not be allowed, what is better work around for people do want to use `Concat`? 

have tried below, absolutely not work -_-

```python
from pypika import functions as fn
from tortoise.functions import Function
from tortoise.functions import Max

class Concat(Function):
    database_func = fn.Concat

Book.all().annotate(
  long_info=Max(Concat("name", "(", fn.Coalesce(fn.Trim("subject"), "others"), ")"))
  ).values(
    "long_info"
  ).sql()

# SELECT MAX(CONCAT(`name`,'(',COALESCE(TRIM('subject'),'others'),')')) `long_info` FROM `book`
```

Slight difference vs pr, it treat field "subject" as plain str.

related to #828  #829 

## How Has This Been Tested?

add tests

## Checklist:

### DIFF
--- a/tortoise/functions.py
+++ b/tortoise/functions.py
@@ -106,7 +106,7 @@ class Function:
     def _resolve_default_values(self, model: "Type[Model]", table: Table) -> Iterator[Any]:
         for default_value in self.default_values:
             if isinstance(default_value, Function):
-                yield default_value.resolve(model, table)['field']
+                yield default_value.resolve(model, table)["field"]
             else:
                 yield default_value
 
@@ -237,9 +237,10 @@ class Concat(Function):
     Concate field or constant text.
 
      :samp:`Concat("{FIELD_NAME}", {ANOTHER_FIELD_NAMES or CONSTANT_TEXT}, *args)`
-     """
+    """
 
     database_func = functions.Concat
+
 
 ##############################################################################
 # Aggregate functions
--- a/tortoise/functions.py
+++ b/tortoise/functions.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Optional, Type, Union, cast
+from typing import TYPE_CHECKING, Any, Iterator, Optional, Type, Union, cast
 
 from pypika import Case, Table, functions
 from pypika.functions import DistinctOptionFunction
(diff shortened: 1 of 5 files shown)


tortoise/functions.py
```diff
@@ -103,6 +103,13 @@ def _resolve_field_for_model(self, model: "Type[Model]", table: Table, field: st
 
         return {"joins": joins, "field": field}
 
+    def _resolve_default_values(self, model: "Type[Model]", table: Table) -> Iterator[Any]:
```
long2ice, line 106: What's for this?

long2ice (approved):

tortoise/functions.py
elisong: Iterate over default values if it meet a function.  For concat operation , it may happen to exist a function in positon argument later than first argument

elisong:
correct style format by use `make style`

elisong:
Update document for `Concat`

elisong:
DB like sqlite3 has no support for `CONCAT` (i mentioned in doc),  `make test_mysql` will ok

long2ice:
You can skip test by dialect

long2ice (approved):

long2ice:
OK, could your update changelog for that?

elisong:
My pleasure, update changelog under 0.17.7 ?

long2ice:
Thanks!

PR #882, opened 2021-08-26
Show detail of cyclic fk references

waketzheng:
We'd better show the table name and its references when raising ConfigurationError due to cyclic fk references.

### DIFF
--- a/tortoise/backends/base/schema_generator.py
+++ b/tortoise/backends/base/schema_generator.py
@@ -437,7 +437,12 @@ class BaseSchemaGenerator:
                     if t["references"].issubset(created_tables | {t["table"]})
                 )
             except StopIteration:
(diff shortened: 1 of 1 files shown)


caniko:
`t = tables_to_create[0]` lead to an index error for me.

tortoise/backends/base/schema_generator.py
```diff
@@ -437,7 +437,12 @@ def get_create_schema_sql(self, safe: bool = True) -> str:
                     if t["references"].issubset(created_tables | {t["table"]})
                 )
             except StopIteration:
-                raise ConfigurationError("Can't create schema due to cyclic fk references")
+                t = tables_to_create[0]
```
caniko, line 440: This raised an index error for me. Should probably do a try catch for this also, and what does empty `tables_to_create` mean in this context?

PR #896, opened 2021-09-02
Add Model.raw method to support the raw sql query

long2ice:
## Description

Add Model.raw method to support the raw sql query
## Motivation and Context

We can use that to write complex sql and return instance list
## How Has This Been Tested?

Add test
## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,11 @@ Changelog
 
 0.17
 ====
+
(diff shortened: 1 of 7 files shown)


s3rius:
Hi, @long2ice, I faced an issue and this feature would be really helpful. When can it be released?

long2ice:
soon

PR #897, opened 2021-09-02
Update _clone to work with QuerySet subclasses

lsapan:
## Description
Previously, any calls to `_clone` would result in the new QuerySet instance being initialized as a standard QuerySet. This updates `_clone` to use the active class instead. This is a tiny change.

## Motivation and Context
This allows the use of custom QuerySet classes which can have convenience methods for filtering, etc. Custom QuerySets are a [very helpful feature](https://docs.djangoproject.com/en/dev/topics/db/managers/) of Django.

## How Has This Been Tested?
I've tested the changes locally in my app. There shouldn't be any issues as this simply replaces the hardcoding of `QuerySet` with `self.__class__`. For a standard QuerySet, this will yield the same result. I haven't had time to get the test suite running yet.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Changelog
 0.17.8
 ------
 - Add `Model.raw` method to support the raw sql query.
+- Fix `QuerySet` subclass being lost when `_clone` is run on the instance.
 
 0.17.7
 ------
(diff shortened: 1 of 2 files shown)


long2ice (approved):

long2ice:
Thanks! Could you update changelog also?

lsapan:
No problem, thanks for an awesome library! And sure thing, just did. 👍

long2ice (approved):

long2ice:
Thanks!

Issue #919, opened 2021-09-21
Model.all().limit(None) fails


slavaGanzin:
Prior to latest release `Model.all().limit(None)` works as it should, but now it fails with
```
queryset.py 422 limit
if limit < 0:

TypeError:
'<' not supported between instances of 'NoneType' and 'int'
```
Besides, as stated in docs it should `Raises ParamsError – Limit should be non-negative number.`

I think it should be non-negative number or None.
Because it's convenient to have an optional limit on your queries. And it's very annoying to write some weird dances like:

```
query = Model.all()
if limit:
   query.limit(limit)
```

yoann9344:
I agree it could be convenient !
But for now you still can do this :
```python
query.limit(limit or 1000000)
```
And it has the advantage to be more explicit than [this](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/queryset.py#L442).

slavaGanzin:
@yoann9344 Thanks for the code. It's better than mine

Everything is even worse than I thought. 
I tried `limit(math.inf)` but it happens infinity is not a number in python. What a life!

yoann9344:
`import sys; sys.maxsize` is more appropriate in this case. infinites are for mathematics purpose, it doesn't really exist in your computer memory ! ;)

PR #922, opened 2021-09-22
WIP: Make querysets' methods limit and offset optional

yoann9344:
## Description
Now None can be pass to methods offset and limit :
```python
limit = None
offset = None
(
    queryset
    .limit(limit)
    .offset(offset)
)
```

As proposed in #919, it allows to always call a limit with a default value at None. Then no SQL request is made with maxsize limit, or to use an offset with a 0 value.

## Modifications
If their position parameter is None, then it simply return the queryset itself without cloning it. Considering that queryset has not changed we don't need to clone but it might be against rule of pure functional programming.

## How Has This Been Tested?
Not tested yet

## Checklist:

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -414,12 +414,15 @@ class QuerySet(AwaitableQuery[MODEL]):
         queryset._orderings = new_ordering
         return queryset
 
-    def limit(self, limit: int) -> "QuerySet[MODEL]":
+    def limit(self, limit: Optional[int]) -> "QuerySet[MODEL]":
         """
         Limits QuerySet to given length.
 
         :raises ParamsError: Limit should be non-negative number.
         """
+        if limit is None:
+            return self
+
         if limit < 0:
             raise ParamsError("Limit should be non-negative number")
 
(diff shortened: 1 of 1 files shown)


MojixCoder:
why it would be `None`? I think the developer should handle it and it shouldn't accept `None` values

long2ice:
I agree that

yoann9344:
It is not a big issue, just a matter a of coding style :
```python
queryset = (
    Model.filter(self.filters)
    .annotate(self.annotation)
    .prefetch(self.prefetch)
)

if self.limit is not None:
    queryset.limit(self.limit)
    if self.offset is not None:
        queryset.offset(self.offset)
```
Vs
```python
queryset = (
    Model.filter(self.filters)
    .annotate(self.annotation)
    .prefetch(self.prefetch)
    .limit(self.limit)
    .offset(self.offset)
)
```
I just prefer the the second option.
And I haven't dug enough yet the wheel behind the limit, so I don't really know if `queryset.limit(limit or 0)` could do the job ! In this case there is mostly no debate !

MojixCoder:
yeah but it's only one side.  
maybe someone really wants to use limit or offset but somehow he is passing `None` to the `.limit()` or `.offset()` methods.
If tortoise doesn't throw errors then it's too hard to debug why limit or offset is not working.  

That's why I think current approach is better and it's better that developer handles this situation

yoann9344:
Well seen ! That's makes sense !

Issue #928, opened 2021-09-27
Indexes defined in Meta class do not make use of the "exists" parameter in their template


dylankirby:
**Describe the bug**
User defined indexes do not append `IF NOT EXISTS` to their SQL statement.

This is due to the index templates not using the exists parameter in their templates.

Confirmed this issue exists whether or not you define a name for the index.

**Traceback**
`tortoise.exceptions.OperationalError: relation "idx_note_user_i_e5c588" already exists`

**To Reproduce**
Steps to reproduce the behavior, preferaby a small code snippet.
```
from tortoise import fields
from tortoise.models import Model
from tortoise.contrib.postgres.indexes import HashIndex

class Note(Model):
    id = fields.IntField(pk=True)
    note_id = fields.CharField(max_length=255, null=True)
    user_id = fields.CharField(max_length=255, null=True)
    user_name = fields.CharField(max_length=255, null=True)
    text = fields.TextField(null=True)

    class Meta:
        table = "note"
        unique_together = (
            (
                "user_id",
                "note_id",
            ),
        )
        indexes = [
            HashIndex(fields=("user_name",))
        ]
```

**Expected behavior**

User created indexes would make use of the `exists` parameter in their template definition, and thus not throw and exception on subsequent schema generation

**Additional context**

I've opened a PR to fix this

PR #929, opened 2021-09-27
Support `safe` parameter during index creation by including `exists` in index templates

dylankirby:
Description of change:
Add exists to Index templates
Add tests for safe/unsafe index creation

Notes:

Title says it all, current index templates don't support `exists` parameter despite it being passed to the format call.

This PR allows for safe index creation, ensuring that subsequent `generate_schemas` calls don't raise an exception

## Description

Not much more detail to be added.

## Motivation and Context

During a migration of my existing postgres code to Tortoise I noticed that user defined indexes in the meta class would raise exceptions on subsequent application startups.

Looking into `tortoise.backends.base.schema_generator:309-319` I noticed that a different conditional branch creates the SQL for user defined indexes, namely it calls `get_sql` function on the Index class, which does in fact pass `exists` to `format` but `exists` isn't used anywhere in either the base Index template, or the PostgreSQL Index template

https://github.com/tortoise/tortoise-orm/issues/928

## How Has This Been Tested?

Tested this manually by verifying that before the fix, calling `generate_schemas` twice for the model in the ticket would raise an exception.

After the fix this does not occur.

Intel Mac
Added unit tests and verified after altering Index templates
Used the defined model in the [ticket](https://github.com/tortoise/tortoise-orm/issues/928) and ran a small script that would connect to my local postgres and call `generate_schemas`

I see no downstream effects

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,7 @@ Fixed
 - Fix `QuerySet` subclass being lost when `_clone` is run on the instance.
 - Fix bug in `.values` with `source_field`. (#844)
 - Fix `contrib.blacksheep` exception handlers, use builtin json response. (#914)
+- Fix Indexes defined in Meta class do not make use of `exists` parameter in their template (#928)
 Changed
 ^^^^^^^
 - Allow negative values with `IntEnumField`. (#889)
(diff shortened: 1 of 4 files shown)


long2ice:
Thanks! Could you update changelog?

dylankirby:
@long2ice Updated the changelog!

long2ice:
Thank!

PR #934, opened 2021-09-30
Add black toml config

yoann9344:
Add the black config used by `poetry style`  in __pyproject.toml__


long2ice:
Can set it in setup.cfg like other tools?

yoann9344:
As far as I know black uses only pyproject.toml

long2ice:
OK, could you remove repeated opts of `black` in `Makefile`?

PR #935, opened 2021-09-30
WIP : Model Meta attribute method to apply to a field

yoann9344:
A way to convert a field before each SQL request. I have not found any other way to reach this behavior.

The commit allow to do :
```python
class Plop(Model):
    lastname = CharField()
    x = IntegerField()
    y = IntegerField()

    class Meta:
            transformers = {
                'lastname': str.upper,
                'x', lambda x: x + 10,  # position_with_offset
            }

Plop.get(lastname='plop')  # will filter all the lastname='PLOP'
Plop.filter(x=0, y=0)  # will filter all the x=10 and y=0
Plop.filter(Q(x=0) || Q(y=0))  # will filter all the x=10 and y=0
plop = Plop(...)
plop.lastname = 'plop'  # Now lastname should be 'PLOP' in database
```

It should work fine with all the cases above. I think it will require to add many tests !

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -239,6 +239,19 @@ The ``Meta`` class
 
             manager = CustomManager()
 
+    .. attribute:: transformers
+        :annotation: = dict[str, collections.abc.Callable[[Any], Any]]
+
+        Specify ``transformers`` to force all queries to be converted with the related Callable.
+        It should be a dictionary, as key a field, as value the Callable to receive and return the converted field's value.
+
+        .. code-block:: python3
+
+            transformers = {
+                'lastname': str.upper,
+                'x', lambda x: x + 10,  # position_with_offset
+            }
+
 ``ForeignKeyField``
 -------------------
 
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -206,10 +206,10 @@ class MetaInfo:
     )
 
     def __init__(self, meta: "Model.Meta") -> None:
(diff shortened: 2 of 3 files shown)


long2ice:
What about custom field and override `to_db_value` method?

yoann9344:
I don't achieve to make it works with `to_db_value`.
I have tried this :
```python
def test_convert(*args, **kwargs):
    print("CONVERT", args, kwargs)


class Town(Model):
    id = fields.IntField(pk=True)
    # name = fields.TextField(null=True, source_field="Commune", to_db_value=test_convert)
    name = fields.TextField(null=True, source_field="Commune", to_db_value="LILLE")

ModelRelationWithTown(town__name="Lille")
```
Neither hardcoded changes something and the method is not even called. Maybe I'm doing something wrong :/ ? And this query works smooth with this patch.

long2ice:
```py
class MyField(Field):
    def to_db_value(self, value: Any, instance: "Union[Type[Model], Model]") -> Any:
           pass
```

yoann9344:
It works smooth thx !
```python
class UpperTextField(fields.TextField):
    def to_db_value(self, value: Any, instance: "Union[Type[Model], Model]") -> Any:
        return value.upper()
```

But it is not really convenient ! I auto-generated the models on a sql file. So diff between the versions will become more difficult if I change the fields. I will keep my monkey patch for now. Do you think there is a risk to use my current method (`MetaInfo.transformers`) ?

PR #943, opened 2021-10-08
Add Case-When support

elisong:
Add Case-When support

## Description

Use  `Case(When(condition1,..., conditionN, then=...), When(...), default=...)` for building classic `CASE WHEN ... THEN ... ELSE ... END` sql snippet.

-  **condition** can be Q object or keyword filter,   like input in `Model.filter()`, `Model.exclude()`

## Motivation and Context

Using "CASE-WHEN-ELSE-END" sql snippet to derive new field is a common practice. This PR makes it more convenient/readable to achieve it than using `RawSQL`.

## How Has This Been Tested?

Please refer to test cases in `tests.test_case_when.py` for instance.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Added
 - Add `Rand`/`Random` function in contrib. (#944)
 Changed
 ^^^^^^^
-- Remove `Function`,`Aggregate` from 'functions.py' to 'expressions.py'. (#944)
+- Remove `Function`,`Aggregate` from 'functions.py' to 'expressions.py'. (#943)
 
 0.17.8
 ------


long2ice:
What aboue move them to `expressions.py`? I think they are not functions. Btw, please resolve conflicts.

elisong:
Agree, I have tried that. But if place it in `expressions.py`, it cause Circular (or cyclic) imports (`functions.py` <-> `expressions.py`).

Can it be considered as 'Generic' function ?

long2ice:
OK...Can then have same parent class? Or subclass of `Function`?

elisong:
「the former expects When instance., but the latter expects field: Union[str, F, ArithmeticExpression, "Function"] as first argument.」 If `When` inherits from 'Function', it seems not a problem......

long2ice:
That's real a problem, in django, they are all expressions

elisong:
Done

PR #948, opened 2021-10-10
Update base.py

nikolas-dev:
Added blank=True/False for required field

## Checklist:

### DIFF
--- a/tortoise/fields/base.py
+++ b/tortoise/fields/base.py
@@ -134,6 +134,7 @@ class Field(metaclass=_FieldMeta):
         generated: bool = False,
         pk: bool = False,
         null: bool = False,
+        blank: bool = False,
         default: Any = None,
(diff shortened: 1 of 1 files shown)


elisong:
`def validate(self, value)` method only check validators user provided. As for 'required' checking, either add logic in validator alone or just call `self.required` ?

long2ice:
Yes, don't need another param, and old logic has a problem, maybe we need change to this:
```py
def validate(self, value: Any):
    if self.null and value is None:
        return
    if self.required and value is None:
        raise ValidationError(f"Field {self.model_field_name} can't  be None ")
    for v in self.validators:
        try:
            if isinstance(value, Enum):
                v(value.value)
            else:
                v(value)
        except ValidationError as exc:
            raise ValidationError(f"{self.model_field_name}: {exc}")
```

PR #964, opened 2021-10-28
fix spelling mistakes

chr233:
I fixed a spelling mistakes in the docs ：`models => modules`

## Checklist:

- [*] I have updated the documentation accordingly.


abondar:
@chr233 Hi!

In your screenshots mistake you get is because you use kwarg `models` which is non-existent
Fix you are making to the docs is not relevant to issue, because you changing module name which could be any valid string

What you need to change to make this PR beneficial - is to change docstring in `tortoise/__init__.py`

PR #1055, opened 2022-01-21
Fix for None values in prefetch when fk_id==0

konchunas:
## Description
This PR fixes a bug with `prefetch_related` not actually being prefetched for entries with `fk_id==0`.
It happened because `getattr` returned `0` which is interpreted as `False` for `if` condition check.

## Motivation and Context
I have a table where foreign key id = 0 is a valid value.
Somehow `prefetch_related` always sets `fk_id` to `None` and  `fk` to `NoneAwaitable` for such records.

## How Has This Been Tested?
My project runs on Postgres and It works well. It doesn't break anything else.
I've made a simplest test for this case.
The change doesn't affect other tests, but I only ran `make test`

## Checklist:
- [ ] My code follows the code style of this project. (NOT SURE)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,12 @@ Changelog
 
 0.18
 ====
+0.18.3
+------
+Fixed
+^^^^^
+- Fix for prefetch_related with foreign key id == 0.
+
 0.18.2
 ------
 Fixed
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -511,7 +511,7 @@ class BaseExecutor:
         related_objects_for_fetch: Dict[str, list] = {}
         relation_key_field = f"{field}_id"
         for instance in instance_list:
-            if getattr(instance, relation_key_field):
+            if getattr(instance, relation_key_field) != None:
                 key = cast(RelationalField, instance._meta.fields_map[field]).to_field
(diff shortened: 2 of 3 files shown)


tortoise/backends/base/executor.py
```diff
@@ -511,7 +511,7 @@ async def _prefetch_direct_relation(
         related_objects_for_fetch: Dict[str, list] = {}
         relation_key_field = f"{field}_id"
         for instance in instance_list:
-            if getattr(instance, relation_key_field):
+            if getattr(instance, relation_key_field) != None:
```
insspb, line 514:
```suggestion
            if getattr(instance, relation_key_field) is not None:
```

insspb (changes requested):

CHANGELOG.rst
```diff
@@ -8,6 +8,12 @@ Changelog
 
 0.18
 ====
+0.18.3
```
long2ice, line 11: Please add in 0.19.2

long2ice (approved):

CHANGELOG.rst
konchunas: Done, thank you

abondar:
@konchunas Hi!

Sorry this PR got forgotten 

Could you please rebase on fresh develop and move it to new changelog section (0.21.4)?

PR #1084, opened 2022-03-12
Re-added support for __set_name__ as the Python class datamodel presc…

WoLpH:
The Python class datamodel normally supports the [__set_name__](https://docs.python.org/3/reference/datamodel.html#object.__set_name__) method on a class, allowing you to do:
```python
class A:
    x = C()  # Automatically calls: x.__set_name__(A, 'x')
```

This is a very useful feature for custom fields but the `Model` metaclass has removed this functionality. With this small patch it's re-added.

## Description

Add `__set_name__` support to the `Model` metaclass
Add a `__set_name__` type stub to `Field` because mypy doesn't support `__set_name__` yet: https://github.com/python/mypy/issues/8057
Added a test to confirm it works

## Motivation and Context

I'm building a native Postgres enum field, and having the name of the field available in the class makes this a lot easier.
Instead of doing this:
```python
class SomeModel(Model):
    some_enum = EnumField('some_enum')
```

I can now do:
```python
class SomeModel(Model):
    some_enum = EnumField()
```

## How Has This Been Tested?

I've added a simple test to confirm that the test works. And I've run `make test` and `make style ci`

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Changelog
 Added
 ^^^^^
 - Added psycopg backend support.
+- Added __set_name__ support for custom fields
 Fixed
 ^^^^^
 - Fix `bulk_create` doesn't work correctly with more than 1 update_fields. (#1046)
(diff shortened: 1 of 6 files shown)


WoLpH:
Since we're testing with pytest, it might be beneficial to allow `assert` statements within the tests at least :)
https://docs.pytest.org/en/latest/how-to/assert.html

PR #1129, opened 2022-05-14
Implement pydantic.from_tortoise_orm for lists of models

lsabi:
This PR provides a new function, similar to `from_tortoise_orm` but for a list of models instead of a single one.

This way, if there's some common logic that returns multiple instances (via a `list`) of a `Model`, they can be easily converted into a list of `Pydantic Models`.
## Description

Added a function that iterates over the list of Models passed as parameter, and calls `from_tortoise_orm` for each one. It then `gathers` all coroutines and returns the result.

Basically, it's a wrapper around `from_tortoise_orm`
## How Has This Been Tested?

I've created the test `test_address_list` in `tests/contrib/test_pydantic.py`
## Checklist:

### DIFF
--- a/tests/contrib/test_pydantic.py
+++ b/tests/contrib/test_pydantic.py
@@ -839,6 +839,40 @@ class TestPydantic(test.TestCase):
             },
         )
 
+    async def test_address_list(self):
+        addressp = await self.Address_Pydantic.from_tortoise_orm_list([await Address.get(street="Ocean")])
+        # print(addressp.json(indent=4))
+        addressdict = addressp[0].dict()
+
+        # Remove timestamps
+        del addressdict["event"]["tournament"]["created"]
+        del addressdict["event"]["modified"]
+
+        self.assertEqual(
+            addressdict,
+            {
+                "city": "Santa Monica",
+                "street": "Ocean",
+                "event": {
+                    "event_id": self.event.event_id,
+                    "name": "Test",
+                    "tournament": {
+                        "id": self.tournament.id,
+                        "name": "New Tournament",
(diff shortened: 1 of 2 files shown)


tortoise/contrib/pydantic/base.py
```diff
@@ -86,6 +87,33 @@ async def from_tortoise_orm(cls, obj: "Model") -> "PydanticModel":
...
+            This is due to how the ``asyncio`` framework forces I/O to happen in explicit ``await``
+            statements. Hence we can only do lazy-fetching during an awaited method.
+
+        :param objs: The List of Model instances you want serialized.
+        """
+        res = []
+        for obj in objs:
+            res.push(self.from_tortoise_orm(obj)
+        res = await asyncio.gather(*res)
+        return res
```
insspb, line 115: push? In what python version list objects have push?
Try this:
```suggestion
        return await asyncio.gather(*(self.from_tortoise_orm(obj) for obj in objs))
```

tests/contrib/test_pydantic.py
```diff
@@ -839,6 +839,40 @@ async def test_address(self):
             },
         )
 
+    async def test_address_list(self):
+        addressp = await self.Address_Pydantic.from_tortoise_orm_list([await Address.get(street="Ocean")])
+        # print(addressp.json(indent=4))
```
insspb, line 844:
```suggestion
```

insspb (changes requested):
Not using this project, and do not see tests in work. But I am sure, that `push` not exist for list objects, so it should fail on syntax check.

Issue #1155, opened 2022-06-10
Connection with Oracle --- ORA-01435: user does not exist (but it actually exists)


thakryptex:
### Describe the bug
Hi.
I'm trying to connect to existing Oracle database using `Tortoise ORM` and `asyncodbc`.
So I can connect to this DB using database tool in PyCharm or using `pyodbc` or `asyncodbc`. But when I run Tortoise code, I'm getting this error:

> {IntegrityError}('HY000', '[HY000] [Oracle][ODBC][Ora]ORA-01435: user does not exist\n\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x06\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x16\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00 (1435) (SQLExecDirectW)')

But user exists and, like I said, I can connect to DB and make queries using this user (and I see him in users table).
So it's a bug. I've tried to debug, but can't find it for now.

So, example like this is working:
```
dsn = 'DRIVER={ODBC Driver for Oracle};DBQ=host:port/database;UID=user;PWD=pass;'
conn = await asyncodbc.connect(dsn=dsn, loop=loop)
cur = await conn.cursor()
await cur.execute('SELECT * FROM schema.table FETCH NEXT 10 ROWS ONLY;')
rows = await cur.fetchall()
print(rows)
```

### To Reproduce
To reproduce it you have to have existing Oracle database, existing user in that database, schema named like username and some data there. And you need tortoise model to get data from that table.
DSN looks like this:
```
"connections": {
    "ibd": "oracle://user:password@host:port/database?driver={ODBC Driver for Oracle}"
}
```

And then code:
```
class OracleTable(Model):
    class Meta:
        table = "some_table_name"

    field1 = fields.CharField(max_length=100, pk=True)
    field2= fields.CharField(max_length=100, null=True)

async def run():
    await Tortoise.init(config_file='tortoise.json')
    res = await OracleTable.all().limit(10) # fails here (user not exist) if I ran this line using "Evaluate Expression" in PyCharm, but freezes forever if I just run code normally.

if __name__ == "__main__":
    run_async(run())
```

### Expected behavior
[...]

PR #1156, opened 2022-06-13
Fix work with Oracle and added new 'schema' attribute to Model's Meta

thakryptex:
## Description
- Updated the connection string for Oracle Client (DBQ key).
- Added new `schema` attribute to Model's Meta.
- Fixed ORA-01435 error while using Oracle database (#1155).
- Updated docs.

## Motivation and Context
- I had a problem, that I can't connect to existing Oracle database. It was throwing an error that database info is not provided or something like that. So I've checked how it was written and found that existing connection string for Oracle connection wasn't full. That lead me to the first fix: I updated the connection string for Oracle Client.
- Then I had a problem that I need to specify the schema that I'm working on (as for MsSQL and Oracle, because I have several DB). And I found in code that it was almost ready to set schema and properly paste it into queries. So few changes and that worked.
- Then I had a error from Oracle that user doesn't exists. I created an issue #1155. Then I finally found the reason of this message. And fixed it. It was putting database name into query, while schema name was needed. So I've passed user attribute instead of database attribute, because in Oracle database schema name is equal to a username.

## How Has This Been Tested?
I've run all of the existing tests and they all passed.
And I've tested this code in my project, that it do exactly what I wanted and what wasn't working before.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,11 +10,13 @@ Changelog
 ====
 0.19.2
 ------
+Added
+^^^^^
+- Added `schema` attribute to Model's Meta to specify exact schema to use with the model.
 Fixed
 ^^^^^
 - Mixin does not work. (#1133)
 - `using_db` wrong position in model shortcut methods. (#1150)
-- Added `schema` attribute to Model's Meta to specify exact schema to use with the model.
 - Fixed connection to `Oracle` database by adding database info to DBQ in connection string.
 - Fixed ORA-01435 error while using `Oracle` database (#1155)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ Fixed
 ^^^^^
 - Mixin does not work. (#1133)
 - `using_db` wrong position in model shortcut methods. (#1150)
+- Added `schema` attribute to Model's Meta to specify exact schema to use with the model.
+- Fixed connection to Oracle database by adding database info to DBQ in connection string.
 
 0.19.1
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,7 +15,8 @@ Fixed
 - Mixin does not work. (#1133)
 - `using_db` wrong position in model shortcut methods. (#1150)
 - Added `schema` attribute to Model's Meta to specify exact schema to use with the model.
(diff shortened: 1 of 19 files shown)


long2ice:
Thanks for your contribution! So what's the result when you run `make test_oracle` in local?

CHANGELOG.rst
```diff
@@ -14,6 +14,9 @@ Fixed
 ^^^^^
 - Mixin does not work. (#1133)
 - `using_db` wrong position in model shortcut methods. (#1150)
+- Added `schema` attribute to Model's Meta to specify exact schema to use with the model.
```
long2ice, line 17: You should format like before, just add `Added` section

long2ice (approved):

CHANGELOG.rst
thakryptex: ok, fixed

thakryptex:
> So what's the result when you run `make test_oracle` in local?

For now, I don't have any local Oracle databases, so tests are failing, because of that.

long2ice:
Do you have docker?

long2ice (approved):

thakryptex:
> Do you have docker?

Yes. Do you have configured Dockerfile for these tests?

long2ice:
Just run this: `docker run -d --network=host -e ORACLE_PASSWORD=123456 --name oracle --restart=always gvenzl/oracle-xe`

thakryptex:
> you run `make test_oracle` in local?

Did it work previously? Maybe I have setup problems with oracle or maybe it didn't work with oracle before, but all of the tests failed.
First errors was something about "You can insert null values into tournament_id" or something.
Then it was "ORA-00955: name is already used by an existing object", so database is not truncated after the tests (or because of test interruption).
So, I can't find the way to test it for now 😅

thakryptex:
@long2ice, wow, you're from Chongqing. I've been there. Great city, I like it 😊

long2ice:
Thanks! So how about this PR? Did oracle run success?

long2ice:
Not worked before, ok, now you just need make the ci pass

long2ice:
Looks like something error, I rerun it

thakryptex:
> now you just need make the ci pass

I did it :)

long2ice (approved):

PR #1163, opened 2022-06-23
Fixed: MySQL ssl option cast

capcom6:
Added cast of `ssl` option for MySQL backend.

## Description
Added `ssl` option to `cast` dictionary in `config_generator.py`.

## Motivation and Context
When using `ssl` option in MySQL connection string like `mysql://root:root@localhost:3306/db?ssl=True` it passed to driver as string. So we get `AttributeError: 'str' object has no attribute 'wrap_bio'` because driver expects `bool` or `SSLContext`.

After casting `ssl` option value to `bool` connection established successsfully.

## How Has This Been Tested?

Tested manually on Azure hosted MySQL server.

## Checklist:


long2ice (approved):

long2ice:
What about unittest and changlog?

long2ice (approved):

long2ice:
Thanks!

PR #1172, opened 2022-07-04
fix: use correct type hinting for QuerySetSingle

### LINKED ISSUE #701: Type annotation on QuerySetSingle
**Describe the bug**
When I try to use after a `Model.get(...).prefetch_related(...)` PyRight complains about not resolving the `MODEL` TypeVar.

Looking into the source code, I noticed that [QuerySetSingle](https://github.com/tortoise/tortoise-orm/blob/ea3a6f36e36add39800afce34ff154b74b2978c1/tortoise/queryset.py#L57) is annotated with both `T_co = TypeVar("T_co", covariant=True)` and `MODEL = TypeVar("MODEL", bound="Model")`.

It seems that the `MODEL` is not bound to any type in the annotations of a few methods (prefetch_related, annotate, only, ...).

Either changing all to `MODEL` or `T_co` seems to fix PyRight complaints.

Astrea49:
## Description
This PR makes `QuerySetSingle`'s functions use `T_co` fully (like the class), instead of using `MODEL` for no reason.

## Motivation and Context
Honestly? The major reason why I did this is because the type hinting annoyed me when I was trying to use `prefetch_related` with a `QuerySetSingle`. Instead of returning the proper `QuerySetSingle` for the object, it returned, literally, `QuerySetSingle[MODEL]`, which wasn't all that helpful. I saw that this was a bug other people had too, and decided to fix it just because it seemed simple enough.

After doing this, `prefetch_related` does work as expected with `QuerySetSingle`.

Fixes #701.

## Checklist:

- [x] All new and existing tests passed. (Besides for the ones that were failing before this change.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Fixed
 - Fixed connection to `Oracle` database by adding database info to DBQ in connection string.
 - Fixed ORA-01435 error while using `Oracle` database (#1155)
 - Fixed processing of `ssl` option in MySQL connection string.
+- Fixed type hinting for `QuerySetSingle`.
 
 0.19.1
 ------
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -52,16 +52,15 @@ if TYPE_CHECKING:  # pragma: nocoverage
     from tortoise.models import Model
 
 MODEL = TypeVar("MODEL", bound="Model")
-T_co = TypeVar("T_co", covariant=True)
 
 
-class QuerySetSingle(Protocol[T_co]):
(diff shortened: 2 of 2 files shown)


Astrea49:
Seems like the testing picked up some things that my machine didn't 😅.

I think I might have to bring back `T_co`, which is unfortunate but fair enough.

Astrea49:
There we go - all fixed! I ended up using Actions in order to run the test suite (was too lazy to set up WSL), and due to me not realizing how Actions work in forked repositories, I had to force push the same commit to trigger the action. Hopefully that isn't a *big* deal.

The title and description of this PR were also updated.

long2ice:
Thanks!

PR #1185, opened 2022-07-21
Make error message more clear when app not found in cls.apps

deajan:
When initializing tortoise, we may get `tortoise.exceptions.ConfigurationError: No app with name 'somename' registered.` if 'somename' can't be resolved.
This improves the error message to give a hint where the error might be.

## Description
Just a longer string with more explanation.

## Motivation and Context
To be fair, I've struggled about an hour to find why tortoise would fail on `init()` with the above message. I was first thinking my file tree was somehow wrong and tortoise couldn't find my models.
In the end I've found that I made a typo in a ForeignKeyField in one of my model classes. The original error message isn't really helpful since it mentions that `no app` with a given name is registered, so it points to the config dict where apps are defined.

This is just an improvement to the error message so newcommers like me will get more hints.

## How Has This Been Tested?
Obviously, changing the error message won't affect anything ;)

## Checklist:

Thanks for considering this PR.

### DIFF
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -119,7 +119,9 @@ class Tortoise:
                 return cls.apps[related_app_name][related_model_name]
             except KeyError:
                 if related_app_name not in cls.apps:
-                    raise ConfigurationError(f"No app with name '{related_app_name}' registered.")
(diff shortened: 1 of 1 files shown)


long2ice (approved):

long2ice:
Thanks!

PR #1192, opened 2022-07-27
fix: include schema in query cache key

K900:
## Description
Include schema name in the query cache key. Also, change the key to a tuple to avoid unnecessary string formatting.

## Motivation and Context
Fixes random errors when accessing two tables with the same name in different schemas (e.g. "a.foo" and "b.foo").

## How Has This Been Tested?
Tested on an internal project. Unfortunately couldn't find a good way to unit test this, as the default tests only use SQLite, which doesn't really have schemas...

## Checklist:


K900:
The tests not passing doesn't seem to be related to this change...

long2ice (approved):

long2ice:
Thanks!

PR #1261, opened 2022-09-23
Fix run_async RuntimeError

mike1936:
## Description
`run_async` helper function raising `RuntimeError` sometimes when the event_loop is None.
My working case for this `RuntimeError` to occur is calling the `run_async` function inside a `prefect.flow`, which is just a wrapped function provided by another package named `prefect` for workflow orchestration.

## Motivation and Context
`RuntimeError` raised by `asyncio.get_event_loop` not properly dealt by `run_async` function.
The event_loop is newly generated if get None then set to asyncio's event_loop 

## How Has This Been Tested?
NA

## Checklist:

### DIFF
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -683,7 +683,11 @@ def run_async(coro: Coroutine) -> None:
 
         run_async(do_stuff())
     """
-    loop = asyncio.get_event_loop()
+    try:
+        loop = asyncio.get_event_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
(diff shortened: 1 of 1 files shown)


waketzheng:
How to reproduce your bug?

And how about change run_async to be as follows:
```py
def run_async(coro: Coroutine) -> None:
    async def runner():
        try:
            await coro
        finally:
            await connections.close_all(discard=True)

    asyncio.run(runner())
```

abondar:
If it is actual error please describe it in issue with steps to reproduce

PR #1291, opened 2022-11-16
fix: add support for decimals in max/min validators

### LINKED ISSUE #1290: MinValueValidator does not work on DecimalField
**Describe the bug**
MinValueValidator and MaxValueValidator are

**To Reproduce**
```
class Model:
    # The following will immediately be rejected
    example1 = fields.DecimalField(max_digits=12, decimal_places=3, validators=(MinValueValidator(Decimal("0.00")),))
    # The following will be rejected when the value is compared in the db
    example2 = fields.DecimalField(max_digits=12, decimal_places=3, validators=(MinValueValidator(0),))

**Expected behavior**
It should intuitively work.

**Additional context**

https://github.com/tortoise/tortoise-orm/blob/48ea2dfbff3abc0ffc1b8ed85653f6cdf08bd57c/tortoise/validators.py#L63-L78

The two isinstance() calls could be augmented with a `Decimal` and I don't immediately see anything wrong with it.
However, it seems to almost explicitly exclude Decimal, so I'm not sure. Maybe @MojtabaArezoomand knows more as this was done in 48ea2dfbff3abc0ffc1b8ed85653f6cdf08bd57c.

mojixcoder:
Fixes #1290.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Fixed
 - Fix ignoring zero value of limit. (#1270)
 - Fix ForeignKeyField is none when fk is integer 0. (#1274)
 - Fix limit ignore zero. (#1270)
+- Fix min/max value validators for decimal fields. (#1291)
 
 0.19.2
 ------


MojtabaArezoomand:
> Thanks! Also changelog is required

Done

PR #1305, opened 2022-12-17
Make queryset typehints more specific

Ovsyanka83:
## Checklist:

- [] My code follows the code style of this project.
- [] My change requires a change to the documentation.
- [] I have added the changelog accordingly.
- [] I have added tests to cover my changes.

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -17,13 +17,14 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    overload,
 )
 
 from pypika import JoinType, Order, Table
 from pypika.functions import Cast, Count
 from pypika.queries import QueryBuilder
 from pypika.terms import Case, Field, Term, ValueWrapper
-from typing_extensions import Protocol
+from typing_extensions import Literal, Protocol
 
 from tortoise.backends.base.client import BaseDBAsyncClient, Capabilities
 from tortoise.exceptions import (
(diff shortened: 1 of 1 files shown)


Ovsyanka83:
@long2ice  hi! I can't really add a reviewer so I hope it's okay if I just ping you here. These changes should not affect any tests and are essentially quality of life improvements.

Ovsyanka83:
@long2ice should be fine now.

tortoise/queryset.py
```diff
@@ -1596,13 +1608,23 @@ def _make_query(self) -> None:
...
+        ...
+
+    def __await__(self) -> Generator[Any, None, Union[List[Dict[str, Any]], Dict[str, Any]]]:
         if self._db is None:
             self._db = self._choose_db()  # type: ignore
         self._make_query()
         return self._execute().__await__()  # pylint: disable=E1101
 
-    async def __aiter__(self) -> AsyncIterator[dict]:
+    async def __aiter__(self: "ValuesQuery[Any]") -> AsyncIterator[dict[str, Any]]:
```
Ovsyanka83, line 1627: Seems like the following makes a bit more sense with regards to typehints and logic because it doesn't make much sense to use `__aiter__` when SINGLE is True because we would be iterating over keys of a dict. 

```suggestion
    async def __aiter__(self: "ValuesQuery[Literal[False]]") -> AsyncIterator[dict[str, Any]]:
```

PR #1316, opened 2023-01-05
Pydantic typing fix

ZakharovDenis:
## Description
Changed config type for pydantic model creator.

## Motivation and Context
I faced MRO issue when i tried to create type-correct config class for `pydantic_model_creator`.
Example
```
from tortoise import fields
from pydantic import BaseConfig


def to_upper(s: str) -> str:
    return s.upper()


class ConfigClass(BaseConfig):
    alias_generator = to_upper
    allow_population_by_field_name = True


class ModelClass(Model):
    name: str = fields.TextField()
    age: int = fields.IntField()


pydantic_model = pydantic_model_creator(
    ModelClass,
    name="get_model",
    config_class=ConfigClass,
)
```

If you try to run this code you will face MRO issue
```
Cannot create a consistent method resolution
order (MRO) for bases BaseConfig, Config
```

If `ConfigClass` does not inherits `BaseConfig` everything works fine, but typing is incorrect. So, this MR main aim is to fix some typing errors.

## How Has This Been Tested?
1. Changed some tests to make config classes inherit `BaseConfig`
2. To check code style and correct work used `make check` and `make test`.

## Checklist:


ZakharovDenis:
@long2ice , hello. I rebased my branch to current develop, so all check have passed. What I have to do next? This MR should not affect any workability, because it's basically typehinting improvements. Plus it fixes possible MRO exception. 
If this MR needs more explanation or improvements, I'm ready

PR #1336, opened 2023-02-12
Fix type hint for param fields in Index, allowing tuples of any length

iburadempa:


iburadempa:
The wrong annotation was introduced in 7898280cfa936b9360d22e5042e1fda3cccce43d. [Python doc on tuples](https://docs.python.org/3/library/typing.html#typing.Tuple).

PR #1366, opened 2023-04-04
Fix fetching related object with ID = 0

s-rakowski:
Currently, when trying to fetch a related object that has an integer primary key that equals to 0, the foreign key is set to `None` instead and the related object is not fetched.

## Description
Although it's uncommon, it's possible to have a row with integer primary key equal to 0 or even negative.
An application I am working at has a table with such a row, and another table relates to it, `prefetch_related` and `fetch_related` doesn't work correctly. Instead of fetching the related object, the foreign key field is set to `None`.
This pull request tries to fix that.

## Motivation and Context
This pull request attempts to fix the issue described above.

## How Has This Been Tested?
I've added a test in test_relations.py called `test_fetch_fk_zero`. It creates two Tournament objects with hard-coded IDs of 0 and 1, and then creates an Event for both of them. The test fails before my change, because the Tournament with ID=0 is not fetched when trying to use both `fetch_related` and `prefetch_related` on an Event.
The change doesn't seem to affect other tests, but because this is my first attempt to contribute to this project, it would be good for someone else to verify.

## Checklist:

### DIFF
--- a/tests/test_relations.py
+++ b/tests/test_relations.py
@@ -75,6 +75,24 @@ class TestRelations(test.TestCase):
         result = await Event.filter(pk=event.pk).values_list("event_id", "participants__name")
         self.assertEqual(len(result), 2)
 
+    async def test_fetch_fk_zero(self):
+        tournament_0 = await Tournament.create(name="tournament zero", id=0)
+        tournament_1 = await Tournament.create(name="tournament one", id=1)
+        await Event.create(name="event-one", tournament=tournament_1)
+        await Event.create(name="event-zero", tournament=tournament_0)
+
+        for name in ("event-one", "event-zero"):
+            e = await Event.get(name=name)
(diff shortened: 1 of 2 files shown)


s-rakowski:
Looks like my test fails with MySQL. I'll check what happens later.

s-rakowski:
It turns out there has already been a similar issue discussed in #1275, sorry.
I moved my tests to already existing test case `test_0_value_fk`.
This test case skips MySQL, however it should be possible to make it pass with MySQL. This is because by default MySQL treats ID=0 as a request to get next number from the sequence. The following queries should allow the test to pass, at least with the `tournaments` table:
```sql
SET sql_mode = 'NO_AUTO_VALUE_ON_ZERO';
ALTER TABLE tournament AUTO_INCREMENT = 0;
```
However, I'm not sure if there's a good way to conditionally add these to the test case.

PR #1374, opened 2023-04-17
Add type hints to AwaitableQuery children

Ovsyanka83:
## Description
* Made it possible to type hint the generic arg in `BulkUpdateQuery` and `BulkCreateQuery`
* Type hinted the generic arg in all usages of `BulkUpdateQuery` and `BulkCreateQuery`
* Replaced the usages of "Type[Model]" with "Self" in `Model` to simplify the hints a little

## Motivation and Context
Whenever you do `bulk_update` or `bulk_create`, your model type gets lost because it is not type hinted in the return type of any one of the methods, which makes type checking ineffective when further interacting with any of these objects.  

** Example **
The following code revealed `List[Model@__await__]` before the changes:

```python
reveal_type(await UniqueName.bulk_create([UniqueName() for _ in range(1000)]))
```

Now it reveals:
`List[UniqueName]` 

## How Has This Been Tested?
I checked whether they work on a real live project and played around with them in tortoise's tests

## Checklist:

### DIFF
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -58,6 +58,7 @@ from tortoise.queryset import (
     QuerySetSingle,
     RawSQLQuery,
 )
+from typing_extensions import Self
 from tortoise.router import router
 from tortoise.signals import Signals
 from tortoise.transactions import in_transaction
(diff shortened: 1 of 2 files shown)


Ovsyanka83:
@long2ice hi! Fixed a few more type hints here and there :)

mrparalon:
This is really cool! Any plans when it will be released?

PR #1457, opened 2023-08-21
Fix TestCase | only "models" works as app_label

esrehmki:
- in the current implementation only "models" works as app_label
    - there is no way to change that during run time as it's hard coded
    - the fix introduces another _APP_LABEL global variable that is set
       in the initializer

Enables testing of apps with other labels than "models" in test classes of tortoise/contrib/test
automagically from the TORTOISE_ORM config dict.

## Description
Following the design of the file, a new global variable is introduced that is defined during the
call of the initializer function. Later occurences of the app_label function  parameter in _setUpDB
member functions, that don't expose the app_label variable, are provided with _APP_LABEL as
default value.

## Motivation and Context
App labels other than 'models' were not correctly propagated to the test functions, as 'models'
was hardcoded as app_label in the _setUpDB member functions.

## How Has This Been Tested?
Existing tests of the package pass. 

## Checklist:

### DIFF
--- a/tortoise/contrib/test/__init__.py
+++ b/tortoise/contrib/test/__init__.py
@@ -43,6 +43,7 @@ _SELECTOR = None
 _LOOP: AbstractEventLoop = None  # type: ignore
 _MODULES: Iterable[Union[str, ModuleType]] = []
 _CONN_CONFIG: dict = {}
+_APP_LABEL = None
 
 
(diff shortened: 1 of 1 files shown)


waketzheng:
Looks good. @esrehmki Cloud you fix the conflicts?

PR #1465, opened 2023-08-25
Fix pydantic v2 pydantic_model_creator nullable field not optional

waketzheng:
Fix pydantic v2 pydantic_model_creator nullable field not optional(#1454)

## How Has This Been Tested?
1. Tests for examples/fastapi passed
2. Tests for examples/blacksheep not work, because it does not support pydantic V2

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Added
 
 Fixed
 ^^^^^
+- Fix pydantic v2 pydantic_model_creator nullable field not optional. (#1454)
 - Fix foreign key constraint not generated on MSSQL Server. (#1400)
 - Fix testcase error with python3.11 (#1308)
 
--- a/examples/blacksheep/models.py
+++ b/examples/blacksheep/models.py
@@ -1,5 +1,9 @@
+import tortoise
 from tortoise import fields, models
 from tortoise.contrib.pydantic import pydantic_model_creator
+
+if tortoise.__version__ >= "0.20":
+    raise RuntimeError("blacksheep not support pydantic V2, use tortoise-orm<0.20 instead!")
 
 
 class Users(models.Model):
--- a/examples/fastapi/main.py
+++ b/examples/fastapi/main.py
@@ -22,7 +22,7 @@ async def get_users():
 
 @app.post("/users", response_model=User_Pydantic)
 async def create_user(user: UserIn_Pydantic):
-    user_obj = await Users.create(**user.dict(exclude_unset=True))
+    user_obj = await Users.create(**user.model_dump(exclude_unset=True))
     return await User_Pydantic.from_tortoise_orm(user_obj)
 
 
@@ -33,7 +33,7 @@ async def get_user(user_id: int):
 
 @app.put("/user/{user_id}", response_model=User_Pydantic)
 async def update_user(user_id: int, user: UserIn_Pydantic):
-    await Users.filter(id=user_id).update(**user.dict(exclude_unset=True))
+    await Users.filter(id=user_id).update(**user.model_dump(exclude_unset=True))
     return await User_Pydantic.from_queryset_single(Users.get(id=user_id))
 
 
--- a/examples/pydantic/tutorial_1.py
+++ b/examples/pydantic/tutorial_1.py
@@ -5,7 +5,7 @@ Here we introduce:
 * Creating a Pydantic model from a Tortoise model
 * Docstrings & doc-comments are used
 * Evaluating the generated schema
-* Simple serialisation with both .dict() and .json()
+* Simple serialisation with both .model_dump() and .json()
 """
 from tortoise import Tortoise, fields, run_async
 from tortoise.contrib.pydantic import pydantic_model_creator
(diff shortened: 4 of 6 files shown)


plusiv:
This PR solve a lot of things... I think it should be taken as **hotfix** @long2ice .

plusiv:
I was reviewing the @waketzheng 's code. And, I humbly submit that in order to set as nullable `ForeignKeyFieldInstance` and `OneToOneFieldInstance`, the following portion of the code:

https://github.com/tortoise/tortoise-orm/blob/ba593c3888741122b832e19a45d4504a951fa7ae/tortoise/contrib/pydantic/creator.py#L352-L365

Should be set to:
```python
        if (
            field_type is relational.ForeignKeyFieldInstance
            or field_type is relational.OneToOneFieldInstance
            or field_type is relational.BackwardOneToOneRelation
        ):
            is_to_one_relation = True
            model = get_submodel(fdesc["python_type"])
            if model:
                if fdesc.get("nullable"):
                    json_schema_extra["nullable"] = True
                    # set default_factory to None in the field config
                    fconfig["default_factory"] = lambda: None
                if fdesc.get("nullable") or field_default is not None:
                    model = Optional[model]  # type: ignore

                # create property with that config
                properties[fname] = (model, Field(**fconfig))
```

waketzheng:
@plusiv After set default_factory to None in the field config, unittest failed at tests/contrib/test_pydantic.py

plusiv:
> ```python
> properties[fname] = (model, Field(**fconfig))
> ```

Hi again @waketzheng !

I've cloned your changes and tested with the changes that I suggested before, and figure out that the reason is because the hardcoded schema used in the test `assertion` has so many errors since the relational fields are not set in the proper way to make nullable relationships fields/objects in the schema.

To explain my point, let's take a look to the models `Event` and `Reporter`

https://github.com/tortoise/tortoise-orm/blob/c030c9cebd7aba2697832d9c4769428805238f73/tests/testmodels.py#L74-L113

Since the field `alias` of the model `Event` is set to be `null`, the schema looks like:
https://github.com/tortoise/tortoise-orm/blob/c030c9cebd7aba2697832d9c4769428805238f73/tests/contrib/test_pydantic.py#L191-L198

But, with the nullable `reporter` field of the same  `Event` model doesn't happen the same in its schema:

https://github.com/tortoise/tortoise-orm/blob/c030c9cebd7aba2697832d9c4769428805238f73/tests/contrib/test_pydantic.py#L84-L99

Notice that the element `{"type": "null"},` in the `anyOf` property list and the property `"nullable": True,` are missing in the schema definition.

I think that re-generating the hardcoded schema with the changes that I suggested before and replacing the old ones with the new ones, should make the test run without problems. I could help with this if you like.

plusiv:
Hello again @waketzheng !

I was reviewing the function `pydantic_model_creator`, and I'm wondering why did you just added `IntField` and `TextField` in:

https://github.com/waketzheng/tortoise-orm/blob/ba593c3888741122b832e19a45d4504a951fa7ae/tortoise/contrib/pydantic/creator.py#L415-L424

This is causing setting as **required** `IntField`s and `TextField`s marked as nullable.

Note: I'm just trying to understand the project logic in order to collaborate in a future.

waketzheng:
> > ```python
> > properties[fname] = (model, Field(**fconfig))
> > ```
> 
> Hi again @waketzheng !
> 
> I've cloned your changes and tested with the changes that I suggested before, and figure out that the reason is because the hardcoded schema used in the test `assertion` has so many errors since the relational fields are not set in the proper way to make nullable relationships fields/objects in the schema.
> 
> To explain my point, let's take a look to the models `Event` and `Reporter`
> 
> https://github.com/tortoise/tortoise-orm/blob/c030c9cebd7aba2697832d9c4769428805238f73/tests/testmodels.py#L74-L113
> 
> Since the field `alias` of the model `Event` is set to be `null`, the schema looks like:
> 
> https://github.com/tortoise/tortoise-orm/blob/c030c9cebd7aba2697832d9c4769428805238f73/tests/contrib/test_pydantic.py#L191-L198
> 
> But, with the nullable `reporter` field of the same `Event` model doesn't happen the same in its schema:
> 
> https://github.com/tortoise/tortoise-orm/blob/c030c9cebd7aba2697832d9c4769428805238f73/tests/contrib/test_pydantic.py#L84-L99
> 
> Notice that the element `{"type": "null"},` in the `anyOf` property list and the property `"nullable": True,` are missing in the schema definition.
> 
> I think that re-generating the hardcoded schema with the changes that I suggested before and replacing the old ones with the new ones, should make the test run without problems. I could help with this if you like.

@plusiv You are right! Unittest should be updated.

waketzheng:
> Hello again @waketzheng !
> 
> I was reviewing the function `pydantic_model_creator`, and I'm wondering why did you just added `IntField` and `TextField` in:
> 
> https://github.com/waketzheng/tortoise-orm/blob/ba593c3888741122b832e19a45d4504a951fa7ae/tortoise/contrib/pydantic/creator.py#L415-L424
> 
> This is causing setting as **required** `IntField`s and `TextField`s, even if marked as nullable.
> 
> Note: I'm just trying to understand the project logic in order to collaborate in a future.

This line is to respect to unittest and compare with old version logic, after tests code updated, if can be removed.

plusiv:
> This line is to respect to unittest and compare with old version logic, after tests code updated, if can be removed.

Good! Seems razonable now!


> @plusiv You are right! Unittest should be updated.

Got it, I think it can be updated on this PR, or you think a new one is necessary @long2ice ?? 

I am willing to assist if you agree.

plusiv:
Hey everyone, Did this PR got stuck?

waketzheng:
> Thanks! Just need resolve conflicts.

conflicts fixed, but need you to rerun ci action manually.

PR #1471, opened 2023-08-30
option to set validators of pydantic model

lucasbrahm:
Pydantic provides Before, After, Wrap and Plain validators, Field validators and Model validators. This is nice for cleaning and validating fields. For example, you can remove any non numeric characters for a phone field and then validate, like this:

`class Register(BaseModel):
    phone: str
    
    @field_validator('phone')
    @classmethod
    def validate_document(cls, v: str) -> str:
        new_phone = re.sub(r'\D', '', v)
        if len(new_phone) < 7:
            raise ValueError('Invalid phone number')
        return new_phone`

So both phone '123-4567' and '1234567' will be validated and it will be stored as '1234567'.

The problem of tortoise "pydantic_model_creator" is that it exposes few options when calling pydantic "create_model". Also, tortoise field validators only validate data, without modifying/cleaning data.

Fortunately, you can add validators in pyndatic by passing a dict to the "__validators__" argument when calling "create_model".

The idea of this PR is to add "__validators__" option in tortoise "pydantic_model_creator" which will be used during the call of pydantic "create_model"


tortoise/contrib/pydantic/creator.py
```diff
@@ -131,6 +131,7 @@ def pydantic_model_creator(
     exclude_readonly: bool = False,
     meta_override: Optional[Type] = None,
     model_config: Optional[ConfigDict] = None,
+    __validators__: dict[str, Any] | None = None,
```
plusiv, line 134: IMHO this should be received as `validators` (without the underscores at beginning and end).

lucasbrahm:
Made the adjustments. Also updated the docstring.
Not sure how to add it to the changelog

lucasbrahm:
Is it ok?

MMartin09:
Hi, `validators` parameter is already present in the documentation of version 0.20.0 (here)[https://tortoise.github.io/contrib/pydantic.html?h=pyda#tortoise.contrib.pydantic.creator.pydantic_model_creator]

This parameter is, however not present in version 0.20.0. I am receiving `TypeError: pydantic_model_creator() got an unexpected keyword argument 'validators'`. I also looked into the code of the target version and saw that it is not there 
https://github.com/tortoise/tortoise-orm/blob/0c81200bbaa3f407c5dce9e6ae34894899eef0dc/tortoise/contrib/pydantic/creator.py#L120C2-L120C28

Is there a way to reach this behaviour in version 0.20.0?

PR #1488, opened 2023-09-29
Minor type hint fixes

Ovsyanka83:
There isn't much to say here: the old type hints were invalid and these are correct.


Ovsyanka83:
@long2ice A few really small fixes. A variable was used instead of type hints and Tuple with a single element was used instead of `tuple[str, ...]`

PR #1493, opened 2023-10-10
Add __all__ to __init__.py for type checkers

Ovsyanka83:
When you generate migrations using aerich, it imports `BaseDBAsyncClient` directly from tortoise but tortoise doesn't export it which is a problem for some type checkers such as pyright. This fix resolves this. 

## How Has This Been Tested?
No need for testing

## Checklist:

- [X] I have updated the documentation accordingly.
- [X] I have added the changelog accordingly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,10 @@ Changelog
 0.20
 ====
 
+Added
+^^^^^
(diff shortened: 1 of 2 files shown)


Ovsyanka83:
@long2ice A gentle ping to take a look. Note that this can break some integrations which imported "*" from tortoise and used some of the internal stuff defined/imported in the `__init__.py` but it is extremely unlikely.

Ovsyanka83:
@long2ice done! Big thanks for a quick reply.

PR #1498, opened 2023-10-15
Feature: async default value

YAGregor:
allow Field's default value not only be callable, literal value, but also async function

## Description

when set Field's default keyword option to async function, then while save and create model, if the field never set, the field will be set as the async function's result

## Motivation and Context

I write a model and I want a serial of unique, auto increasing serial names generate from redis (by aioredis), but the field's default do not support retrieve value from async function, I have to override create method, that's cumbersome.

and tortoise is an async orm, it should support async first.

## How Has This Been Tested?

I create a new model for new test case, test model's behavior when combine async default with `save`, `create`, and when default field value was covered by set attribute.

I run tests on linux, include sqlite's test cases.

Almost no effect.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Added
 - Enhancement for FastAPI lifespan support (#1371)
 - Add __eq__ method to Q to more easily test dynamically-built queries (#1506)
 - Added PlainToTsQuery function for postgres (#1347)
+- Allow field's default keyword to be async function (#1498)
 
 Fixed
 ^^^^^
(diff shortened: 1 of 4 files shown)


YAGregor:
I've done the change log, do you think we should update the doc? Seems that this behavior is natural, and there is no need to @long2ice

abondar:
@YAGregor hi!

Please rebase PR on actual develop and resolve conflict in Changelog
Changes should go to 0.21.0 section

CHANGELOG.rst
```diff
@@ -1333,3 +1333,4 @@ Docs/examples:
       await Tournament.filter(
           events__name__in=['1', '3']
       ).order_by('-events__participants__name').distinct()
+* Allow field's default keyword to be async function (#1498)
```
abondar, line 1336: Seems like this entry is in wrong place

abondar (approved):

PR #1416, opened 2023-07-03
fix: check attr exists before setattr to aviod KeyError

ezioruan:
## Description
After the service has been running for a period of time, the tortoise orm will suddenly report an error and need to restart the service to repair it.

Additional context

Traceback (most recent call last):
File "/usr/local/lib/python3.8/dist-packages/tortoise/models.py", line 717, in _init_from_db
setattr(self, model_field, kwargs[key])
KeyError: 'images'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/data/frontend/api.py", line 3296, in viewRecord
scheme_obj = await Scheme.get_or_none(uni_id=scheme_id)
File "/usr/local/lib/python3.8/dist-packages/tortoise/queryset.py", line 908, in _execute
instance_list = await self._db.executor_class(
File "/usr/local/lib/python3.8/dist-packages/tortoise/backends/base/executor.py", line 134, in execute_select
instance: "Model" = self.model._init_from_db(
File "/usr/local/lib/python3.8/dist-packages/tortoise/models.py", line 734, in _init_from_db
setattr(self, key, meta.fields_map[key].to_python_value(value))
KeyError: 'watch_user_id'
Traceback (most recent call last):
File "/data/frontend/api.py", line 1446, in schemeList
s['sale_num'] = await Order.filter(scheme_id=s['uni_id'], status=1).count()
File "/usr/local/lib/python3.8/dist-packages/tortoise/queryset.py", line 1174, in _execute
count = list(dict(result[0]).values())[0] - self.offset
IndexError: list index out of range
Traceback (most recent call last):
File "/usr/local/lib/python3.8/dist-packages/tortoise/models.py", line 717, in _init_from_db
setattr(self, model_field, kwargs[key])
KeyError: 'nickname'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/data/frontend/api.py", line 474, in getSubscribe
wx_user_obj = await WXUserInfo.get_or_none(user_id=request.ctx.uid, app_id=ACTIVE_WX_APP_ID)
File "/usr/local/lib/python3.8/dist-packages/tortoise/queryset.py", line 908, in _execute
instance_list = await self._db.executor_class(
File "/usr/local/lib/python3.8/dist-packages/tortoise/backends/base/executor.py", line 134, in execute_select
instance: "Model" = self.model._init_from_db(
File "/usr/local/lib/python3.8/dist-packages/tortoise/models.py", line 734, in _init_from_db
[...]

PR #1571, opened 2024-03-18
fix: check attr exists before setattr to aviod KeyError #1416

smartydroid:
## Checklist:


abondar:
Did you manage to understand root cause of issue? This way it seems it will just hide the problem, not fix it

Issue #1575, opened 2024-03-26
Field 'DatetimeField' has error when I use '__year' or '__month'. report this:" 'int' object has no attribute 'utcoffset' "


h2magic-axious:
**Describe the bug**
This is my model:
```python
class Transfer(AbstractModel):
    class Meta:
        table = "transfer"

    account = fields.ForeignKeyField("models.Account", related_name="transfers")
    category = fields.ForeignKeyField("models.Category")

    title = fields.CharField(max_length=30)
    amount = fields.DecimalField(max_digits=10, decimal_places=2)
    remark = fields.CharField(max_length=128, default="")
    created_at = fields.DatetimeField(auto_now_add=True)
```

**To Reproduce**
This is filter:
```python
@router.get("/test")
async def test_all_transfers():
    transfers = await Transfer.filter(created_at__year=2024)
    return response_success(transfers)
```

**Expected behavior**
I get follow:
```shell
AttributeError: 'int' object has no attribute 'utcoffset'
```

**Additional context**
this is full stack:
```
ERROR:    Exception in ASGI application
  + Exception Group Traceback (most recent call last):
  |   File "D:\coding\money-manage\venv\Lib\site-packages\starlette\_utils.py", line 87, in collapse_excgroups
  |     yield
  |   File "D:\coding\money-manage\venv\Lib\site-packages\starlette\middleware\base.py", line 190, in __call__
  |     async with anyio.create_task_group() as task_group:
  |   File "D:\coding\money-manage\venv\Lib\site-packages\anyio\_backends\_asyncio.py", line 678, in __aexit__
  |     raise BaseExceptionGroup(
  | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "D:\coding\money-manage\venv\Lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 407, in run_asgi
    |     result = await app(  # type: ignore[func-returns-value]
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "D:\coding\money-manage\venv\Lib\site-packages\uvicorn\middleware\proxy_headers.py", line 69, in __call__
    |     return await self.app(scope, receive, send)
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1580, opened 2024-04-07
fix: `DatetimeField` use __year report `'int' object has no attribute 'utcoffset'`

waketzheng:
Fix `DatetimeField` use '__year' report `'int' object has no attribute 'utcoffset'`. (#1575)
When config with `"use_tz": True`, the bug  can be reproduced.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,7 @@ Fixed
 - Fix pydantic v2.5 unittest error. (#1535)
 - Fix pydantic_model_creator `exclude_readonly` parameter not working.
 - Fix annotation propagation for non-filter queries. (#1590)
+- Fix `DatetimeField` use '__year' report `'int' object has no attribute 'utcoffset'`. (#1575)
 
 0.20.0
 ------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,5 +180,5 @@ source = ["tortoise"]
 [tool.coverage.report]
 show_missing = true
 
-[tool.ruff]
+[tool.ruff.lint]
 ignore = ["E501"]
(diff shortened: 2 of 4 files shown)


abondar:
Hello!

Please rebase on actual develop branch and add corresponding test, that will show that problem is gone

waketzheng:
The bug can be reproduce by:
```py
from datetime import datetime

class Event(Model):
    id = fields.IntField(pk=True)
    name = fields.TextField()
    created = fields.DatetimeField(null=True)

    class Meta:
        table = "event"

    def __str__(self):
        return self.name


async def run():
    # url = 'mysql://root:123456@127.0.0.1:3306/test_year'
    url = "asyncpg://postgres:postgres@127.0.0.1:5432/test_year"
    await Tortoise.init(db_url=url, modules={"models": ["__main__"]}, use_tz=True)

    event = await Event.create(name="Test", created=datetime.now())
    print(event.created)

    events = await Event.filter(created__year=2024)
    print(events)


if __name__ == "__main__":
    run_async(run())
```

abondar:
Please add such testcase

```    async def test_filter_by_year(self):
        with patch.dict(os.environ, {"USE_TZ": "True"}):
            obj = await testmodels.DatetimeFields.create(datetime=datetime(year=2024, month=1, day=1))
            filtered_obj = await testmodels.DatetimeFields.filter(datetime__year=2024).first()
            assert obj == filtered_obj```

to `tests.fields.test_time.TestDatetimeFields`

abondar:
Seems like mssql doesn't support extract syntax, and uses some kind of `CONVERT()` syntax instead
I think fixing it is out of scope for this PR, so you could just exclude mssql from this testcase

waketzheng:
@abondar done.

Issue #1583, opened 2024-04-10
model.update_or_create KeyError


jiangying000:
https://github.com/tortoise/tortoise-orm/blob/0c8efde78c4dae480b229e75790631297cab7fb3/tortoise/models.py#L1065

would cause KeyError for following code:

```
class Faqs(BaseModel):
    id = fields.UUIDField(pk=True)
    question = fields.TextField(null=True)

@pytest.mark.asyncio
async def test_tortoise_update_or_create():
    # suppose entry exists
    faq = await Faqs.get(id="6aadafff-e0b9-497b-aeb2-65498a35b6f5")
    # modify question
    faq.question = "q111234"
    # assign new id
    faq.id = uuid.uuid4()
    # update_or_create, KeyError
    new_faq = await Faqs.update_or_create(id=faq.id, defaults=dict(faq))


```
ERROR like:
```
    @classmethod
    async def get_or_create(
        cls: Type[MODEL],
        defaults: Optional[dict] = None,
        using_db: Optional[BaseDBAsyncClient] = None,
        **kwargs: Any,
    ) -> Tuple[MODEL, bool]:
        """
        Fetches the object if exists (filtering on the provided parameters),
        else creates an instance with any unspecified parameters as default values.
    
        :param defaults: Default values to be added to a created instance if it can't be fetched.
        :param using_db: Specific DB connection to use instead of default bound
        :param kwargs: Query parameters.
        :raises IntegrityError: If create failed
        :raises TransactionManagementError: If transaction error
        """
        if not defaults:
            defaults = {}
        db = using_db or cls._choose_db(True)
        async with in_transaction(connection_name=db.connection_name) as connection:
            try:
                return (
                    await cls.select_for_update()
                    .filter(**kwargs)
                    .using_db(connection)
                    .get(),
                    False,
                )
            except DoesNotExist:
                try:
                    return (
>                       await cls.create(using_db=connection, **defaults, **kwargs),
                        True,
                    )
E                   KeyError: 'id'

..\.venv\lib\site-packages\tortoise\models.py:1102: KeyError



```

waketzheng:
I got different errors with the following code snippet:
```py
import uuid

from tortoise import Tortoise, fields, run_async
from tortoise.models import Model


class Faqs(Model):
    id = fields.UUIDField(pk=True)
    question = fields.TextField(null=True)


async def run():
    await Tortoise.init(db_url="sqlite://:memory:", modules={"models": ["__main__"]})
[...]

PR #1584, opened 2024-04-10
fix: models get_or_create keyerror

jiangying000:
#1583


abondar:
Could you please rebase your PR on actual develop branch?

jiangying000:
> Could you please rebase your PR on actual develop branch? Also please add small test on it and info about change to changelog

yes, i did this

tortoise/models.py
waketzheng, line 1: 1. `kwargs` is 'Query parameters', `defaults` is 'Default values used to update the object', when they have same key with different values, value of `kwargs` should be covered by `defaults`. That is to say code must be: `merged_defaults = {**kwargs, **defaults}` 
2. merged_defaults = ... should be in front of `try:` (just put it after line 1082)
3. `return ...` shoud be stay as one line, do not make it into three lines.

jiangying000: thanks, done

abondar: Even though that is how it works in django - for me it seems strange that `default` value overrides explicit value

I would expect opposite behaviour of this

I dunno if we should proceed with this change, as leaving merging on user side seems like more transparent behaviour

jiangying000: i would also expect explicit kv overrides the default and do merge rather than throw err is ok for me

abondar: It seems to me - in your case you should do explicit `get_or_create` without `defaults` and them make usual `update()` for returned object
That shouldn't affect performance, as it is basically what we already do

Django also, from 5.0, introduced separate `create_defaults` param for `update_to_create` for similar cases to make them less confusing
May be we should look into doing it too, but it seems there could be some issues in tortoise to use `select_for_update` with `get_or_create`

waketzheng: > Even though that is how it works in django - for me it seems strange that `default` value overrides explicit value
> 
> I would expect opposite behaviour of this
> 
> I dunno if we should proceed with this change, as leaving merging on user side seems like more transparent behaviour

Our team use tortoise-orm for a long time, and we are proficient in django. So we prefer it have the same result with django.

Or maybe an exception can be raised when kwargs and defaults have same key with different values:
```py
for key in (defaults.keys() & kwargs.keys()):
    if (default_value := defaults[key]) != (query_value := kwargs[key]):
        raise tortoise.exceptions.ParamError(f'Conflict value with {key=}: {default_value=}  vs {query_value=}')
```

abondar: > Or maybe an exception can be raised when kwargs and defaults have same key with different values:

I would prefer that, as django behaviour seems confusing to me here, but doing solution opposite to them - would be also confusing

jiangying000: > > Or maybe an exception can be raised when kwargs and defaults have same key with different values:
> 
> I would prefer that, as django behaviour seems confusing to me here, but doing solution opposite to them - would be also confusing

ok, updated

tests/test_model_methods.py
waketzheng, line 1: Should add case about `kwargs` and `defaults` have same key and same value

tortoise/models.py
```diff
@@ -23,6 +23,7 @@
 from pypika.terms import Term
 from typing_extensions import Self
 
+import tortoise
```
waketzheng, line 26: Remove this line

tortoise/models.py
```diff
@@ -1073,17 +1073,23 @@ async def get_or_create(
...
+        except DoesNotExist:
             try:
-                return (
-                    await cls.select_for_update().filter(**kwargs).using_db(connection).get(),
-                    False,
-                )
-            except DoesNotExist:
+                for key in (defaults.keys() & kwargs.keys()):
+                    if (default_value := defaults[key]) != (query_value := kwargs[key]):
+                        raise tortoise.exceptions.ParamsError(
```
waketzheng, line 1082: Change it to:
```py
raise ParamsError("...")
```
Then run `make style` to reformat this file

tortoise/models.py
waketzheng, line 1: image

waketzheng:
@jiangying000 How is that?

jiangying000:
> @jiangying000 How is that?

sorry, no progress yet, too busy dealing with my employment job tasks

jiangying000:
feel free to take on this pr, or close it

waketzheng:
> feel free to take on this pr, or close it

I have make a pr, could you merge it into your develop branch?
https://github.com/jiangying000/tortoise-orm/pull/1

jiangying000:
> > feel free to take on this pr, or close it
> 
> I have make a pr, could you merge it into your develop branch? [jiangying000#1](https://github.com/jiangying000/tortoise-orm/pull/1)

done, thanks

jiangying000:
> > feel free to take on this pr, or close it
> 
> I have make a pr, could you merge it into your develop branch? [jiangying000#1](https://github.com/jiangying000/tortoise-orm/pull/1)

done, thanks

waketzheng (approved):

waketzheng:
@abondar Could you review this?

abondar:
@waketzheng seems like there is some issue 
tests hangs up somewhere on ODBC tests

Also we would need to update changelog to put change in new section

waketzheng:
> @waketzheng seems like there is some issue tests hangs up somewhere on ODBC tests
> 
> Also we would need to update changelog to put change in new section

@abondar done

waketzheng:
@abondar Cloud you review it?

abondar (approved):

PR #1594, opened 2024-04-25
fix `exclude_readonly` param not working in `pydantic_model_creator`

Abdeldjalil-H:
The `exclude_readonly` paramter in `pydantic_model_creator` is not working, i.e it is ignored. `readOnly` is deleted from `fdesc["constraints"]` so instead we read it from `extra_json_schema`.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,7 +22,7 @@ Fixed
 - Fix order of fields in `ValuesListQuery` when it has more than 10 fields. (#1492)
 - Fix pydantic v2 pydantic_model_creator nullable field not optional. (#1454)
 - Fix pydantic v2.5 unittest error. (#1535)
-- Fix pydantic_model_creator `exclude` parameter not working.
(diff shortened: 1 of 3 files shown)


abondar:
Hi, thanks for PR!
Could you please add some small test, that shows that changed behaviour now works?

Abdeldjalil-H:
I have added a test case @abondar.

abondar:
Seems like that test passes also on previous version, that doesn't have fix

I assume it is because we already exclude `id` somewhere here
https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/contrib/pydantic/creator.py#L272

Did you originally face problem on some more complex case? Are you able to reproduce it using test models?

Abdeldjalil-H:
Yes, I have faced the issue with some models that have `created_at` which is `auto_now` field. I have changed the test case to `modified` on `Event` model. By the way, this was working on previous version (with pydantic 1).

abondar (approved):

Issue #1566, opened 2024-03-11
mssql named instance port number


adi1:
**Describe the bug**
When connecting to an mssql named instance, the connection string generated in `tortoise.mssql.client.MSSQLClient` still includes the port number. Named instances listen on dynamic ports, so the `,{port}` should be left out, otherwise it tries to connect by default to port 1433 which fails.

**To Reproduce**
Create an mssql connection with host containing a named instance, for example `my_server\my_instance`

**Expected behavior**
It should work

**Additional context**
One possible fix could be:
```python
    if "\\" in host:
        # named instance, don't provide port
        server = host
    else:
        server = f"{host},{port}"
    self.dsn = f"DRIVER={driver};SERVER={server};UID={user};PWD={password};"
```

PR #1600, opened 2024-05-01
Fix dsn for mssql named instances

adi1:
## Description

Remove port from an mssql dsn when using named instances.

## Motivation and Context

MSSQL named instances  use dynamic port numbers. You can either connect directly to the port, or to the named instance.
Providing a DB URL like `mssql://user:pass@host\instance/db?driver=FreeTDS` will result in a DSN which looks like `DRIVER=FreeTDS;SERVER=host\instance,1433;UID=user;PWD=pass;` because the default port is added. This fails to connect, because the instance is not listening on port 1443.

https://github.com/tortoise/tortoise-orm/issues/1566

## How Has This Been Tested?

I have only tested locally with and without named instances.

## Checklist:

### DIFF
--- a/tortoise/backends/mssql/client.py
+++ b/tortoise/backends/mssql/client.py
@@ -35,7 +35,12 @@ class MSSQLClient(ODBCClient):
         **kwargs: Any,
     ) -> None:
         super().__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


abondar:
Hello!

Please try to write test for it
You can limit test to only mssql the same way we do it for `postgres/mysql` tests in `tests/backends` as example

PR #1608, opened 2024-05-09
Remove duplicate version info

waketzheng:
## Description

Both `pyproject.toml` and `tortoise/__init__.py` have version info (current is 0.20.1)

There are two ways to make it only one: 
- The current [official way of doing it without duplicating the value](https://github.com/python-poetry/poetry/pull/2366#issuecomment-652418094) is with `importlib.metadata`
- Another way is use a poetry plugin [poetry-version-plugin](https://github.com/tiangolo/poetry-version-plugin)

As we do not need to support python3.7 and `importlib.metadata` is much more simple, so I use it in `tortoise/__init__.py`, and introduce new testcase `tests/test_version.py`

## Motivation and Context

Note: At local development environment, when version in pyproject.toml was changed, `python -c 'from tortoise import __version__;print(__version__)'` will not print the new one, unless you run `make deps` again.

## How Has This Been Tested?

make ci

## Checklist:

### DIFF
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -1,0 +1,8 @@
+import subprocess
+
+from tortoise import __version__
+
+
+def test_version():
+    r = subprocess.run(["poetry", "version", "-s"], capture_output=True)
+    r.stdout.decode() == __version__
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -1,8 +1,19 @@
-import subprocess
+import sys
+from pathlib import Path
 
 from tortoise import __version__
 
+if sys.version_info >= (3, 11):
+    import tomllib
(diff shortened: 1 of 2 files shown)


tests/test_version.py
```diff
@@ -0,0 +1,8 @@
+import subprocess
+
+from tortoise import __version__
+
+
+def test_version():
+    r = subprocess.run(["poetry", "version", "-s"], capture_output=True)
+    r.stdout.decode() == __version__
```
abondar, line 8: assert here probably?

waketzheng: fixed

abondar (approved):

PR #1613, opened 2024-05-13
More efficient `utils.chunk`

Abdeldjalil-H:
The `utils.chunk` function from `tortoise.utils` is converting `instances` into list. I think it is better to not load all the list at one time but rather in chunks as the name suggested. This is useful in cases when the user passes a generator instead of list especially for large datasets.

## Checklist:


Abdeldjalil-H:
It seems that the failing tests have nothing to do with the changes made, because the changed part is not being called at all for the failed test case. Moreover, it passes the first time, then when I changed contrubitors file it fails!

abondar (approved):

PR #1617, opened 2024-05-17
fix index name

Abdeldjalil-H:
Use `index_name` from class `Index` to get index name instead of `BaseSchemaGenerator._generate_index_name`.

## Motivation and Context
This allows you to create your custom index class that inherits from `Index` and override `index_name` method to set index name. 

## How Has This Been Tested?
On existing tests.

## Checklist:

### DIFF
--- a/tortoise/indexes.py
+++ b/tortoise/indexes.py
@@ -40,24 +40,17 @@ class Index:
 
     def get_sql(self, schema_generator: "BaseSchemaGenerator", model: "Type[Model]", safe: bool):
         if self.fields:
-            return self.INDEX_CREATE_TEMPLATE.format(
-                exists="IF NOT EXISTS " if safe else "",
-                index_name=schema_generator.quote(
-                    self.name or schema_generator._generate_index_name("idx", model, self.fields)
-                ),
-                index_type=f" {self.INDEX_TYPE} ",
-                table_name=schema_generator.quote(model._meta.db_table),
-                fields=", ".join(schema_generator.quote(f) for f in self.fields),
-                extra=self.extra,
-            )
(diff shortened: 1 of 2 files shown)


abondar:
Is there reason why defining `name` for Index class doesn't work for you?

tortoise/indexes.py
```diff
@@ -40,24 +40,17 @@ def __init__(
...
-            )
+            fields = ", ".join(schema_generator.quote(f) for f in self.fields)
+        else:
+            expressions = [f"({expression.get_sql()})" for expression in self.expressions]
+            fields = ", ".join(expressions)
 
-        expressions = [f"({expression.get_sql()})" for expression in self.expressions]
         return self.INDEX_CREATE_TEMPLATE.format(
             exists="IF NOT EXISTS " if safe else "",
-            index_name=self.index_name(schema_generator, model),
```
Abdeldjalil-H, line 57: here the method `index_name` is called, but for the first case the code of `index_name` is used instead of calling the method.

Abdeldjalil-H:
In my case I am using a custom index class:

```python

class MyCustomIndex(Index):
    PREFIX = "idx"
    def index_name(self, schema_generator: BaseSchemaGenerator, model: models.Model):
        if self.name:
            return self.name

        table_name = model._meta.db_table

        return f"{self.PREFIX}_{table_name}_{'_'.join(self.fields)}".lower()[:64]
```

Then I wrap all my indexes with this custom class so the name is generated automatically. But when I run the migrations, I still get the old name.

By the way, this is not the case if you pass `expressions` instead of `fields` as you can see in the [previous code](https://github.com/tortoise/tortoise-orm/pull/1617/files#r1605711102).

abondar (approved):

PR #1620, opened 2024-05-22
refactor: optimize m2m add logic

waketzheng:
## Description

Fixes TODO item and make code more clear

## Motivation and Context

The `pk_formatting_func` function only need to be call once

## How Has This Been Tested?

make ci

## Checklist:


abondar:
Although this refactoring seems good to me and I am okay with accepting it as is - I am not sure if anything changed regarding concern in TODO? 

As I understand it - it was about the fact that we needed to have indexes in m2m through table, and after refactoring table is still same and there is no unique index there

waketzheng:
> Although this refactoring seems good to me and I am okay with accepting it as is - I am not sure if anything changed regarding concern in TODO?
> 
> As I understand it - it was about the fact that we needed to have indexes in m2m through table, and after refactoring table is still same and there is no unique index there

@abondar done.

- A simple test script:
```py
# main.py
import time
from typing import Callable

from tortoise import Tortoise, fields, models, run_async


def timeit(func) -> Callable:
    async def runner(*args, **kw) -> None:
        start = time.time()
        await func(*args, **kw)
        end = time.time()
        print(func.__name__, "Cost:", round(end - start, 1))

    return runner


class Team(models.Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(20)


class Event(models.Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(20)
    teams = fields.ManyToManyField("models.Team")


async def main() -> None:
    await _test()


@timeit
async def _test() -> None:
    n = 100_000
    instances = []
    for m in (Team, Event):
        objs = [m(id=i + 1, name=i) for i in range(n)]
        await m.bulk_create(objs)
        instances.append(await m.all())
    for t, e in zip(*instances):
        await e.teams.add(t)
        await e.save()


if __name__ == "__main__":
    run_async(main())
```
Creates 100 thousand records for team, event, event_teams.

Output of branch develop:
_test Cost: 287.6

While output of branch refactor-m2m-add is:
_test Cost: 66.7

To reproduce:
```bash
git clone git@github.com:waketzheng/tortoise-orm
cd tortoise-orm
make deps
python main.py
git checkout refactor-m2m-add
make deps
python main.py
```

poetry.lock
```diff
@@ -859,17 +864,17 @@ toml = ["tomli"]
 
 [[package]]
 name = "coveralls"
-version = "3.3.1"
```
abondar, line 862: @waketzheng Downgrade of coveralls breaks CI

abondar (approved):

Issue #1641, opened 2024-06-06
unique indexes (for unique_together) on nullable columns need to be optionable for nulls not distinct


omidekz:
assume that i have two models `shop, `customer`. each shop can have only one customer with an unique phone number till deletes it. please pay attention that customer's phone_number is not unique in column. that is unique together with its shop and delete state.

```py
from tortoise import Model, fields

class Base(Model):
  deleted_at = fields.DateTimeField(null=True)

class Shop(Base):
  title = fields.CharField(...)

class Customer(Base):
  phone_number = fields.CharField(...) # not used unique here as you expect
  shop = fields.ForeignKeyField('models.Shop', 'customers', fields.CASCADE)
  class Meta:
    unique_together = ('shop_id', 'phone_numver', 'deleted_at')
```
but this simple code is not work as work i expected
in postgres that i tested, created index didnt follow by `nulls not distict`
when not follow. unique index consider nulls not equal.
tortoise version : 0.21.3

abondar:
Hi!

The fact that it does not use `nulls not distinct` is intentional (as much intentional it can be for the case of default behaviour)

And I am pretty sure it shouldn't be default behaviour

Although adding support would be reasonable, I think it should be done by allowing it as param in `tortoise.contrib.postgres.indexes.PostgreSQLIndex`, not by changing behaviour in `unique_together`

You can contribute this change in PR, or may be I'll be able to take it on somewhere later

I think you can bypass this problem pretty easily by modifying your migration scripts if you have one, or creating index manually if you don't use migration system, as indexes declared on models doesn't affect runtime behaviour

omidekz:
Hi
that was great idea. by changes at this [PR](https://github.com/tortoise/tortoise-orm/pull/1642) i solved my problem with following approach:
```py
from tortoise... import PostgresUniqueIndex as PUI, SqliteUniqueIndex as SUI

def unique_index_depends_on_current_db_type(fields, null_fields):
  """this function just support sqlite/postgres unique index and nulls will be consider equal"""
  is_postgres = envs.db_url.lower().startswith('postgres')
  unique_index_class = PUI if is_postgres else SUI
  return unique_index_class(fields=fields, nulls='not distinct' if is_postgres else null_fields)

class Customer(BaseModel):
  class Meta:
    indexes = [unique_index_depends_on_current_db_type(fields=['shop_id','phone_number', 'deleted_at'], null_fields=['deleted_at'])]
```

[...]

PR #1642, opened 2024-06-07
create unqiue index in indexes and related children in contrib.postgres/sqlite.indexes

omidekz:
create a postgresql unique index and consider the `nulls not distinct`

## Description
enhancement was started from this [issue](https://github.com/tortoise/tortoise-orm/issues/1641) 

this PR provide a manual solution for postgres driver. but i think the orm has to provide a base/shared solution.
the solution is a way to consider nulls equal in unique indexes on that drivers that consider nulls not equal.

postgres can done that by : `nulls not distinct`
sqlite and sqlserver too by : index expressions
on mysql i don't know yet.

### DIFF
--- a/tortoise/contrib/postgres/indexes.py
+++ b/tortoise/contrib/postgres/indexes.py
@@ -1,8 +1,10 @@
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Type
 
 from pypika.terms import Term, ValueWrapper
 
+from tortoise.backends.base.schema_generator import BaseSchemaGenerator
 from tortoise.indexes import PartialIndex
+from tortoise.models import Model
 
 
 class PostgreSQLIndex(PartialIndex):
@@ -49,3 +51,26 @@ class HashIndex(PostgreSQLIndex):
 
 class SpGistIndex(PostgreSQLIndex):
     INDEX_TYPE = "SPGIST"
+
+
+class PostgresUniqueIndex(PostgreSQLIndex):
+    INDEX_CREATE_TEMPLATE = PostgreSQLIndex.INDEX_CREATE_TEMPLATE.replace(
+        "CREATE", "CREATE UNIQUE"
+    ).replace("USING", "")
+
+    def __init__(
+        self,
+        *expressions: Term,
+        fields: Optional[Tuple[str]] = None,
+        name: Optional[str] = None,
+        condition: Optional[dict] = None,
+        nulls_not_distinct: bool = False,
+    ):
+        super().__init__(*expressions, fields=fields, name=name, condition=condition)
+        if nulls_not_distinct:
+            self.extra = " nulls not distinct".upper() + self.extra
+
+    def get_sql(self, schema_generator: BaseSchemaGenerator, model: Type[Model], safe: bool):
+        if self.INDEX_TYPE:
+            self.INDEX_TYPE = f"USING {self.INDEX_TYPE}"
+        return super().get_sql(schema_generator, model, safe)
--- a/tortoise/contrib/sqlite/indexes.py
+++ b/tortoise/contrib/sqlite/indexes.py
@@ -1,0 +1,46 @@
+from typing import Dict, Optional, Tuple
+
+from pypika.terms import Term, ValueWrapper
+
+from tortoise.indexes import PartialIndex
+
+
+class SqliteIndex(PartialIndex):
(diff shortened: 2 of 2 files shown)


omidekz:
june 8 related changes:
 - add `UniqueIndex` in `tortoise.indexes`
 - reimplement contrib.postgres.indexes.UniqueIndex
 - implemet contribe.sqlite.indexes.UniqueIndex
Description: todo

abondar:
Why you decided to make it as separate class? 
I am concerned that it complicated class hierarchy - and now, for example, it is not clear how to make unique index of non-default type for postgres, as there are now separately `HashIndex` for example, and Unique index, and it is not clear how to make unique `HashIndex`

omidekz:
Hi i get your point. i'll fix shared unique index
but i have a challenge with partialindex.
at the `tortoise.indexes` file line of 75 where you are use `ValueWrapper` k = wrapped value. and we need k is not null
if i check the wrapped_value is startswith(is) . its mean i produce some other issues.

tortoise/contrib/postgres/indexes.py
```diff
@@ -49,3 +51,26 @@ class HashIndex(PostgreSQLIndex):
...
+        condition: Optional[dict] = None,
+        nulls_not_distinct: bool = False,
+    ):
+        super().__init__(*expressions, fields=fields, name=name, condition=condition)
+        if nulls_not_distinct:
+            self.extra = " nulls not distinct".upper() + self.extra
+
+    def get_sql(self, schema_generator: BaseSchemaGenerator, model: Type[Model], safe: bool):
+        if self.INDEX_TYPE:
+            self.INDEX_TYPE = f"USING {self.INDEX_TYPE}"
```
abondar, line 75: Sorry, I was wrong about this, I forgot that postgresql allows only default b-tree index to be unique, so it is something we don't need to support

tortoise/contrib/sqlite/indexes.py
```diff
@@ -0,0 +1,46 @@
...
+
+from tortoise.indexes import PartialIndex
+
+
+class SqliteIndex(PartialIndex):
+    INDEX_CREATE_TEMPLATE = PartialIndex.INDEX_CREATE_TEMPLATE.replace("INDEX", "INDEX {exists} ")
+
+
+class SqliteUniqueIndex(SqliteIndex):
+    INDEX_TYPE = "unique".upper()
```
abondar, line 13: Why not just `"UNIQUE"`?

tortoise/contrib/sqlite/indexes.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(
+        self,
+        *expressions: Term,
+        fields: Optional[Tuple[str, ...]] = None,
+        name: Optional[str] = None,
+        condition: Optional[Dict[str, str]] = None,
+        where_expre: Optional[str] = None,
```
abondar, line 21: I am a little at lost here
Why do we need both `condition` and `where_expre`?

tortoise/contrib/sqlite/indexes.py
```diff
@@ -0,0 +1,46 @@
...
+            if isinstance(cond, str)
+            and cond.strip().lower().startswith(("is ", "isnot ", "<", ">", "=", "!="))
+            else "="
+        )
+        if op == "=":
+            cond = ValueWrapper(cond)
+        return str(f"{key} {op} {cond}")
+
+    def _gen_condition(self, conditions: Dict[str, str]):
+        conditions = " AND ".join(tuple(map(self._gen_field_cond, conditions.items())))
```
abondar, line 45: please use comprehensions

```
" AND ".join(self._gen_field_cond(k, v) for k, v in condition.items())
```

With that you will also be able to make better signature and typing for `_gen_field_cond`

tortoise/contrib/sqlite/indexes.py
```diff
@@ -0,0 +1,46 @@
...
+        _condition = condition
+        if condition:
+            condition = None
+        super().__init__(*expressions, fields=fields, name=name, condition=condition)
+        if _condition:
+            # TODO: what is the best practice to inject where expression?
+            self.extra = where_expre or self._gen_condition(_condition)
+
+    @classmethod
+    def _gen_field_cond(cls, kv: tuple):
```
abondar, line 32: Add return typing

PR #1657, opened 2024-06-17
Add init_memory_sqlite decorator

waketzheng:
## Description

Almost every file in examples include the following two lines:
```
```

## Motivation and Context

- [x] Proposal 1: Add  a decorator to initial memory sqlite
- [ ] Proposal 2: Add a optional argument into run_async to init Tortoise
- Example1:
```py
from tortoise import fields, models, run_async

class User(models.Model):
    name = fields.TextField()

@init_memory_sqlite
async def run():
    user = await User.create(name='J')
    print(user.name)

if __name__ == '__main__':
    run_async(run())
```
- Example2:
```py
# app/models.py

class User(models.Model):
    name = fields.TextField()
```
```py
# main.py
from tortoise import run_async

from app.models import User

@init_memory_sqlite(['app.models'])
async def run():
    user = await User.create(name='J')
    print(user.name)

if __name__ == '__main__':
    run_async(run())
```

## How Has This Been Tested?

make ci

## Checklist:


examples/basic.py
```diff
@@ -18,10 +19,8 @@ def __str__(self):
         return self.name
 
 
-async def run():
-    await Tortoise.init(db_url="sqlite://:memory:", modules={"models": ["__main__"]})
-    await Tortoise.generate_schemas()
-
+@init_memory_sqlite
```
abondar, line 22: May be we should make it configurable? To let it pass another `modules` value

Because if you allow only `__main__` I am not sure it would be useful enough to make it part of contrib

waketzheng: @abondar I made a refactor to support `models` argument.

abondar (approved):

PR #1689, opened 2024-08-01
Fix `bulk_update` when using a pk other than `id`

abvf:
In method `bulk_update`, the primary key name is hard-coded as `"id"`; therefore, it throws an error when using a primary key with a name other than `id`.

## Checklist:


abondar:
Hi!

Thanks for your submission
Could you please add some kind of test that reproduces issue?
You can write it somewhere here https://github.com/tortoise/tortoise-orm/blob/develop/tests/test_update.py#L37

abvf:
Hi @abondar,

I have added a test with the Event model, as its primary key is not `id`. The test will fail without the fix committed in this pull request.

abondar (approved):

PR #1696, opened 2024-08-11
Fix deadlock for Sanic apps with `generate_schema=True` and > 1 workers

Pirulax:
## Description/Motivation and Context
Fixes deadlock caused by multiple workers trying to initialize the schema at the same time.
In my case this deadlock was caught by `asyncpg`.
I did try making a minimum repro sample, but failed, and honestly, don't have the time to fiddle with it.
I can't post my app, as it's private.

## How Has This Been Tested?
I ran it in my own app where the deadlock previously occurred, now it doesn't and the schema is created as expected.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Changelog
 Fixed
 ^^^^^
 - Fix bug in `pydantic_model_creator` when a foreign key is not included in `include` param. (#1430)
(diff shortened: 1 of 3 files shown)


abondar:
Hi!

Please run `make lint` to see lint errors

Pirulax:
Hello,
Not sure if there are docs of this somewhere, but on Windows both `make lint` and `make style` fails out of the box.
I had to manually install `black` and `isort` using `pipx`.

abondar (approved):

PR #1737, opened 2024-10-14
Add support for case insensitive regexp + add support for REGEXP SQLite module

larsschwegmann:
This PR adds support for case insentive regex filtering (postgres and sqlite). It also adds support for the REGEXP sqlite module (needs to be installed in sqlite to work).

## Description
Postgres and SQLite offer ways to query with regular expression while ignoring case, similiarly how we already have case insensitive contains, etc. This PR adds support for these filters.

This PR also adds support for [SQLite](https://www.sqlite.org/lang_expr.html) (regexp() function needs to be supplied at runtime, e.g. via [an extension](https://github.com/nalgeon/sqlean/blob/main/docs/regexp.md)). It is installed by default on most linux distributions of sqlite.

## Motivation and Context
I want to query case insensitive regex patterns. Having sqlite support is good for completeness. Many people including me use sqlite for unit tests, so having feature parity across db backends is aa big plus.

## How Has This Been Tested?
Ran the tests suite

## Checklist:

### DIFF
--- a/tortoise/backends/sqlite/client.py
+++ b/tortoise/backends/sqlite/client.py
@@ -12,7 +12,7 @@ from typing import (
     Sequence,
     Tuple,
     TypeVar,
-    cast,
+    cast, override,
 )
 
 import aiosqlite
@@ -29,7 +29,7 @@ from tortoise.backends.base.client import (
 )
 from tortoise.backends.sqlite.executor import SqliteExecutor
 from tortoise.backends.sqlite.schema_generator import SqliteSchemaGenerator
-from tortoise.contrib.sqlite.regex import install_regexp_function
+from tortoise.contrib.sqlite.regex import install_regexp_functions as install_regexp_functions_to_db
 from tortoise.connection import connections
 from tortoise.exceptions import (
     IntegrityError,
@@ -85,7 +85,6 @@ class SqliteClient(BaseDBAsyncClient):
             for pragma, val in self.pragmas.items():
                 cursor = await self._connection.execute(f"PRAGMA {pragma}={val}")
                 await cursor.close()
-            await install_regexp_function(self._connection)
             self.log.debug(
                 "Created connection %s with params: filename=%s %s",
                 self._connection,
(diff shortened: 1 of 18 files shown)


abondar:
@larsschwegmann Hi!

Sorry I didn't pay attention to this PR, seems like I missed it

Could you please add some tests for that feature?

larsschwegmann:
Sorry, I also missed your response while i was on vacation. I'll add tests later today 👍

larsschwegmann:
@abondar Hi again! I added tests for the new functionality. Please take a look and let me know what you think. Feel free to merge if it looks okay to you 😄

tortoise/backends/sqlite/client.py
```diff
@@ -81,6 +82,7 @@ async def create_connection(self, with_db: bool) -> None:
             for pragma, val in self.pragmas.items():
                 cursor = await self._connection.execute(f"PRAGMA {pragma}={val}")
                 await cursor.close()
+            await install_regexp_function(self._connection)
```
abondar: Seems a bit strange that we are automatically installing new functions in user db just on connection

May be we should add some kind of flag to make it explicit and only in case user need it? 
If we do so, also would be nice to check if it was activated inside function and propagate proper error for user to prompt him to switch that flag

Even though it sounds like unneeded extra steps - it seems reasonable to not modify user db without explicit requests to do so

larsschwegmann: yeah, you are probably right about that. although it does not modify the db, it could introduce bad behaviour in case a user has already overwritten the `regexp` or `match` functions. I like the approach with the flag you proposed, I'll try and implement that 😄

larsschwegmann:
Hi @abondar !
I made the installation of the regexp user functions to the sqlite engine optional via a config credential. I added a new run group to the makefile (`make test_sqlite_regexp`) that runs the sqlite tests with `TORTOISE_TEST_DB=sqlite://:memory:?install_regexp_functions=True`. This is how users can also enable it in their own sqlite configs via the DB URL. I also added an entry to the `Capabilities` class, so that every db client can expose that info and the tests are excluded when using the vanilla `sqlite://:memory:` engine.
I hope you are fine with my solution. Please let me know if this works for you 😄

abondar:
@larsschwegmann Hi!

Sorry, I missed your update and now branch is stale
Could you please rebase it, so I will be able to merge it?

larsschwegmann:
Hi @abondar , thanks for reaching out :)
I have rebased the branch 👍

larsschwegmann:
Hi @abondar , could you please check when you have time? I have rebased again

abondar (approved):

larsschwegmann:
Thank you for merging @abondar ! 😄 
Do you happen to know if there are plans already when the next version will be released?

abondar:
I think it should be soon enough
Afaik - @henadzit preparing new version now

PR #1747, opened 2024-10-23
Disable validation of annotations

### LINKED ISSUE #1502: Problem with Sum with F and Field validator MinValue
The Model

```python
class Deal(models.Model):
   ...
    broker_payed_money = fields.FloatField(default=0, validators=[MinValueValidator(0)])

    broker_commission = fields.FloatField(null=False, validators=[GreaterThanValidator(0), LessThanValidator(1.0)])

    developer_payed_money = fields.FloatField(null=False, default=0, validators=[MinValueValidator(0)])
   ...
```
and the query
```python
result = await Deal.filter(user=user).annotate(
        balance=Sum(F('developer_payed_money') * F('broker_commission') - F('broker_payed_money')),
    ).values_list('balance')
```
raises the Exception
```
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/cors.py", line 84, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/usr/local/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 706, in __call__
    await route.handle(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 66, in app
    response = await func(request)
  File "/usr/local/lib/python3.10/site-packages/fastapi/routing.py", line 235, in app
    raw_response = await run_endpoint_function(
  File "/usr/local/lib/python3.10/site-packages/fastapi/routing.py", line 161, in run_endpoint_function
    return await dependant.call(**values)
  File "/opt/project/app/routers/cabinet/users.py", line 25, in init_data
    return await init_data_handler(user)
  File "/opt/project/app/handlers/cabinet/users.py", line 31, in init_data_handler
    await _get_balances(user)
  File "/opt/project/app/handlers/cabinet/users.py", line 23, in _get_balances
    result = await Deal.annotate(
  File "/usr/local/lib/python3.10/site-packages/tortoise/queryset.py", line 1506, in _execute
    lst_values = list(map(listmap, result))
  File "/usr/local/lib/python3.10/site-packages/tortoise/queryset.py", line 1505, in <lambda>
    listmap = lambda entry: tuple(func(entry[column]) for column, func in columns)  # noqa
  File "/usr/local/lib/python3.10/site-packages/tortoise/queryset.py", line 1505, in <genexpr>
    listmap = lambda entry: tuple(func(entry[column]) for column, func in columns)  # noqa
  File "/usr/local/lib/python3.10/site-packages/tortoise/fields/base.py", line 225, in to_python_value
    self.validate(value)
  File "/usr/local/lib/python3.10/site-packages/tortoise/fields/base.py", line 244, in validate
    raise ValidationError(f"{self.model_field_name}: {exc}")
tortoise.exceptions.ValidationError: broker_payed_money: Value should be greater or equal to 0
```

But drop validator MinValueValidator from broker_payed_money avoids this problem.

Query column `balance`  may be negative.

Version 0.19.2, tried 0.20.0

henadzit:
## Description
* This should fix #1502.
* Extends `Field.to_python_value` with the `validate` argument with the default value `True` which regulates whether validation should be run.
* Disables validation for annotations.
* ⚠️ This is a not backwards compatible change for those who extended `Field` and overrode `to_python_value` because its signature changed 

## Motivation and Context

When an arithmetic expressions on a model field is used in annotations, the field's `to_python_value` is used to process the database value. `to_python_value` is running validations by default and if the result of aggregation function doesn't pass field's validation, an exception happens.

```python
class ValidatorModel(Model):
  max_value = fields.IntField(null=True, validators=[MaxValueValidator(20.0)])

await ValidatorModel.create(max_value=4)
await ValidatorModel.create(max_value=4)
await ValidatorModel.annotate(sum=Sum(F("max_value") * F("max_value"))).values("sum")
```

will cause `tortoise.exceptions.ValidationError: max_value: Value should be less or equal to 20.0`

## Other ideas

At the moment the validation is being run when the object is being loaded from the database, for instance, [here](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/models.py#L749) and [here](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/queryset.py#L1429). Does it really need to happen? The introduction of the `validate` argument can be used to disable validation in these cases.

## How Has This Been Tested
* Added tests exposing the issue
* Tested the code from #1502:

```python
from tortoise.functions import Sum
from tortoise.expressions import F, Q
from tortoise.validators import MinValueValidator

class Deal(Model):
    test = fields.IntField()
    broker_payed_money = fields.FloatField(default=0, validators=[MinValueValidator(0)])

    broker_commission = fields.FloatField(null=False)

    developer_payed_money = fields.FloatField(
        null=False, default=0, validators=[MinValueValidator(0)]
    )


async def main():
    # Initializing Tortoise and creating the schema

    await Tortoise.init(  # type: ignore
        db_url="sqlite://:memory:",
    )

    await Deal.create(
        test=1, broker_payed_money=10, broker_commission=0.1, developer_payed_money=20
    )

    result = await Deal.annotate(
        balance=Sum(F("developer_payed_money") * F("broker_commission") - F("broker_payed_money")),
    ).values_list("balance")
    print("res", result)


run_async(main())
```

## Checklist:

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -95,7 +95,7 @@ to store and query Enum types.
         def to_db_value(self, value: Enum, instance) -> str:
             return value.value
 
-        def to_python_value(self, value: str) -> Enum:
+        def to_python_value(self, value: str, validate: bool = True) -> Enum:
             try:
                 return self._enum_type(value)
             except Exception:
--- a/tests/fields/subclass_fields.py
+++ b/tests/fields/subclass_fields.py
@@ -30,8 +30,9 @@ class EnumField(CharField):
 
         return value.value
 
-    def to_python_value(self, value):
-        self.validate(value)
+    def to_python_value(self, value, validate: bool = True):
+        if validate:
+            self.validate(value)
 
         if value is None or isinstance(value, self.enum_type):
             return value
@@ -66,8 +67,9 @@ class IntEnumField(IntField):
 
         return value.value
 
-    def to_python_value(self, value: Any) -> Any:
-        self.validate(value)
+    def to_python_value(self, value: Any, validate: bool = True) -> Any:
+        if validate:
+            self.validate(value)
 
         if value is None or isinstance(value, self.enum_type):
             return value
--- a/tests/test_aggregation.py
+++ b/tests/test_aggregation.py
@@ -1,8 +1,9 @@
-from tests.testmodels import Author, Book, Event, MinRelation, Team, Tournament
+from decimal import Decimal
+from tests.testmodels import Author, Book, Event, MinRelation, Team, Tournament, ValidatorModel
 from tortoise.contrib import test
 from tortoise.contrib.test.condition import In
 from tortoise.exceptions import ConfigurationError
-from tortoise.expressions import Q
+from tortoise.expressions import F, Q
 from tortoise.functions import Avg, Coalesce, Concat, Count, Lower, Max, Min, Sum, Trim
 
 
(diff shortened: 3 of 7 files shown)


henadzit:
Hi @abondar! This PR contains a non backwards compatible change. How should I treat the changelog? Should I introduce `0.22.1` there with the note of this change as part of this PR? Thanks!

abondar:
Hi!

Why you have decided to go with approach of turning off validation, rather than just making sure that annotated field instance doesn't validators? 

Seems like if we change [this line](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/queryset.py#L1423)
From 
```
        if field in self.annotations:
            annotation = self.annotations[field]
            field_object = getattr(annotation, "field_object", None)
            if field_object:
                return field_object.to_python_value
            return lambda x: x
```

To 
```
        if field in self.annotations:
            annotation = self.annotations[field]
            field_object = getattr(annotation, "field_object", None)
            if field_object:
                new_field_class = deepcopy(field_object)
                new_field_class.validators = []
                return new_field_class.to_python_value
            return lambda x: x
         
```

We can exterminate root problem, of validators leaking to dynamically generated fields. 
And that way we won't have to introduce breaking changes

henadzit:
@abondar thank you for reviewing and challenging my approach!

I gave it a bit more thought and realized that validation on loading records from the database is not common. I prepared another pull request, please have a look!  

https://github.com/tortoise/tortoise-orm/pull/1750

Issue #1299, opened 2022-11-27
Can't use annotated field in CASE-WHEN Clause


kirillismad:
**Describe the bug**
I have `tortoise.exceptions.FieldError: Unknown filter param 'waiting_time__gte'. Allowed base values are ['client_info', 'created_at', 'guest_info', 'id', 'manager', 'manager_id', 'priority', 'source']`

**To Reproduce**
Model:
```python
class Chat(Model):
    id = fields.UUIDField(pk=True)
    created_at = fields.DatetimeField(auto_now_add=True)
    manager = fields.ForeignKeyField(
        'models.Manager',
        related_name='chat_requests',
        null=True,
        on_delete=fields.SET_NULL,
    )
    priority = fields.IntField(
        validators=[
            MinValueValidator(MIN_PRIORITY),
            MaxValueValidator(MAX_PRIORITY),
        ]
    )
    source = fields.IntEnumField(ChatSource)
    guest_info = fields.JSONField(null=True)
    client_info = fields.JSONField(null=True)
```

Query:
```python
queryset = Chat.all()
queryset = queryset.annotate(
    type=Case(
        When(Q(guest_info__isnull=False), then='GUEST'),
        When(Q(client_info__isnull=False), then='CLIENT'),
    ),
)
queryset = queryset.annotate(
    waiting_time=RawSQL('CURRENT_TIMESTAMP') - F('created_at'),
)

queryset = queryset.annotate(
    ignore_priority=Case(
        When(Q(priority=1, waiting_time__gte=timedelta(minutes=10)), then=True),
        When(Q(priority__in=(2, 3), waiting_time__gte=timedelta(minutes=15)), then=True),
        When(Q(priority__in=(4, 5), waiting_time__gte=timedelta(minutes=25)), then=True),
        default=False,
    ),
)
```

**Expected behavior**
Expected result is: queryset, which I can await and get `Iterable[Chat]`, with attribute `ignore_priority`

**Additional context**
Full traceback:
```python
File "/usr/local/lib/python3.9/site-packages/tortoise/queryset.py", line 998, in __await__
    self._make_query()
  File "/usr/local/lib/python3.9/site-packages/tortoise/queryset.py", line 960, in _make_query
    self.resolve_filters(
  File "/usr/local/lib/python3.9/site-packages/tortoise/queryset.py", line 126, in resolve_filters
    has_aggregate = self._resolve_annotate()
[...]

Issue #477, opened 2020-08-05
Annotate by field of the related model


mirrorrim:
Sometimes I need to annotate queryset with field of the related model.  

For example I want to get comments with an annotated user name for these models.

```python
from tortoise import fields
from tortoise.models import Model


class User(Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(max_length=100, null=True, default=None)
    comments: fields.ReverseRelation["Comment"]


class Comment(Model):
    id = fields.IntField(pk=True)
    user: fields.ForeignKeyRelation["User"] = fields.ForeignKeyField(
        "models. User", on_delete=fields.CASCADE, related_name="comments"
    )
    text = fields.TextField()
```

In Django I can do something like this:

```python
comments = Comment.object.annotate(user_name=F("user__name"))
```

It would be nice to have similar functionality in Tortoise ORM:

```python
comments = await Comment.annotate(user_name=F("user__name"))
```

Thanks.

lqmanh:
Not professional here but isn't it already supported by Tortoise ORM?
Have you had a look at [this](https://tortoise-orm.readthedocs.io/en/latest/examples/basic.html?highlight=annotate#functions)?

mirrorrim:
@lqmanh thanks for the asnwer! Indeed, this will work if you use an additional function, such as Coalesce (which is OK for my case), but sometimes I don't need any modifications, only join with additional field.

lqmanh:
@mirrorrim Then you can just use `F` expression like that?

mirrorrim:
@lqmanh unfortunately it doesn't work. It raises an exception like this:

```python
 In[3] i = await IncomePayment.annotate(
  ...:     description=F("invoice__description"),
  ...: )
Traceback (most recent call last):
  File "/Users/mirrorrim/Library/Caches/pypoetry/virtualenvs/server-wQAZuLdb-py3.8/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3341, in run_code
    await eval(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-717be70f04d7>", line 1, in <module>
    i = await IncomePayment.annotate(
  File "/Users/mirrorrim/Library/Caches/pypoetry/virtualenvs/server-wQAZuLdb-py3.8/lib/python3.8/site-packages/tortoise/models.py", line 1049, in annotate
    return QuerySet(cls).annotate(**kwargs)
[...]

PR #1748, opened 2024-10-24
Annotated Fields in Case-When

henadzit:
## Description
This PR:
- Enables use of annotations in Case-When, e.g. 
```python
IntFields.all()\
  .annotate(intnum_plus_1=F("intnum") + 1)\
  .annotate(bigger_than_10=Case(When(Q(intnum_plus_1__gte=10), then=True), default=False))
```
- Introduces `ResolveContext` that holds the information required to resolve expressions
- Fixes https://github.com/tortoise/tortoise-orm/issues/1299

## Motivation and Context

Django lets you use annotations in Case-When, so it would be nice to have it too.

`ResolveContext` simplifies the code by incapsulating all the information required for resolving expressions, i.e. the model, the table, annotations and custom filters as opposed to the old way of assigning `_annotations` and `_custom_filters` to expressions. I'm planning on using it to resolve `F` expressions, for instance, to allow referring to related models in `F` ([related issue](https://github.com/tortoise/tortoise-orm/issues/477)).

## How Has This Been Tested?
- Added tests
- The following snippet can be used to compare the behavior prior and after the changes

```python
import logging
import sys

from tortoise.expressions import F, Q, Case, When

fmt = logging.Formatter(
    fmt="%(asctime)s - %(name)s:%(lineno)d - %(levelname)s - %(message)s",
    datefmt="%Y-%m-%d %H:%M:%S",
)
sh = logging.StreamHandler(sys.stdout)
sh.setLevel(logging.DEBUG)
sh.setFormatter(fmt)

logger_tortoise = logging.getLogger("tortoise")
logger_tortoise.setLevel(logging.DEBUG)
logger_tortoise.addHandler(sh)


class User(Model):
    value = fields.IntField()


async def main():
    # Initializing Tortoise and creating the schema

    await Tortoise.init(  # type: ignore
        db_url="sqlite://:memory:",
    )

    await User.create(value=1)
    await User.create(value=2)

    queryset = User.all()
    queryset = queryset.annotate(
        value2=F("value") + 1,
    )
    queryset = queryset.annotate(
        ignore_priority=Case(
            When(Q(value2=2), then=True),
            default=False,
        ),
    )
    res = await queryset
    for user in res:
        print(user.value, user.value2, user.ignore_priority)


run_async(main())
```

## Checklist:

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -35,7 +35,7 @@ from tortoise.exceptions import (
     MultipleObjectsReturned,
     ParamsError,
 )
-from tortoise.expressions import Expression, F, Q, RawSQL
+from tortoise.expressions import Expression, F, Q, RawSQL, ResolveContext
 from tortoise.fields.relational import (
     ForeignKeyFieldInstance,
     OneToOneFieldInstance,
@@ -94,6 +94,7 @@ class AwaitableQuery(Generic[MODEL]):
         "_db",
         "capabilities",
         "_annotations",
+        "_custom_filters",
     )
 
     def __init__(self, model: Type[MODEL]) -> None:
@@ -103,6 +104,7 @@ class AwaitableQuery(Generic[MODEL]):
         self._db: BaseDBAsyncClient = None  # type: ignore
         self.capabilities: Capabilities = model._meta.db.capabilities
         self._annotations: Dict[str, Expression] = {}
+        self._custom_filters: Dict[str, FilterInfoDict] = {}
 
     def _choose_db(self, for_write: bool = False) -> BaseDBAsyncClient:
         """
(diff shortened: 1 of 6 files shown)


tortoise/queryset.py
```diff
@@ -293,7 +316,6 @@ class QuerySet(AwaitableQuery[MODEL]):
         "_q_objects",
         "_distinct",
         "_having",
-        "_custom_filters",
```
henadzit, line 296: I'm having a hard time to understand why `_custom_filters` is changed to `custom_filters` in `QuerySet` subclasses. I would love to know that!

abondar: Honestly - I don't remember :) 

I remember last time I tried to fix it - I ran in some kind of problem related to the fact, that processing of some of this fields have different logic in QuerySet and it's children, so reusing it caused some issues about joins not working properly or something like that

If you have time to experiment with that - you can try to fix it and create PR

abondar (approved):

PR #1758, opened 2024-11-07
refactor: reduce duplicated code of `tortoise.queryset`

waketzheng:
## Description

Reduce duplicated code

## Motivation and Context

refactor

## How Has This Been Tested?

make ci

## Checklist:


henadzit (commented):
Thank you for working on this!

I left a few comments. The summary is that I like the deduplication of ordering parsing but it feels to me that the `_fetch_single` method makes the code more convoluted. Maybe, other maintainers might have different opinion though. Thanks!

tortoise/queryset.py
```diff
@@ -453,59 +453,65 @@ def order_by(self, *orderings: str) -> "QuerySet[MODEL]":
         queryset._orderings = new_ordering
         return queryset
 
-    def latest(self, *orderings: str) -> "QuerySetSingle[Optional[MODEL]]":
+    def _single_queryset(
```
henadzit, line 456: Does it really make sense to have this method? It does not reduce the number of lines really but makes the logic more complicated because it has logical branches.

It replaces pretty clean code which is more like configuration than logic:

Screenshot 2024-11-11 at 00 35 08

henadzit (commented):
Hey! Is this ready for review again?

tortoise/queryset.py
waketzheng: @henadzit You can see the difference here: https://github.com/waketzheng/tortoise-orm/commit/d340109a51cc0a03ec0d8c0dad0bd1dbc31ea78b
There are 5 lines reduced, and 4 `# type: ignore` marks removed.

henadzit: It reduces the code base by 5 lines but adds quite a lot of complexity to the code. It either clones a queryset or modifies it in place. Imagine, if the Python's standard `sorted` function would sometimes return a new sorted list and in some cases sort the list in place. It would be so weird! I feels to me not worth it.

waketzheng: @henadzit I agree that simple is better than complex. I refactor the code to use a new function `_as_single`. https://github.com/tortoise/tortoise-orm/pull/1758/commits/ec0ea01fab6a0b06d15bd2e580db3478fa747fe6

tortoise/queryset.py
```diff
@@ -426,6 +426,33 @@ def exclude(self, *args: Q, **kwargs: Any) -> "QuerySet[MODEL]":
...
+        Convert ordering from strings to standard items for queryset.
+
+        :param orderings: What columns/order to order by
+        :param reverse:  Whether reverse order
+        :return: standard ordering for QuerySet.
+
+        :raises FieldError: If a field provided does not exist in model.
+        """
+        if not orderings:
+            raise FieldError("No fields passed")
```
henadzit, line 442: I think it would be better to change `latest` and `earliest` signatures to avoid raising this exception at all.

```python
def earliest(self, ordering: str, *orderings: str) -> QuerySetSingle[Optional[MODEL]]:
  queryset = self._clone()
  queryset._orderings = self._parse_orderings(ordering + orderings, reverse=True)
  return queryset._as_single()
```

This way the signature itself will make at least one order argument required. It will help to avoid having this check in  `order_by`: 

`queryset._orderings = self._parse_orderings(orderings) if orderings else []`  

because `_parse_orderings` won't be raising exceptions anymore.

henadzit (commented):
Hey, thank you for patience and sorry for not leaving all comments at once! One more comment from me but the code looks good overall

tortoise/queryset.py
henadzit: Thank you for collaborating with me on this!

tortoise/queryset.py
waketzheng: `ordering: str, *orderings: str` style is not readability, so I move `not orderings` check back to latest/earliest and remove the if-else in `order_by`

henadzit: Alright, works for me!

henadzit (approved):
Looks good to me! Thank you for patience!

henadzit (approved):

henadzit:
@abondar can you please have a look when you get a chance? The changes look good to me!

abondar (approved):
@waketzheng thank you for your work!

@henadzit thank you for helpful review, lately I don't have that much time to keep with everything :(

Issue #1703, opened 2024-09-03
🐛 Issue: `Model.annotate(...).update(...)` not working


DavideGalilei:
### Description
`Model.annotate(...).update(...)` does not work, giving an error.

```python-traceback
tortoise.exceptions.FieldError: There is no non-virtual field bonus_salary on Model Employee
```

### Example code
```python
from tortoise import Tortoise, Model, fields, run_async
from datetime import datetime, timedelta
from tortoise.expressions import Case, When, Q, F

class Employee(Model):
    id = fields.IntField(pk=True)
    name = fields.TextField()
    employed_at = fields.DatetimeField(auto_now_add=True)
    salary = fields.BigIntField()

async def main():
    await Tortoise.init(
        db_url='postgres://username:password@localhost:5432/postgres',
        modules={'models': ['__main__']},
    )
    await Tortoise.generate_schemas()

    bob = await Employee.create(name='Bob', employed_at=datetime.now() - timedelta(days=30), salary=1000)

    # If Bob was employed before yesterday, he gets a 10% bonus
    # Doesn't work
    await bob.filter(id=bob.id).select_for_update().annotate(
        bonus_salary=Case(
            When(Q(employed_at__lt=datetime.now() - timedelta(days=1)), then=F('salary') * 1.1),
            default=F('salary'),
        )
    ).update(salary=F('bonus_salary'))

    await bob.refresh_from_db()
    print(f"NEW SALARY:", bob.salary)  # Bob's salary is now 110% of the original value

run_async(main())
```

#### Traceback
```python-traceback
Traceback (most recent call last):
  File "/root/.local/lib/python3.12/site-packages/tortoise/expressions.py", line 53, in resolver_arithmetic_expression
    arithmetic_expression_or_field.name = model._meta.fields_db_projection[name]
                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'bonus_salary'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/root/tortoise_bug/test.py", line 66, in <module>
    run_async(main())
  File "/root/.local/lib/python3.12/site-packages/tortoise/__init__.py", line 624, in run_async
    loop.run_until_complete(coro)
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/root/tortoise_bug/test.py", line 34, in main
[...]

Issue #1496, opened 2023-10-14
Support annotate F() expression cross model relation


YAGregor:
**Is your feature request related to a problem? Please describe.**

to reproduce:
```
import asyncio

from tortoise import Model, fields, Tortoise
from tortoise.expressions import F


async def test():
    return ""


class A(Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(max_length=255, default="name_of_a")


class B(Model):
    id = fields.IntField(pk=True)
    a = fields.ForeignKeyField("models.A")
    name = fields.CharField(max_length=255, default="name_of_b")


async def main():
    await Tortoise.init(db_url="sqlite://:memory:", modules={"models": ["__main__"]})
    await Tortoise.generate_schemas()

    a = await A.create()
    b = await B.create(a=a)

    b2 = await B.annotate(a_name=F("a__name"), ya_b_name=F("name")).get(id=1)

    print(b2.a_name, b2.ya_b_name)  # in django orm we got b2.a_name == "name_of_a", but tortoise treat F(a__name") as string seems that F only works on fields in queryset's model its self

    await Tortoise.close_connections()


if __name__ == '__main__':
    asyncio.run(main())

```

**Describe the solution you'd like**

F() be able to reference fields of foreign key's model

**Describe alternatives you've considered**

**Additional context**
Add any other context about the feature request here.

YAGregor:
btw, I don't think F extend pypika's field is a good idea,I tried to implement this feature but seem its hard to extend feature.
Is your code struct imitate Django ORM? I suggest to organize query like tree, and "dump" the tree recursively just like [`peewee` ](https://github.com/coleifer/peewee/blob/01b2d94a4029858b9d2b1bee6fac40eed27ad9ad/peewee.py#L715)

xalteropsx:
> btw, I don't think F extend pypika's field is a good idea,I tried to implement this feature but seem its hard to extend feature. Is your code struct imitate Django ORM? I suggest to organize query like tree, and "dump" the tree recursively just like [`peewee` ](https://github.com/coleifer/peewee/blob/01b2d94a4029858b9d2b1bee6fac40eed27ad9ad/peewee.py#L715)

```py

@router.post("/module-group/", status_code=status.HTTP_201_CREATED)
def create_group(module: scmod, group: sc_mod1, database: Session = Depends(get_db)):
    if not any(module.dict().values()):
[...]

Issue #995, opened 2021-12-01
Use F expression in Annotation


Arteemoon:
```python
await User.annotate(
        tournament_count=Count('tournaments'),
        cc = F('tournament_count') / 2
    )
```

in console always get  -> "tortoise.exceptions.OperationalError: column "tournament_count" does not exist"
version 0.17.8

divadsn:
Duplicate of #683

Issue #683, opened 2021-03-27
Multiple annotations don't work


ArtemBernatskyy:
```python
old_lotteries_pool = (
    await Lottery.filter(Q(status=LotteryStatus.ENDED) & Q(has_winners=False))
    .prefetch_related("tickets")
    .annotate(total_tickets=Count("tickets"))
    .annotate(lottery_pool=F("total_tickets") * F("ticket_price"))
)
```

I receive `tortoise.exceptions.OperationalError: column "total_tickets" does not exist` (in Django we could use multiple annotations)

PS I know it could be `RTFM` but I've tried searching issues / reading docs but without success, thx in advance for any help!
PPS here is models that I use

```python
from tortoise import fields
from tortoise.models import Model

from app.constants import LotteryStatus
from app.validators import PositiveValueValidator


class User(Model):
    """User table"""

    id = fields.BigIntField(pk=True)  # same as https://discordpy.readthedocs.io/en/latest/api.html#discord.User.id
    balance = fields.data.DecimalField(
        max_digits=15, decimal_places=2, default=0, validators=[PositiveValueValidator()]
    )
    created_at = fields.DatetimeField(auto_now_add=True)
    modified_at = fields.DatetimeField(auto_now=True)

    lotteries: fields.ReverseRelation["Lottery"]
    tickets: fields.ReverseRelation["Ticket"]

    def __str__(self):
        return f"#{self.id}"


class Lottery(Model):
    """Lottery table"""

    id = fields.UUIDField(pk=True)
    name = fields.CharField(max_length=255, unique=True)
    ticket_price = fields.data.DecimalField(max_digits=15, decimal_places=2, default=10)
    strike_date_eta = fields.data.DatetimeField()
    strike_eth_block = fields.IntField()
    winning_tickets = fields.JSONField(null=True)
    # there is a possibility when lottery doesn't have winners, in this case we will add lottery pool to the total pool
    # then if someone wins another lottery he will get total pool
    has_winners = fields.BooleanField(default=False)
    status = fields.CharEnumField(
        enum_type=LotteryStatus,
        default=LotteryStatus.STARTED,
    )
    participants = fields.ManyToManyField("app.User", related_name="lotteries", through="ticket")
    created_at = fields.DatetimeField(auto_now_add=True)
    modified_at = fields.DatetimeField(auto_now=True)

    tickets: fields.ReverseRelation["Ticket"]

    def __str__(self):
        return self.name


class Ticket(Model):
[...]

PR #1761, opened 2024-11-08
Support annotations and joins in F()

henadzit:
## Description

This PR:
- Adds support for referencing relationship fields in `F()`, e.g. `F("author__name")` that automatically introduces joins to the query
- Allows referencing annotations in `F()`, e.g. `IntFields.annotate(intnum_plus_1=F("intnum") + 1).annotate(intnum_plus_2=F("intnum_plus_1") + 1)`

In order to achieve that, the following implementation details changed
- The behavior of `Expression.resolve` was unified to always return a `ResolveResult`. Previously it could have been either a tuple or dictionary.
- `F` was changed to be an `Expression` instead of pypika's `Field`. This allowed to move the `F`'s resolution code inside `F.resolve` and simplify the resolution code in general because `F` can be treated as other `Expression`s.
- `CombinedExpression` was introduced to handle arithmetic operations on `F()`, e.g. `F("a") + 1`. It was inspired by Django's [CombinedExpression](https://github.com/django/django/blob/042b381e2e37c0c37b8a8f6cc9947f1a2ebfa0dd/django/db/models/expressions.py#L702) and replaces pypika's arithmetic. `CombinedExpression` is a child of `Expression`, so it can be resolved without knowing the exact type of the object as opposed to pypika's `ArithmeticExpression` that was used previosly.

As part of this PR, I had to change how custom functions can be defined. It changed from

```
from tortoise.expressions import F
from pypika.terms import Function

class JsonSet(Function):
    def __init__(self, field: F, expression: str, value: Any):
        super().__init__("JSON_SET", field, expression, value)
```

to

```
from tortoise.functions import Function
from pypika.terms import Function as PupikaFunction

# it is a tortoise Function now, not Pypika's Function
class PypikaJsonSet(Function):
    def __init__(self, field: F, expression: str, value: Any):
        super().__init__("JSON_SET", field, expression, value)
    
    database_func = PypikaJsonSet
```

Effectively now custom functions are supposed to be a tortoise `Function`, not a Pypika's `Function`. The change had to be done because the Pypika's `Function` is no longer compatible with `F` and, frankly, having the interface that accepts both Pypika's functions and tortoise functions is a bit messy and makes the code quite convoluted.

## Motivation and Context

This introduces functionality supported by Django but also there is a clear demand for these features in the tortoise community. The following issues should be fixed by this PR
- fix https://github.com/tortoise/tortoise-orm/issues/1703
- fix https://github.com/tortoise/tortoise-orm/issues/683
- fix https://github.com/tortoise/tortoise-orm/issues/995
- fix https://github.com/tortoise/tortoise-orm/issues/477
- fix https://github.com/tortoise/tortoise-orm/issues/1496

## How Has This Been Tested?
- Tested on a sample web app
- Added tests

## Checklist:

### DIFF
--- a/tortoise/expressions.py
+++ b/tortoise/expressions.py
@@ -123,10 +123,19 @@ class F(Expression):
         joins: List[TableCriterionTuple] = []
         output_field = None
         if self.name.split("__")[0] in resolve_context.model._meta.fetch_fields:
+            # field in the format of "related_field__field" or "related_field__another_rel_field__field"
             term, joins, output_field = resolve_nested_field(
                 resolve_context.model, resolve_context.table, self.name
             )
+        elif self.name in resolve_context.annotations:
+            # reference to another annotation, e.g. M.annotate(f1=...).annotate(f2=F("f1")).values('field')
+            annotation = resolve_context.annotations[self.name]
+            if isinstance(annotation, Term):
+                term = annotation
+            else:
+                term = annotation.resolve(resolve_context).term
         else:
+            # a regular model field, e.g. F("id")
             try:
                 term.name = resolve_context.model._meta.fields_db_projection[self.name]
 
--- a/tortoise/expressions.py
+++ b/tortoise/expressions.py
@@ -1,6 +1,7 @@
-from enum import Enum, auto
 import operator
-from dataclasses import dataclass, field as dataclass_field
+from dataclasses import dataclass
+from dataclasses import field as dataclass_field
+from enum import Enum, auto
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -23,8 +24,8 @@ from pypika.terms import Function as PypikaFunction
 from pypika.terms import Term
 from pypika.utils import format_alias_sql
 
-from tortoise.fields.base import Field
 from tortoise.exceptions import ConfigurationError, FieldError, OperationalError
+from tortoise.fields.base import Field
 from tortoise.fields.relational import (
     BackwardFKRelation,
     ForeignKeyFieldInstance,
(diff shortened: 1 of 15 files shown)


tortoise/expressions.py
```diff
@@ -498,19 +552,28 @@ def resolve(self, resolve_context: ResolveContext) -> dict:
...
+                joins=function_arg.joins,
+                output_field=function_arg.output_field,  # type: ignore
+            )
+        else:
+            function_arg = self.field.resolve(resolve_context)
+            term = self._get_function_field(function_arg.term, *default_values)
+            res = ResolveResult(
+                term=term,
+                joins=function_arg.joins,
+                output_field=function_arg.output_field,  # type: ignore
```
abondar, line 570: Why you had to use so many type ignores?

abondar (approved):

tortoise/expressions.py
henadzit: Great catch! This is something I never ran into before. Basically the `Field` class has [the following type checking code](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/fields/base.py#L162):

```
    # These methods are just to make IDE/Linters happy:
    if TYPE_CHECKING:

        def __new__(cls, *args: Any, **kwargs: Any) -> "Field[VALUE]":
            return super().__new__(cls)

        @overload
        def __get__(self, instance: None, owner: Type["Model"]) -> "Field[VALUE]": ...

        @overload
        def __get__(self, instance: "Model", owner: Type["Model"]) -> VALUE: ...

        def __get__(
            self, instance: Optional["Model"], owner: Type["Model"]
        ) -> "Field[VALUE] | VALUE": ...

        def __set__(self, instance: "Model", value: VALUE) -> None: ...
```

that makes it possible to have type hints for both the `Model` classes but also for `Model` instances. However, this seems to limit who can have the field attribute to only `Model` and its subclasses because of `instance` and `owner` arg typing. 

I added the `Field` property to `ResolveResult` to be able to use `Field` for type conversions and this is what caused warnings that are being ignored.

I tried to add overloaded versions of `__get__` to `Field` but this looks ugly because `Field` has to know about the `ResolveResult`:

```
@overload
        def __get__(self, instance: None, owner: Type["Model"]) -> "Field[VALUE]": ...

        @overload
        def __get__(self, instance: "Model", owner: Type["Model"]) -> VALUE: ...

        @overload
        def __get__(
            self, instance: ResolveResult, owner: Type["ResolveResult"]
        ) -> "Field[VALUE]": ...

        def __get__(
            self,
            instance: Union[Optional["Model"], ResolveResult],
            owner: Union[Type["Model"], Type[ResolveResult]],
        ) -> "Field[VALUE] | VALUE": ...
```

Please let me know if you have ideas on how to handle this better, thanks!

PR #1763, opened 2024-11-09
Make JSONField support type annotation and OpanAPI document generation

LanceMoe:
Make JSONField support type annotation and OpanAPI document generation

## Description
JSONField is defined as Any, which does not display properly when using FastAPI to automatically generate OpenAPI documents.
This PR did two things:
- Manually specifying `field_type` enabled the document to be generated properly.
- Python generics allow for better type hints in IDEs (Pylance and Mypy).

## Motivation and Context
I have been using FastAPI + tortoise-orm as the preferred solution for all my project.
My project front-end uses TypeScript + React, and uses `openapi-typescript` to automatically convert the OpenAPI document generated by FastAPI into a type that can be used by TypeScript. However, the JSONField of tortoise-orm does not support type hints, which requires a lot of type cast in my front-end code, so I tried to add type support to JSONField.

## How Has This Been Tested?
In my project, it works correctly, so I'm going to submit a PR to see if it helps with tortoise-orm. But I'm not sure if it will pass the existing tests. I'd like to add as many tests as possible, and hope to get help from the project maintainers.

Screenshots of my project:
image
image
image

## Checklist:


LanceMoe:
I built a local test environment and changed some code to ensure that the change was compatible with the previous version.
I also added relevant tests and documentation.

image

LanceMoe:
I fixed some code so that it now should pass mypy check (`make lint`).

image

abondar:
@LanceMoe Hi!

Please check out CI - seems like tests are failing

LanceMoe:
@abondar 
Thanks for your reply, I had only tested it in python 3.10 before and it worked fine. I just set up the test and it does fail in python 3.8 and 3.9, the reason is because staticmethod is not considered callable before python 3.10.
Now I fixed the test and it should work in older versions of python now.
Please try running CI again, thanks!

image

but Codacy Static Code Analysis throw a problem, the lambda is unnecessary on python >= 3.10, but is necessary for python <= 3.9

image

So I had to move it outside class

image

abondar (approved):

PR #1764, opened 2024-11-09
Enhance atomic decorator to inject transaction client for DB operations

zinzinday:
Description

Enhanced the `atomic` decorator to inject a transaction client (`BaseDBAsyncClient`) into wrapped functions. Updated function signatures to accept `client` as a parameter, allowing transaction management by specifying `using_db=client` in database operations. Improved naming convention from `session` to `client` to ensure semantic clarity.

## Motivation and Context

This change is necessary to ensure all queries within a function wrapped by the `atomic` decorator use the same transaction context for consistency. The modification solves potential issues with transaction integrity by making sure that all database operations within the decorator's scope are managed by a single `client` instance. This also aligns naming with the actual type and purpose of the injected object, improving code readability and maintainability.

## How Has This Been Tested?

Tested in a local development environment:
- Verified that all database operations within `@atomic`-decorated functions use the injected `client` consistently.
- Conducted tests to confirm transaction rollback on error and commit on successful completion.
- Checked function compatibility when `client` is provided as an argument, ensuring all database operations reference the same `client` for a unified transaction.

## Checklist:

### DIFF
--- a/tortoise/transactions.py
+++ b/tortoise/transactions.py
@@ -53,8 +53,8 @@ def atomic(connection_name: Optional[str] = None) -> Callable[[F], F]:
     def wrapper(func: F) -> F:
         @wraps(func)
         async def wrapped(*args, **kwargs):
-            async with in_transaction(connection_name):
-                return await func(*args, **kwargs)
+            async with in_transaction(connection_name) as client:
+                return await func(client, *args, **kwargs)
 
         return cast(F, wrapped)
 


abondar:
Sorry, I don't understand what issue you are solving

If you need to get transaction object in your function you are decorating - you can probably use [this function](https://github.com/tortoise/tortoise-orm/blob/9e392e40f5d05042f16a80c65e5eedf46bf48da8/tortoise/transactions.py#L14) to get it

If it is not what you want - please elaborate what issue you are fixing

zinzinday:
> Sorry, I don't understand what issue you are solving
> 
> If you need to get transaction object in your function you are decorating - you can probably use [this function](https://github.com/tortoise/tortoise-orm/blob/9e392e40f5d05042f16a80c65e5eedf46bf48da8/tortoise/transactions.py#L14) to get it
> 
> If it is not what you want - please elaborate what issue you are fixing



@abondar  Thank you for your response. Allow me to elaborate on the issue we’re addressing and why injecting the `client` (transaction object) directly is important in certain contexts.

### Issue & Purpose

In our use case, we need access to the specific transaction `client` (or `BaseDBAsyncClient`) inside the decorated function to ensure that **all database operations within this function use the same transaction context consistently**. This becomes essential when:

1. **Multiple queries are involved** within the function, and we need all these queries to be bound by the same transaction connection.
   
2. We want to **pass the same `client` explicitly to different model methods or query calls** (using `using_db=client`) so they’re guaranteed to use the same transaction context. Without this, there can be cases where Tortoise ORM might default to a different database connection or transaction context if `using_db` isn’t specified, potentially leading to inconsistency in transaction behavior.

### Why Direct `client` Injection?

- **Transactional Integrity**: By injecting `client`, we ensure all operations within the function share the same transaction, which is crucial for atomic operations. This helps us manage complex transaction scenarios (e.g., multiple writes and reads in the same transaction) where each operation needs to work with the same session.
  
- **Error Handling & Rollback**: If any operation fails within the function, all changes should be rolled back. Having the `client` allows us to ensure that the rollback will encompass every operation within this specific transaction context.

### Why Not Use `in_transaction()` Separately?

Using `in_transaction()` in each function independently would indeed provide a transaction context, but it introduces redundancy and potential inconsistencies when:
- We have many functions with the same structure needing this transaction setup.
- We require explicit control over the `client` to be passed through several levels of function calls without manually wrapping each.

### Summary

In summary, injecting the `client` simplifies transaction management by ensuring that all queries in a decorated function operate within the same transaction context. This solution not only promotes consistency but also enhances code readability and maintainability for cases requiring multiple transaction-bound operations.

abondar:
Sorry, but this is quite serious API break, that will affect all users, and will give any kind of advantage only to limited pool of users. 

As I said above - there are other methods to explicitly get current transaction object and you can use them, to achieve what you are going for, without breaking current api

Issue #1181, opened 2022-07-14
How to set or change table name dynamiclly?


smil4ever:
I can most probably fix this myself with some kind of hacky method that will override the table name generation, but instead it would be cool to have this feature in orm itself where you can give the option to prefix all the tables with a prefix given in a configuration.

Issue #1026, opened 2021-12-30
Allow function to be passed to `table`


Kyzegs:
**Is your feature request related to a problem? Please describe.**
I'm connecting my Python application to Laravel, a web framework. Laravel generates its database tables with underscores, but Tortoise doesn't. I've looked through the documentation and was unable to find a solution. You could always specify the table in each model, but it'll get pretty annoying when you have hundreds of them.

**Describe the solution you'd like**
It would be awesome if we could pass a function to `table` in the `Meta` class. [peewee has this functionality](http://docs.peewee-orm.com/en/latest/peewee/models.html?highlight=table%20generation#table-names). Adopting something like this should make the ORM more compatible with outside sources.

**Describe alternatives you've considered**
I've tried the mentioned solution. Currently, I'm just explicitly setting the table in each of my models.

shanebdavis:
I agree. I specifically want snake_case_table_names, but in general, it would be nice to establish a base-class with a function for generating table-names from class names.

henadzit:
Added `table_name_generator` argument to `Tortoise.init`.

PR #1770, opened 2024-11-14
Add `table_name_generator` attribute to Meta for dynamic table name generation

Abdeldjalil-H:
Add Table Name Generator Feature

## Description

This PR adds a new `table_name_generator` attribute to the Model Meta class. This allows users to dynamically generate table names using a custom function instead of explicitly setting the table name or relying on defaults.

The feature includes:
- New `table_name_generator` Meta attribute that accepts a callable
- Generator function receives the model class as an argument and returns the table name
- Maintains precedence where explicit `table` name overrides generated names
- Comprehensive test suite and documentation

## Motivation and Context

This change provides users with more flexibility in how table names are generated. This should solve #1026 #1181

The feature promotes cleaner code by centralizing table naming logic and reducing repetition across models.

## How Has This Been Tested?

The implementation includes a comprehensive test suite that verifies:
1. Basic table name generation functionality
2. Precedence of explicit table names over generated names
3. Default behavior when no generator is provided

Tests are implemented using the existing tortoise.contrib.test.TestCase class and follow the project's testing patterns. All tests pass in the current test environment.

Test scenarios cover:
- Models with table name generator
- Models with both explicit table name and generator
- Models without any table name configuration

## Checklist:

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -197,7 +197,8 @@ The ``Meta`` class
 
         Set this to a callable that will generate the table name for the model.
         The callable should accept the model class as its argument and return a string.
-        If both ``table`` and ``table_name_generator`` are specified, ``table`` takes precedence.
+        Cannot be used together with ``table`` attribute - attempting to set both will raise 
+        a `ConfigurationError`.
 
         .. code-block:: python3
 
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -192,6 +192,21 @@ The ``Meta`` class
 
         Set this to configure a manual table name, instead of a generated one
 
+    .. attribute:: table_name_generator
+        :annotation: = None
+
+        Set this to a callable that will generate the table name for the model.
+        The callable should accept the model class as its argument and return a string.
+        If both ``table`` and ``table_name_generator`` are specified, ``table`` takes precedence.
+
+        .. code-block:: python3
+
+            def generate_table_name(model_cls):
+                return f"prefix_{model_cls.__name__.lower()}"
+
+            class Meta:
+                table_name_generator = generate_table_name
+
     .. attribute:: table_description
         :annotation: = ""
 
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -192,22 +192,6 @@ The ``Meta`` class
 
         Set this to configure a manual table name, instead of a generated one
 
-    .. attribute:: table_name_generator
-        :annotation: = None
-
-        Set this to a callable that will generate the table name for the model.
-        The callable should accept the model class as its argument and return a string.
-        Cannot be used together with ``table`` attribute - attempting to set both will raise 
-        a `ConfigurationError`.
-
-        .. code-block:: python3
-
-            def generate_table_name(model_cls):
-                return f"prefix_{model_cls.__name__.lower()}"
-
-            class Meta:
-                table_name_generator = generate_table_name
-
     .. attribute:: table_description
         :annotation: = ""
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,7 @@ Added
 ^^^^^
 - Add POSIX Regex support for PostgreSQL and MySQL (#1714)
 - support app=None for tortoise.contrib.fastapi.RegisterTortoise (#1733)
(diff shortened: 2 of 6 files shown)


docs/models.rst
```diff
@@ -192,6 +192,21 @@ The ``Meta`` class
 
         Set this to configure a manual table name, instead of a generated one
 
+    .. attribute:: table_name_generator
+        :annotation: = None
+
+        Set this to a callable that will generate the table name for the model.
+        The callable should accept the model class as its argument and return a string.
+        If both ``table`` and ``table_name_generator`` are specified, ``table`` takes precedence.
```
abondar, line 200: That seems confusing to me, and it can lead to confusing behaviour 

May be it would be better to raise configuration error in case both are specified?

Abdeldjalil-H: I thought about that, but imagine if you have an abstract model that all other models inherit from it, and for some table you want to use a custom name, how would you achieve this?

abondar: I may be missing something, but I am pretty sure meta is not inherited 

```


class TournamentAbs(Model):
    name = fields.TextField()

    events: fields.ReverseRelation["Event"]

    class Meta:
        abstract = True
        table = "tournament1"
        table_description = "ttt"

    def __str__(self):
        return self.name


class Tournament(TournamentAbs):
    pass


async def run():

    print(Tournament._meta.db_table)
    print(Tournament._meta.table_description or "None")


if __name__ == "__main__":
    run_async(run())
```

Gives output

```
tournament
None
```

If it was inherited - than `abstract` attribute would be inherited too

abondar:
I am okay with merging it that way - but is it really what users in original issue needed? 

Probably would be a little bit easier for such case, but probably they wanted to have ability to override table name generation globally 

Although not sure how to do it in clean way - may be make some base method to determine table, and it would by default take table from meta, but users could override it? 
Not a super fan of that idea, as it would go against current logic that such thing determined by model meta, but not sure of better ideas right now

abondar:
I think that could work

Although then we would be back to problem of conflict between `Meta.table_name` and this init function, which could be confusing

@henadzit @waketzheng may be you have any good ideas how interface for that could look like?

Abdeldjalil-H:
I think in that case it is not much confusing. The `Meta.table` will take precidence. It is like overriding an existing value.

waketzheng:
Also, we can add an example to show generating snake_case_table_names global.

abondar:
@Abdeldjalil-H are you planning to work on that in near future?

I was thinking if I should wait for it to be ready to include in 0.22.0 release

henadzit:
Hi! Sorry, I'm a bit late here.

A few thoughts/questions from me:
- @Abdeldjalil-H is this a compatible change with aerich? 
- Is there similar functionality in other ORMs?
- I'm not sure that I like that `table_name_generator` is global. What if you have two different groups of models and you want to have different naming pattern for them? Looks like it's possible to achieve that with a Meta class in Django https://stackoverflow.com/questions/51353106/is-there-a-django-setting-to-generate-multiple-words-table-names-for-multiplewor. Maybe, it's something we can explore?

Abdeldjalil-H:
@henadzit I'll check for aerich.
The solution you proposed definitely works, but you need to inherit from `PrefixMeta` on each all models. Of course this is better than defining custom table name for all tables, but I think we can do better. Your solusion inspired me about the following solution

```python
from tortoise.models import Model, ModelMeta


class PrefixTableNameMeta(ModelMeta):
    def __new__(mcs, name: str, *args):
        new_class = super().__new__(mcs, name, *args)
        new_class._meta.db_table = "prefix_" + name.lower()
        return new_class
        
class User(Model, metaclass=PrefixTableNameMeta):
    id = fields.IntField(pk=True)
```

Will this solution will work without the need of any changes in the original code base, it is not user-friendly. I think the simplest solution we can get is to define something that has the following criterias:
1. Easy to define and use by all users.
2. Easily inheritable/overridable
3. Minimum repetition in the code

One example that satisfies this is an optional function on the model itself (not meta). We can call it `get_table_name` or anything.

henadzit (approved):
@Abdeldjalil-H thank you for the great discussion! After thinking about this a bit more, `table_name_generator` seems to be the best solution, it's simple and flexible.

abondar (approved):

PR #1776, opened 2024-11-18
fix: enums not quoted

### LINKED ISSUE #1774: Enums not quoted (bug of pypika-torotise)
**Describe the bug**
Reported here: https://github.com/tortoise/pypika-tortoise/issues/7

**To Reproduce**
```py
from tortoise import Model, fields, run_async
from tortoise.fields.base import StrEnum


class MyModel(Model):
    id = fields.IntField(pk=True)
    name = fields.TextField()


class MyEnum(StrEnum):
    A = "a"


@init_memory_sqlite
async def do():
    await MyModel.create(name="a")
    qs = MyModel.filter(name=MyEnum.A)
    qs._make_query()
    print(qs.query)
    # expected   SELECT "id","name" FROM "mymodel" WHERE "name"='a'
    # actual     SELECT "id","name" FROM "mymodel" WHERE "name"=a
    print(await MyModel.filter(name=MyEnum.A).exists())
    # expected True
    # actual raises tortoise.exceptions.OperationalError: no such column: a


run_async(do())
```

**Expected behavior**
Support filter by enum

**Additional context**
Will be fixed after this PR(https://github.com/tortoise/pypika-tortoise/pull/10) merged.

waketzheng:
## Description

Fixes #1774
Depends on https://github.com/tortoise/pypika-tortoise/pull/10

## Motivation and Context

Create this PR to verify the latest commit of pypika-tortoise

## How Has This Been Tested?

make ci

## Checklist:


tests/test_filters.py
```diff
@@ -29,6 +39,14 @@ async def test_equal(self):
             set(await CharFields.filter(char="moo").values_list("char", flat=True)), {"moo"}
         )
 
+    async def test_enum(self):
+        self.assertEqual(
+            set(await CharFields.filter(char=MyEnum.moo).values_list("char", flat=True)), {"moo"}
```
henadzit, line 44: I don't think this is a correct test case. `char` is a str field, why should it be possible to search by an Enum.

tests/test_q.py
```diff
@@ -119,10 +119,16 @@ class TestQCall(TestCase):
...
-            model=IntFields, table=IntFields._meta.basequery, annotations={}, custom_filters={}
+            model=IntFields,
+            table=IntFields._meta.basequery,  # type:ignore[arg-type]
+            annotations={},
+            custom_filters={},
         )
         self.char_fields_context = ResolveContext(
-            model=CharFields, table=CharFields._meta.basequery, annotations={}, custom_filters={}
+            model=CharFields,
+            table=CharFields._meta.basequery,  # type:ignore[arg-type]
```
henadzit, line 129: Why are there so many ignores added in this PR?

henadzit: I'm curious why these weren't an issue before and they are now. Have you looked into that?

waketzheng: Because I add the `py.typed` file to pypika-tortoise

tests/test_filters.py
waketzheng: I try it with Django:
```bash
mkdir something
cd something
python3.11 -m venv venv
source venv/bin/activate
pip install django ipython
python -m django startproject something .
python manage.py startapp foo
echo '
from enum import Enum, StrEnum
from django.db import models

class RawEnum(Enum):
    moo = "moo"

class MyEnum(str, Enum):
    moo = "moo"

class MyStrEnum(StrEnum):
    moo = "moo"

class MyModel(models.Model):
    name = models.CharField(max_length=10)
' > foo/models.py
python -c 's="INSTALLED_APPS = [\n";m="foo";from pathlib import Path;p=Path("something/settings.py");p.write_text(p.read_text().replace(s,s.strip()+repr(m)+",\n"))'
python manage.py makemigrations
python manage.py migrate
python manage.py shell
```
Got:
```
In [1]: from foo.models import *

In [2]: MyModel.objects.create(name='moo')
Out[2]: <MyModel: MyModel object (1)>

In [3]: MyModel.objects.filter(name=RawEnum.moo)
Out[3]: <QuerySet []>

In [4]: MyModel.objects.filter(name=MyEnum.moo)
Out[4]: <QuerySet [<MyModel: MyModel object (1)>]>

In [5]: MyModel.objects.filter(name=MyStrEnum.moo)
Out[5]: <QuerySet [<MyModel: MyModel object (1)>]>
```
I think we can follow the strategy of django:
1. Return empty queryset if the enum object only inherit from Enum (result of `.filter(name=str(enum_obj))`)
2. Use its value to query if the enum object inherit from str and Enum

henadzit: Does Django have a test case or documentation explicitly stating that they support filtering `CharField` with enums? I'm wondering if the ability to find records in your examples is just a weird coincidence.

If django doesn't explicitly support it, I don't think we should either. I think in the ideal world `MyModel.objects.filter(name=RawEnum.moo)` should cause a type error because you are trying to filter by incompatible types. Having a test case for this kinda make this behavior official which I don't think we should be doing. I would suggest making this PR not about searching by enum but about updating pypika-tortoise version.

henadzit: Yeah, that's a good point, I agree!

pyproject.toml
```diff
@@ -36,7 +36,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.8"
-pypika-tortoise = "^0.2.1"
+pypika-tortoise = { git = "https://github.com/waketzheng/pypika-tortoise.git", branch = "fix-enums-not-quoted" }
```
henadzit, line 39: We need to release a new pypika version before we can merge it. @abondar, can you please do it or let me know how we can do it!

abondar: 0.2.2 seems to be released already
https://github.com/tortoise/pypika-tortoise/releases/tag/v0.2.2

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tortoise-orm"
-version = "0.21.7"
```
abondar, line 3: Please don't change it for now - I am planning to release it as 0.22.0 - after PR with table name generation

abondar (dismissed):

henadzit:
@waketzheng no need to change about this PR but it would be great to have two separate PRs in the future when you want to refactor the code. In this example it would be:
- PR for fixing enums not quoted
- Adding type hints

Having everything as a single PR makes it really hard to review actual logic changes. Thanks!

waketzheng:
> @waketzheng no need to change about this PR but it would be great to have two separate PRs in the future when you want to refactor the code. In this example it would be:
> 
> * PR for fixing enums not quoted
> * Adding type hints
> 
> Having everything as a single PR makes it really hard to review actual logic changes. Thanks!

OK~

waketzheng:
@abondar Cloud you merge it

abondar (approved):

PR #1779, opened 2024-11-20
refactor: improve type hints

waketzheng:
## Description

Improve type hints
## Checklist:

### DIFF
--- a/tortoise/__init__.py
+++ b/tortoise/__init__.py
@@ -488,25 +488,26 @@ class Tortoise:
 
         if config_file:
             config = cls._get_config_from_config_file(config_file)
-
-        if db_url:
+        elif db_url:
             if not modules:
                 raise ConfigurationError('You must specify "db_url" and "modules" together')
             config = generate_config(db_url, modules)
+        else:
+            assert config is not None  # To improve type hints
 
         try:
-            connections_config = config["connections"]  # type: ignore
+            connections_config = config["connections"]
         except KeyError:
             raise ConfigurationError('Config must define "connections" section')
 
         try:
-            apps_config = config["apps"]  # type: ignore
+            apps_config = config["apps"]
         except KeyError:
             raise ConfigurationError('Config must define "apps" section')
 
-        use_tz = config.get("use_tz", use_tz)  # type: ignore
-        timezone = config.get("timezone", timezone)  # type: ignore
-        routers = config.get("routers", routers)  # type: ignore
+        use_tz = config.get("use_tz", use_tz)
+        timezone = config.get("timezone", timezone)
+        routers = config.get("routers", routers)
 
         cls.table_name_generator = table_name_generator
 
--- a/tortoise/expressions.py
+++ b/tortoise/expressions.py
@@ -4,18 +4,7 @@ import operator
 from dataclasses import dataclass
 from dataclasses import field as dataclass_field
 from enum import Enum, auto
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    Iterator,
-    List,
-    Optional,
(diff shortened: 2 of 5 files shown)


tortoise/expressions.py
```diff
@@ -158,7 +147,7 @@ def resolve(self, resolve_context: ResolveContext) -> ResolveResult:
             except KeyError:
                 raise FieldError(
                     f"There is no non-virtual field {self.name} on Model {resolve_context.model.__name__}"
-                )
+                ) from None
```
henadzit, line 150: What is it for?

henadzit (commented):
A lot of nice improvements! I left a few minor comments.

tortoise/__init__.py
```diff
@@ -488,25 +488,26 @@ async def init(
 
         if config_file:
             config = cls._get_config_from_config_file(config_file)
-
-        if db_url:
+        elif db_url:
             if not modules:
                 raise ConfigurationError('You must specify "db_url" and "modules" together')
             config = generate_config(db_url, modules)
+        else:
```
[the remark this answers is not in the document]
waketzheng: > I think we should just move the `ConfigurationError` here:
> 
> ```python
> elif not config:
>   raise ConfigurationError(
>                 'You should init either from "config", "config_file" or "db_url"'
>             )
> ```
> 
> I think this should fix the typing issue.

In that case, if `int(bool(config) + bool(config_file) + bool(db_url)) > 1` will not raise ConfigurationError

Or do you mean update code like this:
```
-        if int(bool(config) + bool(config_file) + bool(db_url)) != 1:
-            raise ConfigurationError(
-                'You should init either from "config", "config_file" or "db_url"'
-            )

-        if config_file:
+        if config_file and not config and not db_url:
             config = cls._get_config_from_config_file(config_file)
-        elif db_url:
+        elif db_url and not config and not config_file:
             if not modules:
                 raise ConfigurationError('You must specify "db_url" and "modules" together')
             config = generate_config(db_url, modules)
-        else:
-            assert config is not None  # To improve type hints
+        elif not config or int(bool(config) + bool(config_file) + bool(db_url)) > 1:
+            raise ConfigurationError(
+                'You should init either from "config", "config_file" or "db_url"'
+            )
```

tortoise/expressions.py
waketzheng: Add `from None` will hide the KeyError message. Now that we should the FieldError, there is no need to show the previous KeyError info.

For Example:
- demo.py
```
class FieldError(Exception):
    pass

try:
    {}["a"]
except KeyError:
    raise FieldError("There is no non-virtual")
```
print:
```
  File "demo.py", line 6, in <module>
    {}["a"]
KeyError: 'a'

During handling of the above exception, another exception occurred:

  File "demo.py", line 8, in <module>
    raise FieldError("There is no non-virtual")
__main__.FieldError: There is no non-virtual
```
- demo_with_from_none.py
```py
class FieldError(Exception):
    pass

try:
    {}["a"]
except KeyError:
    raise FieldError("There is no non-virtual") from None
```
print:
```
  File "demo_with_from_none.py", line 8, in <module>
    raise FieldError("There is no non-virtual") from None
__main__.FieldError: There is no non-virtual
```
See more at: https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

henadzit: Got it, thank you for the explanation!

tortoise/__init__.py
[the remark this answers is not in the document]
henadzit: I think I misread the conditions of the first check, sorry for the confusion!

henadzit (approved):

PR #1781, opened 2024-11-21
feat: add mixin for int enum field custom

### LINKED ISSUE #966: Python Enum.IntEnum datatype do not fit on Tortoise.fields.IntEnumField when not int16
**Describe the bug**
Python Enum.IntEnum datatype do not fit on Tortoise.fields.IntEnumField when any element of enumeration is outside of int16 range. 

IMHO, this should not behave like that (or at least configurable), since Python object can handle values larger than that.

**To Reproduce**
```
import enum
import tortoise.fields as fields

class PROTOCOL_ID(enum.IntEnum):
     A = 10000
     B = 40000
     C = 70000

protocol_id = fields.IntEnumField(enum_type=PROTOCOL_ID)
```

Fails with

```
ConfigurationError: The valid range of IntEnumField's values is -32768..32767! 
```

**Expected behavior**
I expect that I could store values larger or smaller than int16 on that. Or atleast a `BigintEnumField`. and `SmallIntEnumField`.

IMHO the naming should reflect the regular Int type fields as;

- Small = 16 bits
- Default = 32bits
- Big = 64Bits

**Additional context**
If not possible, is there a workaround?  I do not want store that protocol static data into a database to increase the query complexity.

waketzheng:
## Description

As #966 said, IntEnumField inherit from SmallIntField, so it can not use a enum_type with value > 32767

## Motivation and Context

This PR add a IntEnumFieldMixin to make it easy to custom a new IntEnumField:
```py
from typing import Type, cast
from tortoise.fields.data import IntEnumFieldMixin, IntEnumType, IntField

class IntEnumInstance(IntEnumFieldMixin, IntField):
    pass

def IntEnumField(enum_type: Type[IntEnumType], **kwargs) -> IntEnumType:
    return cast(IntEnumType, IntEnumInstance(enum_type, **kwargs))
```

## How Has This Been Tested?

make ci

## Checklist:

### DIFF
--- a/tortoise/fields/data.py
+++ b/tortoise/fields/data.py
@@ -6,7 +6,7 @@ import json
 import warnings
 from decimal import Decimal
 from enum import Enum, IntEnum
-from typing import TYPE_CHECKING, Any, Callable, Type, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, Optional, Type, TypeVar, Union, cast
 from uuid import UUID, uuid4
 
 from pypika import functions
@@ -75,17 +75,19 @@ class IntField(Field[int], int):
     """
 
     SQL_TYPE = "INT"
-    VALUE_RANGE = (-2147483648, 2147483647)
     allows_generated = True
 
-    def __init__(self, primary_key: bool | None = None, **kwargs: Any) -> None:
+    def __init__(self, primary_key: Optional[bool] = None, **kwargs: Any) -> None:
         if primary_key or kwargs.get("pk"):
             kwargs["generated"] = bool(kwargs.get("generated", True))
         super().__init__(primary_key=primary_key, **kwargs)
 
     @property
     def constraints(self) -> dict:
-        return {"ge": self.VALUE_RANGE[0], "le": self.VALUE_RANGE[1]}
+        return {
+            "ge": -2147483648,
+            "le": 2147483647,
+        }
 
     class _db_postgres:
         GENERATED_SQL = "SERIAL NOT NULL PRIMARY KEY"
@@ -112,7 +114,13 @@ class BigIntField(IntField):
     """
 
     SQL_TYPE = "BIGINT"
-    VALUE_RANGE = (-9223372036854775808, 9223372036854775807)
+
+    @property
+    def constraints(self) -> dict:
+        return {
+            "ge": -9223372036854775808,
+            "le": 9223372036854775807,
+        }
 
     class _db_postgres:
         GENERATED_SQL = "BIGSERIAL NOT NULL PRIMARY KEY"
@@ -137,7 +145,13 @@ class SmallIntField(IntField):
     """
 
     SQL_TYPE = "SMALLINT"
-    VALUE_RANGE = (-32768, 32767)
+
+    @property
+    def constraints(self) -> dict:
+        return {
+            "ge": -32768,
+            "le": 32767,
+        }
 
     class _db_postgres:
         GENERATED_SQL = "SMALLSERIAL NOT NULL PRIMARY KEY"
@@ -178,7 +192,7 @@ class CharField(Field[str]):
         }
 
     @property
-    def SQL_TYPE(self) -> str:  # type:ignore[override]
+    def SQL_TYPE(self) -> str:  # type: ignore
         return f"VARCHAR({self.max_length})"
 
     class _db_oracle:
@@ -190,7 +204,7 @@ class CharField(Field[str]):
             return f"NVARCHAR2({self.field.max_length})"
 
 
-class TextField(Field[str], str):  # type:ignore[misc]
+class TextField(Field[str], str):  # type: ignore
     """
     Large Text field.
     """
@@ -200,7 +214,7 @@ class TextField(Field[str], str):  # type:ignore[misc]
 
     def __init__(
         self,
-        primary_key: bool | None = None,
+        primary_key: Optional[bool] = None,
         unique: bool = False,
         db_index: bool = False,
         **kwargs: Any,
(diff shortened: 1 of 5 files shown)


tortoise/fields/data.py
```diff
@@ -713,9 +692,51 @@ def to_db_value(
...
+
+class IntEnumFieldInstance(IntEnumFieldMixin, IntField):
+    pass
+
+
+class BigIntEnumFieldInstance(IntEnumFieldMixin, BigIntField):
+    pass
+
+
+def get_instance_class(
```
henadzit, line 707: It doesn't seem to me that choosing the field type dynamically is a good idea for the following reasons:
- At one moment you can have the value fitting a small int but then you extend your enum and you need to change the type of the column. The migration tools need to count for that, etc.
- This is an implicit behavior and in Python explicit is better than implicit. 
- This introduces new field types that we will need to support! Less things to support is better in general.

henadzit (changes requested):
What do you think about making `IntEnumFieldInstance` inheriting the regular `IntField` which will extend the potential values up to 2147483647? This should be enough for 99.9999% cases. If someone needs more, they can create a custom field.

I think we want to keep our Fields simple.

tortoise/fields/data.py
waketzheng: Ok, Let's keep the lib simple. I will remove the auto choosing function and add an example for how to extend IntEnumField to support int32 instead.

tortoise/fields/data.py
```diff
@@ -73,19 +75,17 @@ class IntField(Field[int], int):
     """
 
     SQL_TYPE = "INT"
+    VALUE_RANGE = (-2147483648, 2147483647)
```
henadzit, line 78: I don't think we should treat `VALUE_RANGE` in the same ways  as `SQL_TYPE`. `SQL_TYPE` is present for all fields but `VALUE_RANGE` just for integers. Having it as a `VALUE_RANGE` as a class field might give an impression that this is a part of the interface of all field classes but it isn't. It also makes it possible to get the constraints in two ways: through `VALUE_RANGE` and through `constraints` making it harder to know which is the right one to use. I would suggest to revert this change and keep constraints in the constraints method.

tortoise/fields/data.py
```diff
@@ -713,9 +707,13 @@ def to_db_value(
 IntEnumType = TypeVar("IntEnumType", bound=IntEnum)
 
 
+class IntEnumFieldInstance(IntEnumFieldMixin, SmallIntField):
+    pass
+
+
 def IntEnumField(
```
henadzit, line 714: This is really confusing that `IntEnumField` actually returns `SmallIntField` :(

henadzit (commented):
I still think we need to  make `IntEnumField` an integer field (which is going to require a migration for users unfortunately but this would be a correct way to do it), not a small integer field. This will cover 99.9% cases and we won't have to complicate the code and making extra documentation for a very rare edge case. For the people who need big integers, they can just copy the related code from `data.py` and make their own field.

waketzheng:
> I still think we need to make `IntEnumField` an integer field (which is going to require a migration for users unfortunately but this would be a correct way to do it), not a small integer field. This will cover 99.9% cases and we won't have to complicate the code and making extra documentation for a very rare edge case. For the people who need big integers, they can just copy the related code from `data.py` and make their own field.

I tried this one: 
```diff
-class IntEnumFieldInstance(SmallIntField):
+class IntEnumFieldInstance(IntField):
```
And then run `aerich migrate`, it generated a new migration file:
```py
from tortoise import BaseDBAsyncClient


async def upgrade(db: BaseDBAsyncClient) -> str:
    return """
        ALTER TABLE "users" ALTER COLUMN "gender" TYPE INT USING "gender"::INT;"""


async def downgrade(db: BaseDBAsyncClient) -> str:
    return """
        ALTER TABLE "users" ALTER COLUMN "gender" TYPE SMALLINT USING "gender"::SMALLINT;"""
```
----
1. db field type migrating is boring, especially for projects running in production
2. I strongly suggest to keep backward compatibility until the next major version

tortoise/fields/data.py
waketzheng: Yes, but mostly, IntEnumField used in the following case:
```py
class Gender(IntEnum):
      male = 0
      female = 1
      other = 2

class Users(Model):
    gender = fields.IntEnumField(Gender)
```
And I wish it to be `small int` in db, not int.

abondar:
I am more inclined to have default enum as smallint, as it is indeed covers most sane cases of using enums

Although maybe we should document it somehow better

henadzit:
Let's see how other ORMs implement this!

## Django
[In Django](https://docs.djangoproject.com/en/5.1/ref/models/fields/#enumeration-types) you choose the field, e.g. `CharField`, `SmallIntegerField`, `IntegerField`, etc. and then you specify `choices`. Basically it gives the user flexibility to choose the field type.

## SQLAlchemy
[SQLAlchemy](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Enum) "will make use of the backend’s native “ENUM” type if one is available; otherwise, it uses a VARCHAR datatype".

Native ENUMs are great but Django's approach is more universal and supported by more databases and I like it more personally. I think we can consider implementing something like that in Tortoise instead of the existing enum approach but not as part of this PR obviously.

I agree with Andrey that small int should cover most sane use cases of using enums. I don't like the idea of introducing mixins and making the fields classes and documentation more complicated for resolving some very peculiar use case. I think we can just update the docstring of `IntEnumField` to say that it is a small int. We can also post in the issue that provoked this PR that they can copy the `IntEnumFieldInstance` code reimplement it to support other data types.

PR #1783, opened 2024-11-22
Fix query error when model with o2o pk and m2m field

### LINKED ISSUE #1771: Unable to use ManyToManyField if OneToOneField passed as Primary Key
I don't know if its a issue or something else.

I have a table in which there is no id field i am using OneToOneField as primary key.
Everything working but when i use another field with ManyToManyField i am continuously getting this error 

**tortoise.exceptions.OperationalError: syntax error at or near "None".**

Somehow this looks like **ManyToManyField** is trying to override the **primary_key = True** of **OneToOneField**.

Below is my Tortoise ORM Model:
```
class UserProfile(Model):
    class Gender(Enum):
        MALE = 'Male'
        FEMALE = 'Female'
        OTHERS = 'Others'

        @classmethod
        def as_tuple(cls):
            return [(item.value, item.name.replace('_', ' ')) for item in cls]
    
    user: fields.OneToOneRelation[User] = fields.OneToOneField(
        'models.User', 
        on_delete=fields.CASCADE, 
        related_name='user_profile',
        primary_key=True,
    )
    address: fields.ManyToManyRelation[Address] = fields.ManyToManyField(
        'models.Address',
        related_name='user_profile_address',
        null = True,
    )
    gender = fields.CharEnumField(Gender, max_length=20, null=True)
```
Here is my Django ORM Model:
```
class UserProfile(models.Model):
    """
    Base Class for storing User Profile related information.
    This class hold one to one relation from auth user table
    """

    class Gender(Enum):
        MALE = 'Male'
        FEMALE = 'Female'
        OTHERS = 'Others'

        @classmethod
        def as_tuple(cls):
            return ((item.value, item.name.replace('_', ' ')) for item in cls)

    user = models.OneToOneField(User, primary_key=True,on_delete=models.CASCADE)
    address = models.ManyToManyField(Address, blank=True)
    gender = models.CharField(null=True, max_length=20, choices=Gender.as_tuple())
```

Please help i am not able to understand whats wrong, My problem is i cannot change anything at table schema level because everything is working fine in Django. If i remove the primary_key=True the OneToOneField and ManyToManyField works. But i am unable to save it because it says **tortoise.exceptions.OperationalError: column "id" of relation "user_profile" does not exist**

waketzheng:
## Description

Fixes #1771 

## Checklist:


tortoise/expressions.py
```diff
@@ -359,6 +359,8 @@ def _process_filter_kwarg(
             param = model._meta.get_filter(key)
 
         pk_db_field = model._meta.db_pk_column
+        if isinstance(model._meta.pk, OneToOneFieldInstance):
+            pk_db_field += "_id"
```
abondar, line 363: Seeems a bit strange, that `db_pk_column` does not in fact store db column and we have to patch it in place

Did you try changing the way `db_pk_column` is defined? 
Or something else breaks if you change it there?

waketzheng: Yes, this two lines are not readability. I will refactor later.

henadzit: I agree with Andrey, this looks like an issue which should be fixed in fields and models code, not in the expression code.

tortoise/models.py
```diff
@@ -614,7 +614,9 @@ def __search_for_field_attributes(base: Type, attrs: dict) -> None:
         meta.pk_attr = pk_attr
         meta.pk = fields_map.get(pk_attr)  # type: ignore
         if meta.pk:
-            meta.db_pk_column = meta.pk.source_field or meta.pk_attr
+            meta.db_pk_column = meta.pk.source_field or (
+                meta.pk_attr + "_id" * isinstance(meta.pk, OneToOneFieldInstance)
```
abondar, line 618: May be we should just do it like that?

```
if meta.pk.source:
    meta.db_pk_column = meta.pk.source_field
elif isinstance(meta.pk, OneToOneFieldInstance):
    meta.db_pk_column = f"{meta.pk_attr}_id"
else:
    meta.db_pk_column = meta.pk_attr
```

Yeah, it's a bit more verbose, but I think trying to squeeze 3 outcomes in one expression doesn't really look all that good and readable
```

PR #1788, opened 2024-11-25
fix: TypeError when sorting by a RawSQL

### LINKED ISSUE #1780: TypeError when sorting by a RawSQL
**Describe the bug**
When I have an annotation that's a subclass of the Pypika `Term` and I want to order by that field, it raises an error like `TypeError: Object of type RawSQL is not JSON serializable`.

**To Reproduce**
`Document.all().annotate(foo=RawSQL("'bar'")).order_by("foo").sql()` will raise `TypeError: Object of type RawSQL is not JSON serializable`

**Expected behavior**
It includes the correct code in the `ORDER BY` clause of the query.

**Additional context**
The actual source line for this is [here](https://github.com/tortoise/pypika-tortoise/blob/main/pypika/terms.py#L472-L473). The `value` var at that point looks something like `{'foo': <tortoise.expressions.RawSQL object at 0x32294cf50>}`.

I tried implementing my own subclass of Pypika's `Term` to fix but hit the same problem.

waketzheng:
## Description

Fixes #1780 

## Checklist:

### DIFF
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -575,6 +575,11 @@ class TestQueryset(test.TestCase):
         )
         self.assertEqual(data[0] + 1, data[1])
 
+    async def test_annotate_order_rawsql(self):
+        qs = IntFields.annotate(idp=RawSQL("id+1")).order_by("-idp")
+        data = await qs.first().values_list("id", "idp")
+        self.assertEqual(data[0] + 1, data[1])
+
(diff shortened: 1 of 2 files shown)


abondar (approved):

henadzit (approved):
Thank you for a quick fix! I think we also need to cut a new release since I can imagine it can break things for a lot of people!

abondar (approved):

Issue #1791, opened 2024-11-27
Foreign key input type is not in any way checked to actually represent the same model as in ForeignKey(model=...) parameter


HeroBrine1st:
**Describe the bug**

If fields.ForeignKeyField model parameter is not the same as in fields.ForeignKeyRelation, there's no help pointing at this inconsistency. It will work sometimes, sometimes it will fail 

It is deterministic but took me like 2 hours using various search engines and LLMs to actually spot that by sanity check, when I managed to minimise one passing test's preamble (creating initial objects in database) close enough to second, failing test.

It also should be said somewhere in the start of this issue that I tested this bug only on **SQLite**, using `aiosqlite`.

**To Reproduce**
Create empty folder. Create `testapp` folder in it.

Create file models.py:

```python
from tortoise import fields
from tortoise import Model

class TableA(Model):
    id = fields.IntField(primary_key=True)
    name = fields.CharField(max_length=255)

class TableB(Model):
    id = fields.IntField(primary_key=True)
    name = fields.CharField(max_length=255)

class TableC(Model):
    id = fields.IntField(primary_key=True)
    fkrelation: fields.ForeignKeyRelation[TableA] = fields.ForeignKeyField(
        # Look at it as it is a typo and it should be "models.TableA" like in type annotation
        # In my case it was an actual typo - I copied code and changed only related_name (and strangely changed to absolutely unrelated name (no pun intended), so I changed related_name again when I found that "model" parameter is incorrect).
        model_name="models.TableB",
        related_name="c_list"
    )
    name = fields.CharField(max_length=255)
```

Create file main.py:

<details>
<summary>Long file so I put it under spoiler</summary>

```python
import asyncio
from tortoise import Tortoise
from testapp.models import TableA, TableB, TableC

async def will_fail():
    a = await TableA.create(name="tableA")
    b = await TableB.create(id=5, name="tableB") # unrelated id
    await TableC.create(name="tableC", fkrelation=a)

async def will_probably_pass():
    a = await TableA.create(name="tableA")
    b = await TableB.create(name="tableB")
    assert a.id == b.id # if this is false, next line will fail
    await TableC.create(name="tableC", fkrelation=a) # Here TableA as I intended provided in place of TableB as expected by library

async def will_definitely_pass():
    a = await TableA.create(id=10, name="tableA")
    b = await TableB.create(id=10, name="tableB")
[...]

PR #1785, opened 2024-11-24
Parametrize UPDATE, DELETE and bulk operations

henadzit:
## Description
It's a follow up to https://github.com/tortoise/tortoise-orm/pull/1777.

- Parametrizes UPDATE, DELETE and bulk update and create operations

From my knowledge this should be the last bit of parametrization work.

## Motivation and Context
Parameterized queries are crucial for preventing SQL injection attacks and but also can improve performance of database operations.

Related ticket https://github.com/tortoise/tortoise-orm/issues/81
v1.0.0 target https://github.com/tortoise/tortoise-orm/issues/19

## How Has This Been Tested?
- Tests
- Tested with a web app localy

## Checklist:
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added the changelog accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1792, opened 2024-11-29
Add type validation for foreign key and one to one model consistency

Abdeldjalil-H:
Added validation to ensure consistency between the foreign key's expected model type and the actual model instance being assigned. This addresses an issue where mismatched model types could lead to silent failures or confusing database errors. 

## Description
Key changes:
1. Added type checking in Model._set_kwargs() when setting foreign key values
2. Raises a ValidationError with a clear error message when type mismatch occurs
3. Uses type() is comparison to ensure exact type matching (not allowing inheritance)
## Motivation and Context
This fixes issue #1791 where inconsistencies between `ForeignKeyField` model parameter and actual model types were not caught.

## How Has This Been Tested?
Running tests.

## Checklist:


henadzit:
@Abdeldjalil-H thank you for the PR!

A few comments:
- We need to add tests demonstrating that the check works
- Is the new check covers the case where the field is set to the model, e.g. `instance.relation = relation`? If not, this PR should add it.

Abdeldjalil-H:
@henadzit thank you for your comments. I'll work on that. 

Btw do you think we should use `isinstance` or is it okay to use the exact type comparison?

henadzit:
> Btw do you think we should use `isinstance` or is it okay to use the exact type comparison?

I think either will work since Tortoise only supports abstract class inheritance.

CHANGELOG.rst
```diff
@@ -16,6 +16,10 @@ Fixed
 - Fix unable to use ManyToManyField if OneToOneField passed as Primary Key (#1783)
 - Fix sorting by Term (e.g. RawSQL) (#1788)
 
+Added
```
henadzit: This version has been already released. Can add a note for unreleased version similar to what I've done [here](https://github.com/tortoise/tortoise-orm/pull/1785/files#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43cR12)?

tortoise/queryset.py
```diff
@@ -1217,6 +1217,7 @@ def _make_query(self, **pypika_kwargs) -> Tuple[str, List[Any]]:
             if field_object.pk:
                 raise IntegrityError(f"Field {key} is PK and can not be updated")
             if isinstance(field_object, (ForeignKeyFieldInstance, OneToOneFieldInstance)):
+                self.model._validate_relation_type(key, value)
```
henadzit, line 1207: Nice catch!

What about `BulkUpdateQuery` and `BulkCreateQuery`, is validation happening there too? It would be great to add tests for those.

tests/fields/test_fk.py
```diff
@@ -151,6 +156,11 @@ async def test_minimal__instantiated_create(self):
         tour = await testmodels.Tournament.create(name="Team1")
         await testmodels.MinRelation.create(tournament=tour)
 
+    async def test_minimal__instantiated_create_wrong_type(self):
+        author = await testmodels.Author.create(name="Author1")
+        with self.assertRaises(FieldError):
```
henadzit: You can use `assertRaisesRegex` to verify that it's failing because of the relation type.

tortoise/models.py
```diff
@@ -806,6 +809,17 @@ def _set_pk_val(self, value: Any) -> None:
     Can be used as a field name when doing filtering e.g. ``.filter(pk=...)`` etc...
     """
 
+    @classmethod
+    def _validate_relation_type(cls, field_key: str, value: Optional["Model"]) -> None:
```
henadzit, line 814: I think `value` should be either not optional or this method should skip validation if it is `None` because `None` is a correct value but with the current implementation, this method is going to fail if `None` is passed.

henadzit (commented):
@Abdeldjalil-H thank you for great work! I left a few more comments, please have a look!

CHANGELOG.rst
Abdeldjalil-H: My bad, I haven't noticed that.

tortoise/queryset.py
Abdeldjalil-H: Done, however, there is no way to do bulk create/update without init a model or assign a value.

tortoise/models.py
Abdeldjalil-H: Okay, None is a valid value now.

CHANGELOG.rst
```diff
@@ -11,6 +11,11 @@ Changelog
 
 0.22.2 (unreleased)
 ------
+Added
+^^^^^
+- Added type validation for foreign key fields to ensure type safety. Now raises ``ValidationError`` when assigning foreign key values with incorrect model types (#1792)
+0.22.1
```
henadzit: Formatting doesn't look correct here.

henadzit (commented):
There is an issue with CHANGELOG formatting and CI is failing. Apart from that the changes look good to me!

Abdeldjalil-H:
I don't know if the tests are failing because of the changes or other things. Is the database cleared after each test case? @henadzit

henadzit:
> I don't know if the tests are failing because of the changes or other things. Is the database cleared after each test case? @henadzit

The test cases that are failing are based on `test.TestCase` which is using transactions to isolate tests. I have a hunch that your changes have an unexpected side effect that lead to these failures.

henadzit:
@Abdeldjalil-H can you please rebase your changes and correct the CHANGELOG? I improved test isolation in https://github.com/tortoise/tortoise-orm/pull/1816, maybe, it will resolve issues in this PR.

henadzit (approved):

PR #1794, opened 2024-12-01
Fix bug combined expression connectors

Abdeldjalil-H:
This PR changes `expressions.Connector` values to be explicit strings. This is because the `operator` module does not have a function called `div` (it is `truediv`).

## Motivation and Context
We use enum values instead of names to get the operator function from `operator` module.

## How Has This Been Tested?
Running existing tests.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,12 @@ Changelog
 0.22
 ====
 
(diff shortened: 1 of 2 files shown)


henadzit (approved):
Nice catch!

PR #1797, opened 2024-12-03
Parametrize related field queries

### LINKED ISSUE #81: Change to all-parametrised queries

henadzit:
## Description
This is the final (hopefully) parametrization PR. After this is merged, all queries should be parameterized.

- Updates relational field queries to use parametrized queries
- Refactors QuerySets to use pypika's `get_parameterized_query` instead of doing parameterization itself. It shifts the responsibility of parameterization to pypika. [This PR ](https://github.com/tortoise/pypika-tortoise/pull/20) introduces `QueryBuilder.get_parameterized_query`.

## Motivation and Context
Parameterized queries are crucial for preventing SQL injection attacks and but also can improve performance of database operations.

This PR should finally allow us to close https://github.com/tortoise/tortoise-orm/issues/81.

This also checks one of the requirements for v1.0.0, [related ticket](https://github.com/tortoise/tortoise-orm/issues/19).

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Fixed
 Changed
 ^^^^^^^
 - Parametrizes UPDATE, DELETE, bulk update and create operations (#1785)
+- Parametrizes related field queries (#1797)
 
 0.22.1
 ------
--- a/poetry.lock
+++ b/poetry.lock
@@ -2658,15 +2658,11 @@ name = "pypika-tortoise"
 version = "0.3.2"
 description = "Forked from pypika and streamline just for tortoise-orm"
 optional = false
-python-versions = "^3.8"
-files = []
-develop = false
-
-[package.source]
-type = "git"
-url = "https://github.com/henadzit/pypika-tortoise.git"
-reference = "feat/query-supports-parametrization"
-resolved_reference = "764f99478ead03ecb567e316388d432a3fde6c3d"
+python-versions = "<4.0,>=3.8"
+files = [
+    {file = "pypika_tortoise-0.3.2-py3-none-any.whl", hash = "sha256:c5c52bc4473fe6f3db36cf659340750246ec5dd0f980d04ae7811430e299c3a2"},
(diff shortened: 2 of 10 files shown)


Abdeldjalil-H:
hi @henadzit. I have a question. I've noticed that if we call `.sql(params_inline=True)` on a query that has filters as subqueries, for example
```python
MyModel.filter(id__in=Subquery(OtherModel.filter(field=some_value).values_list("id"))).sql(params_inline=True)
```
then the value of `field` is not included in the raw sql. Does this PR handle this case or we should have a separate issue for this?

Thank you for your time.

henadzit:
> hi @henadzit. I have a question. I've noticed that if we call `.sql(params_inline=True)` on a query that has filters as subqueries, for example
> 
> ```python
> MyModel.filter(id__in=Subquery(OtherModel.filter(field=some_value).values_list("id"))).sql(params_inline=True)
> ```
> 
> then the value of `field` is not included in the raw sql. Does this PR handle this case or we should have a separate issue for this?
> 
> Thank you for your time.

Great catch! This PR doesn't handle it and not related, please open a separate issue, thanks.

abondar (approved):

PR #1810, opened 2024-12-13
Fix N-level transactions

### LINKED ISSUE #1128: Nested transactions deadlock
**Describe the bug**
On Postgres (`TransactionsContextPooled`) using more then one level of nested sub-transactions cause a deadlock.
All returned instances of `NestedTransactionPooledContext` in `__aenter__` tries to acquire same lock on connection object (`_trxlock`)

**To Reproduce**
On simple database setup with postgres try:
```python
async with in_transaction():   # first level transaction, works internally on TransactionContextPooled
   async with in_transaction():   # nested transaction, works internally on NestedTransactionPooledContext, locks _trxlock
     async with in_transaction():  # deadlock, another NestedTransactionPooledContext tries to acquire same _trxlock again
        pass
```

**Expected behavior**
Nested transactions should work.

**Additional context**
I tried to get why NestedTransactionPooledContext acquires lock on __aenter__ when just NestedTransactionContext not (thus nested transactions works on SQLite), but I have no idea. Maybe removing this `lock.acquire()` is a good workaround.

This case is a real use case. I'm trying to write tests, so first level transaction is created at test function level to isolate test cases. Tested code modifies bunch of models in a transaction, in some places update_or_create were used which runs internally inside their own transactions.

The only workaround I found as far is 'expanding' all update_or_create calls inline to avoid their transactions.

henadzit:
## Description
- Allow having more than 2 nested transaction blocks, e.g.
```python
async with in_transaction():
  async with in_transaction():
    async with in_transaction():
      ...
```
- Do no lock in nested transactions. This is not really needed because safepoints are not supposed by Tortoise ORM yet. This enabled the fix for the issue above. When we introduce safepoints, we will have to revisit this again, or... Make stricter rules on using `asyncio.gather` or similar tools within a transaction. For instance, [SQLAlchemy states that it is unsafe](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#session-faq-threadsafe). 
- Refactor the clients code by making `TransactionContext` an interface and moving the original code of `TransactionContext` to `tortoise.backends.sqlite.client.SqliteTransactionContext` because it is really only used by SQLite
- Add a few comments to the clients code. Hopefully, it will be a bit easier to read for the next person.
- Added documentation on transactions.

## Motivation and Context
This should fix https://github.com/tortoise/tortoise-orm/issues/1128.

## How Has This Been Tested?
`make ci` and running an app locally.

## Checklist:

### DIFF
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -69,6 +69,7 @@ class TestConcurrencyIsolated(test.IsolatedTestCase):
         count = await Tournament.all().count()
         self.assertEqual(count, 1000)
 
+
 @test.requireCapability(supports_transactions=True)
 class TestConcurrencyTransactioned(test.TestCase):
     async def test_concurrency_read(self):
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -19,10 +19,6 @@ class TestConcurrencyIsolated(test.IsolatedTestCase):
         all_read = await Tournament.all()
         self.assertEqual(set(all_write), set(all_read))
 
-    async def create_trans_concurrent(self):
-        async with in_transaction():
-            await asyncio.gather(*[Tournament.create(name="Test") for _ in range(100)])
-
     async def test_nonconcurrent_get_or_create(self):
         unas = [await UniqueName.get_or_create(name="c") for _ in range(10)]
         una_created = [una[1] for una in unas if una[1] is True]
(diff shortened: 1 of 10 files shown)


tests/test_concurrency.py
```diff
@@ -41,22 +37,38 @@ async def test_concurrent_get_or_create(self):
 
     @test.skipIf(sys.version_info < (3, 7), "aiocontextvars backport not handling this well")
     @test.requireCapability(supports_transactions=True)
-    async def test_concurrency_transactions_concurrent(self):
```
henadzit, line 44: Just moved the private method inside the test to make it more readable.

tests/test_concurrency.py
```diff
@@ -41,22 +37,38 @@ async def test_concurrent_get_or_create(self):
...
 
-    async def create_trans(self):
-        async with in_transaction():
-            await Tournament.create(name="Test")
-
     @test.skipIf(sys.version_info < (3, 7), "aiocontextvars backport not handling this well")
     @test.requireCapability(supports_transactions=True)
-    async def test_concurrency_transactions(self):
-        await asyncio.gather(*[self.create_trans() for _ in range(100)])
+    async def test_concurrent_transactions_with_single_op(self):
```
henadzit, line 51: ditto

abondar (approved):

PR #1816, opened 2024-12-17
Implement transaction savepoints

henadzit:
## Description
This PR:

- Implements savepoints for Postgres, MySQL, MSSQL and Sqlite
- Improves test isolation. Since Tortoise.init makes global changes, in some cases test runs were impacting each other.
- Updates the docs and add a warning about starting concurrent tasks from within a transaction and starting nested transactions inside it.
- Updates CHANGELOG to plan for a major release 0.23 since the transaction behavior is going to change

## Motivation and Context
Before the changes any exception within a transaction block, even nested, would rollback the whole transaction. Savepoints allow more granularity when handling errors:

```python
    async with in_transaction():
        await MyModel.create(name='foo')
        try:
            # this block will create a savepoint
            async with in_transaction():
                await MyModel.create(name='bar')
                # this will rollback to the savepoint, meaning that
                # the 'bar' record will not be created, however,
                # 'foo' will be created
                raise Exception()
        except Exception:
            pass
```

## How Has This Been Tested?
- `make ci`
- running local app

## Checklist:


CHANGELOG.rst
```diff
@@ -6,14 +6,21 @@ Changelog
 
 .. rst-class:: emphasize-children
 
-0.22
+0.23
 ====
-0.22.3 (unreleased)
+
+0.23.1 (unreleased)
```
abondar: Shouldn't it be 0.23.0?

docs/transactions.rst
```diff
@@ -7,12 +7,31 @@ Transactions
 Tortoise ORM provides a simple way to manage transactions. You can use the
 ``atomic()`` decorator or ``in_transaction()`` context manager.
 
-``atomic()`` and ``in_transaction()`` can be nested, and the outermost block will
-be the one that actually commits the transaciton. Tortoise ORM doesn't support savepoints yet.
+``atomic()`` and ``in_transaction()`` can be nested. The inner blocks will create transaction savepoints,
+and if an exception is raised and then caught outside of a nested block, the transaction will be rolled back
+to the state before the block was entered. The outermost block will be the one that actually commits the transaction.
+The savepoints are supported for Postgres, SQLite, MySQL and SQLite. For other databases, it is advised to
```
abondar: 2 times for SQlite?

abondar (commented):
Thanks, it's cool feature

CHANGELOG.rst
henadzit: Fixed.

docs/transactions.rst
henadzit: Fixed :)

abondar (approved):

PR #1817, opened 2024-12-17
fix: remove unused 'charset' in generate schema encoders

nglehuy:
Remove unused 'charset' in generate schema encoders

## Description

- remove unused 'charset' in generate schema encoders
- update element_type in ArrayField of postgres to be uppercase
- remove extra spaces when generate schema
- Fixes incorrect handling of default value `ArrayField` - use {} wrap in default value for postgres array field as in https://www.postgresql.org/docs/current/arrays.html#ARRAYS-INPUT
- add sqlite.capabilities support_update_limit_order_by=False because sqlite does not support it

## Motivation and Context

Error when using `get_schema_sql` on ArrayField with default value

```bash
self = <tortoise.backends.asyncpg.schema_generator.AsyncpgSchemaGenerator object at 0x106f28440>, default = [0, 0, 0, 0]

    def _escape_default_value(self, default: Any):
        if isinstance(default, bool):
            return default
>       return encoders.get(type(default))(default)  # type: ignore
E       TypeError: escape_sequence() missing 1 required positional argument: 'charset'
```

Encoders functions should accept the same number of arguments

## How Has This Been Tested?

Update `tests/schema/models_postgres_index.py`, `tests/schema/test_generate_schema.py` and run `make test`

## Checklist:

### DIFF
--- a/tests/schema/models_postgres_index.py
+++ b/tests/schema/models_postgres_index.py
@@ -1,5 +1,5 @@
 from tortoise import Model, fields
-from tortoise.contrib.postgres.fields import TSVectorField
+from tortoise.contrib.postgres.fields import TSVectorField, ArrayField
 from tortoise.contrib.postgres.indexes import (
     BloomIndex,
     BrinIndex,
@@ -19,6 +19,7 @@ class Index(Model):
     sp_gist = fields.CharField(max_length=200)
     hash = fields.CharField(max_length=200)
     partial = fields.CharField(max_length=200)
+    array = ArrayField(element_type="text", default=["a", "b", "c"])
 
     class Meta:
         indexes = [
--- a/tests/schema/models_postgres_index.py
+++ b/tests/schema/models_postgres_index.py
@@ -1,5 +1,5 @@
 from tortoise import Model, fields
-from tortoise.contrib.postgres.fields import TSVectorField, ArrayField
+from tortoise.contrib.postgres.fields import TSVectorField
 from tortoise.contrib.postgres.indexes import (
     BloomIndex,
     BrinIndex,
@@ -19,7 +19,6 @@ class Index(Model):
     sp_gist = fields.CharField(max_length=200)
     hash = fields.CharField(max_length=200)
     partial = fields.CharField(max_length=200)
-    array = ArrayField(element_type="text", default=["a", "b", "c"])
 
     class Meta:
         indexes = [
--- a/tortoise/converters.py
+++ b/tortoise/converters.py
@@ -25,7 +25,7 @@ def _escape_unicode(value: str, mapping=None) -> str:
 escape_string = _escape_unicode
 
 
-def escape_item(val: Any, charset, mapping=None) -> str:
+def escape_item(val: Any, mapping=None) -> str:
     if mapping is None:
         mapping = encoders
     encoder = mapping.get(type(val))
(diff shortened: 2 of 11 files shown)


tests/schema/models_postgres_index.py
```diff
@@ -19,6 +19,7 @@ class Index(Model):
     sp_gist = fields.CharField(max_length=200)
     hash = fields.CharField(max_length=200)
     partial = fields.CharField(max_length=200)
+    array = ArrayField(element_type="text", default=["a", "b", "c"])
```
henadzit: The `Index` model is used for testing indexes specifically and there is no index on array in this case. It would be better to use another model for testing arrays.

It might make sense to add a new test or find a more suitable one.

henadzit (commented):
Thank you for working on this! I added a single comment.

tests/schema/models_postgres_index.py
nglehuy: Hi @henadzit, sorry for the late fix, I've moved those to new tests and make some more changes

tortoise/converters.py
```diff
@@ -25,43 +25,44 @@ def _escape_unicode(value: str, mapping=None) -> str:
 escape_string = _escape_unicode
 
 
-def escape_item(val: Any, charset, mapping=None) -> str:
+def escape_item(val: Any, mapping=None) -> str:
+    if isinstance(val, str):
+        return f'"{val}"'
```
henadzit, line 30: Why is it required? Is " going to work for all databases?

tortoise/converters.py
```diff
@@ -25,43 +25,44 @@ def _escape_unicode(value: str, mapping=None) -> str:
...
 
-def escape_sequence(val: Sequence, charset, mapping=None) -> str:
+def escape_sequence(val: Sequence, mapping=None) -> str:
     n = []
     for item in val:
-        quoted = escape_item(item, charset, mapping)
+        quoted = escape_item(item, mapping)
         n.append(quoted)
-    return "(" + ",".join(n) + ")"
+    return "'{" + ",".join(n) + "}'"
```
henadzit, line 61: Is this going to work for all databases or is it Postgres-specific?

henadzit (commented):
Thanks for working on this! Sorry, I am not very familiar with the schema generation code, so I left a few questions.

tortoise/converters.py
nglehuy: this is for `escape_sequence`, which is only used in `_escape_default_value` in `schema_generator` for the types: tuple, list, set, frozenset

those types are used for array field, which is only supported by Postgres (MySQL, MSSQL, ORACLE do not support directly as far as I know)
I don't see those types are used anywhere else, the default for text and json fields are ignore as in https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/base/schema_generator.py#L223

tortoise/converters.py
nglehuy: Because it's used in `escape_sequence`, and it's wrapped by `'{` and `}'`, so the string element in array must be wrapped by `""`

tortoise/converters.py
henadzit: Does it mean that the default Array values weren't working previously?

nglehuy: image

Yeah, the previous one does not work

image

henadzit: Got it, thanks for explaining. I updated the PR description to explicitly state what is fixed.

henadzit (approved):
Thanks!

Issue #1565, opened 2024-03-07
Meta indexes are not created safely


rgabor-dev:
**Describe the bug**
When creating a `PartialIndex` with `Meta`, the index is not created safely nor in the migration file neither with `Tortoise.generate_schemas(safe=True)`. This results a failure when generating schemas.

**To Reproduce**

Tortoise ORM version: 0.20.0

Creating a test model with indexes:
```
from tortoise.models import Model
from tortoise.indexes import PartialIndex
from tortoise import fields

class TestModel(Model):
    name = fields.CharField(max_length=255)
    age = fields.IntField(index=True)

    class Meta:
        indexes = [
            PartialIndex(fields=["name"], condition={"name": "admin"}),
        ]
```

The generated migration:
```
CREATE TABLE IF NOT EXISTS "testmodel" (
    "id" SERIAL NOT NULL PRIMARY KEY,
    "name" VARCHAR(255) NOT NULL,
    "age" INT NOT NULL
);
CREATE INDEX IF NOT EXISTS "idx_testmodel_age_122507" ON "testmodel" ("age");
CREATE  INDEX "idx_testmodel_name_de9da2" ON "testmodel" ("name") WHERE name = 'admin';
```

Calling `await Tortoise.generate_schemas(safe=True)`:
```
tortoise.exceptions.OperationalError: relation "idx_testmodel_name_de9da2" already exists
```

**Expected behavior**
The index should be created safely using `IF NOT EXISTS` in the migration and `generate_schemas` shouldn't fail either when `safe` is set to `True`.

fraoustin:
I purpose a pull request https://github.com/tortoise/tortoise-orm/pull/1828

waketzheng:
@rgabor-dev What's your db, MySQL or PostgreSQL?
It seems that MySQL does not support the `CREATE INDEX IF NOT EXISTS` syntax.

rgabor-dev:
> @rgabor-dev What's your db, MySQL or PostgreSQL? It seems that MySQL does not support the `CREATE INDEX IF NOT EXISTS` syntax.

I'm using PostgreSQL.

fraoustin:
@waketzheng what is your mysql version?

waketzheng:
> @waketzheng what is your mysql version?

This one: `docker pull mysql`

PR #1828, opened 2024-12-26
fraoustin 1565 Meta indexes are created safely

fraoustin:
The index should be created safely using IF NOT EXISTS in the migration and generate_schemas shouldn't fail either when safe is set to True.

## Description
Add the exist format in template index create

## Motivation and Context
Issue #1565: When creating a PartialIndex with Meta, the index is not created safely nor in the migration file neither with Tortoise.generate_schemas(safe=True). This results a failure when generating schemas.

## Checklist:


abondar:
Tests are failing

abondar (approved):

PR #1834, opened 2025-01-02
Add more Codspeed benchmarks

henadzit:
## Description
This PR adds the following Codspeed benchmarks:
- `test_bulk_create_few_fields` and `test_bulk_create_many_fields` that benchmark `bulk_create`
- `test_filter_many_filters` that benchmarks filtering when multiple filters are used

## Motivation and Context
I am working on a performance improvements, there benchmarks are going to provide a baseline.

## How Has This Been Tested?
`make ci`

## Checklist:

### DIFF
--- a/tests/benchmarks/test_bulk_create.py
+++ b/tests/benchmarks/test_bulk_create.py
@@ -9,7 +9,7 @@ def test_bulk_create_few_fields(benchmark):
 
     data = [
         BenchmarkFewFields(
-            level=random.choice([10, 20, 30, 40, 50]), text=f"Insert from C, item {i}"
+            level=random.choice([10, 20, 30, 40, 50]), text=f"Insert from C, item {i}"  # nosec
         )
         for i in range(100)
     ]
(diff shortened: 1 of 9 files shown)


tests/benchmarks/test_bulk_create.py
```diff
@@ -0,0 +1,35 @@
...
+        loop.run_until_complete(_bench())
+
+
+def test_bulk_create_many_fields(benchmark, gen_many_fields_data):
+    loop = asyncio.get_event_loop()
+
+    data = [BenchmarkManyFields(**gen_many_fields_data()) for _ in range(100)]
+
+    @benchmark
+    def bench():
```
waketzheng, line 31: These lines are duplicated with the above function, how about refactor to:
```py
def _bench(data, loop):
    async def _bench():
        await BenchmarkFewFields.bulk_create(data)
    return lambda : loop.run_until_complete(_bench())

def test_bulk_create_many_fields(benchmark, gen_many_fields_data):

    data = [BenchmarkManyFields(**gen_many_fields_data()) for _ in range(100)]
    benchmark(_bench(data, loop))
```

henadzit: - They are not duplicated, they use different models: `BenchmarkFewFields` and `BenchmarkManyFields`
- This is against the convention of how Codspeed proposes to structure their benchmarks. I don't think we need to save a line of code by complicating it with extra functions.

waketzheng (approved):

henadzit:
@waketzheng thank you for reviewing!

PR #1838, opened 2025-01-04
feat: add filtered subquery update test

### LINKED ISSUE #1802: add tests: update on filtered query fails when using subquery
The original bug was fixed but we obviously lack tests that would have prevented the issue. This issue can be resolved by adding more tests covering usage of `Subquery` with updates. This might be a good first time issue.

# Original bug report

**Describe the bug**
As of #1777 / 7f077c169920e2981d74c4e66c5a24884dc241fb and version 0.22.1, update queries using `tortoise.expressions.Subquery` in filters fail with the following exceptions:

```
# sqlite3
sqlite3.ProgrammingError: Incorrect number of bindings supplied. The current statement uses 2, and there are 1 supplied.
# postgres
tortoise.exceptions.OperationalError: invalid input for query argument $1: 'test' ('str' object cannot be interpreted as an integer)
# from postgres' log
UPDATE "event" SET "name"=$1 WHERE "tournament_id" IN (SELECT "id" "0" FROM "tournament" WHERE "id"=$1)
```

**To Reproduce**
Run the following sample code under `tortoise-orm==0.22.1`:

```python
from tortoise import fields
from tortoise import Tortoise
import asyncio


class Tournament(Model):
    name = fields.CharField(max_length=255)


class Event(Model):
    tournament = fields.ForeignKeyField("models.Tournament", related_name="events")
    name = fields.CharField(max_length=255)


async def init():
    await Tortoise.init(
        config={
            "apps": {
                "models": {
                    "models": ["__main__"],
                },
            },
            "use_tz": True,
        }
    )


async def main():
    await init()

    await Tournament.create(name="abc")
    await Event.create(tournament_id=1, name="abc")

    # this succeeds
    affected = await Event.filter(tournament_id=1).update(name="test")
    print("0", affected)

    # this succeeds too
    events = await Event.filter(
        tournament_id__in=Subquery(Tournament.filter(id=1).values_list("id", flat=True))
    )
    print("1", events)

    # this fails
    affected = await Event.filter(
        tournament_id__in=Subquery(
            Tournament.filter(id=1).values_list("id", flat=True)
        ),
    ).update(name="test")
    print("2", affected)

if __name__ == "__main__":
    asyncio.run(main())
```

**Expected behavior**
The generated query should be correct, rows matching the filter should be updated.

**Additional context**
The good news is that it's fixed in #1797, however I cannot say why or how that specific PR fixes the underlying issue.

Invisi:
## Description
Adds a missing test for updates on filtered queries with subqueries.

## Motivation and Context
Should close #1802

## How Has This Been Tested?
`make test`

## Checklist:

### DIFF
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -13,6 +13,7 @@ from tests.testmodels import (
     Event,
     IntFields,
     JSONFields,
+    Reporter,
     Service,
     SmallIntFields,
     SourceFieldPk,
@@ -21,7 +22,7 @@ from tests.testmodels import (
 )
 from tortoise.contrib import test
 from tortoise.contrib.test.condition import In, NotEQ
-from tortoise.expressions import Case, F, Q, When
+from tortoise.expressions import Case, F, Q, Subquery, When
 from tortoise.functions import Function, Upper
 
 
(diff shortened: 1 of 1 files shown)


henadzit (approved):

henadzit:
Thanks for working on this!

PR #1840, opened 2025-01-04
Optimize field conversion to database format

henadzit:
## Description
- Remove `TO_DB_OVERRIDE` that is no longer needed after parametrization changes. The only exception is SQLite for which `sqlite3.register_adapter` was added - this is the same approach used by Django.
- Get rid of `BaseExecutor.column_map`. It was only required because of `TO_DB_OVERRIDE`.
- Cache `timezone.get_use_tz`, `timezone.get_timezone` and `timezone.get_default_timezone`. Since these are called for each field with `auto_now` or `auto_now_add`, it provides a significant speedup.

## Motivation and Context
This PR improves the performance of field value conversion to the DB format (see the detailed report in the comments):
- 52% improvement in the `test_bulk_create_many_fields` benchmark
- 13% improvement in `test_bulk_create_few_fields`
- 9% improvement in `test_create_many_fields`

The speedup is primarily achieved by the removal of `self.model._meta.db` from the `Field. to_db_value` code. The `db` property gets the current connection from `ContextVar` and it was a relatively slow operation to do it for each field for each object in `bulk_create`.

## How Has This Been Tested?
`make ci`

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,15 +6,18 @@ Changelog
 
 .. rst-class:: emphasize-children
 
-0.23
+0.24
 ====
 
-0.23.1
+0.24.0 (unreleased)
 ------
 Fixed
 ^^^^^
 - Rename pypika to pypika_tortoise for fixing package name conflict (#1829)
 - Concurrent connection pool initialization (#1825)
+Changed
+^^^^^^^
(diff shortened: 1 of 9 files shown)


henadzit:
@abondar any chance you can have a look at this? Thanks!

abondar (approved):
Good job!

PR #1845, opened 2025-01-08
fix: mysql generated schema syntax error

### LINKED ISSUE #1836: MySQL failed to create schema with `indexes=[Index(fields=...)]` in Meta
**Describe the bug**
MySQL does not support the `CREATE INDEX IF NOT EXISTS` syntax, so when I defined a model with `class Meta: indexes=[tortoise.indexes.Index(fields=['xxx'])]`, it raised a tortoise.exceptions.OperationalError:
```
(1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IF NOT EXISTS `idx_mymodel_name_2cb673` ON `mymodel` (`name`)' at line 1")
```

**To Reproduce**
- bug.py
```py


class MyModel(Model):
    name = fields.CharField(10)

    class Meta:
        indexes = [Index(fields=["name"])]


async def run() -> None:
    await Tortoise.init(
        db_url="mysql://root:123456@127.0.0.1:3306/test_",
        _create_db=True,
    )
    try:
        print(await MyModel.all())
    finally:


if __name__ == "__main__":
    run_async(run())
```
- Just run `python bug.py`

**Expected behavior**
Success to create table in db

**Additional context**
This bug was cause by changed of the following line: `git diff 968d0d5f444c03c0c6d65d7e8fa31b2f1762e45f cdd18488a7ab312d2dc5fda1d1852515ad62dee6 tortoise/indexes.py`
```diff
diff --git a/tortoise/indexes.py b/tortoise/indexes.py
index 9f39b61..96222e4 100644
--- a/tortoise/indexes.py
+++ b/tortoise/indexes.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 class Index:
     INDEX_TYPE = ""
     INDEX_CREATE_TEMPLATE = (
-        "CREATE{index_type}INDEX {index_name} ON {table_name} ({fields}){extra};"
+        "CREATE{index_type}INDEX {exists}{index_name} ON {table_name} ({fields}){extra};"
     )

     def __init__(
```

waketzheng:
## Description

Fixes #1836 

## Motivation and Context

- Remove ' IF NOT EXISTS' for INDEX creation sql if dialect is mysql.
- Delete extra space in `CREATE  INDEX` when INDEX_TYPE is empty.

## How Has This Been Tested?

make ci

## Checklist:


tortoise/utils.py
```diff
@@ -37,9 +37,11 @@ async def generate_schema_for_client(client: "BaseDBAsyncClient", safe: bool) ->
     :param safe: When set to true, creates the table only when it does not already exist.
     """
     generator = client.schema_generator(client)
-    schema = get_schema_sql(client, safe)
-    logger.debug("Creating schema: %s", schema)
-    if schema:  # pragma: nobranch
+    if schema := generator.get_create_schema_sql(safe):  # pragma: nobranch
+        if client.schema_generator.DIALECT == "mysql":
+            # MySQL does not support 'IF NOT EXISTS' syntax for `INDEX`
+            schema = schema.replace(" INDEX IF NOT EXISTS", " INDEX")
```
henadzit: This code is shared across all supported integrations, we shouldn't be really making hacks for particular databases here since we already use inheritance to customize the behavior of schema generation. Imagine if we start handling all edge cases for all databases here - the code will become unreadable pretty soon.

Looks like the issue is coming from https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/indexes.py#L62. What is going to happen if we replace `self.INDEX_CREATE_TEMPLATE` with `schema_generator.INDEX_CREATE_TEMPLATE`? 

I think it would be more correct to get rid of `Index.get_sql` altogether and just implement SQL generation for `Index` in `schema_generator` subclass, hence being able to customize it for each database.

waketzheng: @henadzit I add a `class _db_mysql` to the Index class.

henadzit: @waketzheng what do you think about my other suggestions? We work quite a lot together on this project and it would help a lot if we communicate better. In this example it would help if you shared your thoughts on why you decided to approach this problem in a different way than you initially planned or than proposed by me.

> I think it would be more correct to get rid of Index.get_sql altogether and just implement SQL generation for Index in schema_generator subclass, hence being able to customize it for each database.

> What is going to happen if we replace self.INDEX_CREATE_TEMPLATE with schema_generator.INDEX_CREATE_TEMPLATE?

waketzheng: @henadzit I think word is limit. For me, code is much easy to understand than word. Many many times, I was able to get the point why sb code like that quickly without reading any comments.
This PR was expected to be a hotfix for the BUG. It may not be the best solution for the creation of Index. But keep the library work correctly is more important that make code better. I would rather fix bug soon and leave the refactor job later.
**Why not replace self.INDEX_CREATE_TEMPLATE with schema_generator.INDEX_CREATE_TEMPLATE:**
image
As the above picture showing, both tortoise/backends/mysql/schema_generator.py and tortoise.indexes.py have a INDEX_CREATE_TEMPLATE var, and they are different value. It mean that get rid of Index.get_sql altogether in schema_generator will cost much time to do the refactor. Let's keep small step forward, don't change too much in such a hotfix PR!

henadzit: 1. It isn't a hot fix PR since the bug hasn't gotten into any release.
2. I proposed the changes and you ignored my comments pretty much and made other changes. That isn't just about getting the point, it's simple politeness. 
3. if you look at any successful open-source projects, you will see a lot of discussions and good communication.
4. Your changes at the moment aren't just a bug fix. You are introducing a new concept where the create index statements can be customized via `class _db_*`. The way it works in open-source and pretty much in any software development, other people will start following the pattern and extending the code base in a similar manner. 
5. We need to be trying to improve the code base with each our change, otherwise it will decay pretty quickly. "I will do a hot fix and then we refactor it later" rarely works.

I'll give a shot at this issue to see if there is an approach where we can simplify things.

henadzit:
[Here's my shot at this issue ](https://github.com/tortoise/tortoise-orm/pull/1847), where I'm removing `Index.sql` and relaying only on `schema_generator` to build correct SQL. Please have a look!

tortoise/utils.py
waketzheng: @henadzit 
1. No matter whether it is a hot fix or not, we should keep the develop branch work fine.
2. Actually, I try your suggestion `get rid of Index.get in schema_generator` and find it hard to do that, so I push commits with another solution, I expected to get your response after compare these two solutions.
3. I read some PRs in poetry and fastapi, they did communicate a lot, but sometimes replied after several weeks. It  is my pleasure that if we can communicate quickly and smoothly.
4. `class _db_*` is not a new concept, you can find it at https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/fields/data.py#L90
5. Everyone have only 24 hours one day, we should focus on crucial things, "rarely" mean that there are jobs more important that this one.

> I'll give a shot at this issue to see if there is an approach where we can simplify things.
OK, I'll pull your branch to local and add some tests.

waketzheng:
> [Here's my shot at this issue ](https://github.com/tortoise/tortoise-orm/pull/1847), where I'm removing `Index.sql` and relaying only on `schema_generator` to build correct SQL. Please have a look!

@henadzit My PR includes some useful tests. How about merge it and then your new PR can use them directly?

tests/utils/test_describe_model.py
```diff
@@ -25,12 +25,19 @@
...
+    if isinstance(obj, Index):
+        return repr(obj)
+    return obj
 
 
 class TestDescribeModels(test.TestCase):
     def test_describe_models_all_serializable(self):
         val = Tortoise.describe_models()
-        json.dumps(val)
+        json.dumps(val, default=dump_index)
```
henadzit, line 40: This hides the issue that `Index` cannot be described in a serializable way, however, it doesn't fix it. I fixed it in [my PR](https://github.com/tortoise/tortoise-orm/pull/1847) by implementing `Index.describe`.

tests/testmodels.py
```diff
@@ -150,6 +151,9 @@ class ModelTestPydanticMetaBackwardRelations3(Model):
 class Node(Model):
     name = fields.CharField(max_length=10)
 
+    class Meta:
+        indexes = [Index(fields=("name",))]
```
henadzit, line 155: It looks like you added this index because it exposes the issue with serializing `Index` in `Model.describe`. I was able to identify it only because I worked on the same issue in my PR, it wouldn't be clear to anyone reviewing this PR.

Even though this index exposes an issue, it isn't documented and completely not obvious why it is here. Someone can remove it later without knowing its purpose and the same regression can happen again. I added a dedicated class for testing indexes in my PR.

PR #1847, opened 2025-01-10
Fix index creation in Tortoise.generate_schemas() for MySQL and Postgres

### LINKED ISSUE #1836: MySQL failed to create schema with `indexes=[Index(fields=...)]` in Meta

henadzit:
## Description
- Replace `CREATE INDEX IF NOT EXISTS` queries with indexes defined inside the table definition for MySQL
- Fix partial index creation for Postgres - `CREATE INDEX "idx_index_partial_c5be6a" ON "index" USING  ("partial") WHERE id = 1;` isn't correct syntax because `USING` should be followed by the index type.
- Replace the logic for index creation from `Index` and its subclasses with `schema_generator`, hence making `schema_generator` a single place to have schema generation logic. This also also makes the `Index` and `schema_generator` relation one directional.
- Fix the issue with using `Model.describe` when `Index` is in `Model.Meta.indexes`

## Motivation and Context
This should fix https://github.com/tortoise/tortoise-orm/issues/1836.

## How Has This Been Tested?
`make ci`

I also tested the generated SQLs against corresponding databases.

## Checklist:

### DIFF
--- a/tortoise/backends/base/schema_generator.py
+++ b/tortoise/backends/base/schema_generator.py
@@ -1,6 +1,6 @@
 import re
 from hashlib import sha256
-from typing import TYPE_CHECKING, Any, List, Set, Type, Union, cast
+from typing import TYPE_CHECKING, Any, List, Optional, Set, Type, Union, cast
 
 from tortoise.exceptions import ConfigurationError
 from tortoise.fields import JSONField, TextField, UUIDField
@@ -23,8 +23,10 @@ class BaseSchemaGenerator:
     DIALECT = "sql"
     TABLE_CREATE_TEMPLATE = 'CREATE TABLE {exists}"{table_name}" ({fields}){extra}{comment};'
     FIELD_TEMPLATE = '"{name}" {type}{nullable}{unique}{primary}{default}{comment}'
-    INDEX_CREATE_TEMPLATE = 'CREATE INDEX {exists}"{index_name}" ON "{table_name}" ({fields});'
-    UNIQUE_INDEX_CREATE_TEMPLATE = INDEX_CREATE_TEMPLATE.replace(" INDEX", " UNIQUE INDEX")
+    INDEX_CREATE_TEMPLATE = (
+        'CREATE {index_type}INDEX {exists}"{index_name}" ON "{table_name}" ({fields}){extra};'
+    )
+    UNIQUE_INDEX_CREATE_TEMPLATE = INDEX_CREATE_TEMPLATE.replace("INDEX", "UNIQUE INDEX")
     UNIQUE_CONSTRAINT_CREATE_TEMPLATE = 'CONSTRAINT "{index_name}" UNIQUE ({fields})'
     GENERATED_PK_TEMPLATE = '"{field_name}" {generated_sql}{comment}'
     FK_TEMPLATE = ' REFERENCES "{table}" ("{field}") ON DELETE {on_delete}{comment}'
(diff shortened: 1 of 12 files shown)


tortoise/backends/base/schema_generator.py
```diff
@@ -324,22 +338,30 @@ def _get_table_sql(self, model: "Type[Model]", safe: bool = True) -> dict:
...
+                        fields.append(field_object.source_field or field)
 
-                    _indexes.append(self._get_index_sql(model, indexes_to_create, safe=safe))
+                    _indexes.append(self._get_index_sql(model, fields, safe=safe))
                 else:
-                    _indexes.append(indexes_list.get_sql(self, model, safe))
+                    if index.fields:
+                        fields = [f for f in index.fields]
+                    else:
+                        fields = [f"({expression.get_sql()})" for expression in index.expressions]
```
abondar: Maybe it would be a little bit clearer with
```
                    if index.fields:
                        fields = [f for f in index.fields]
                    elif index.expressions:
                        fields = [f"({expression.get_sql()})" for expression in index.expressions]
                     else:
                           raise RuntimeError("Invalid index object")
```

Not that much of difference, but could save some headache in case of future refactorings or something like that

abondar:
@waketzheng what do you think about this approach? 

Even though it moves more code - I think end result seems a bit clearer, as we don't introduce new entities and improve existing code by moving out dialect-specific schema generation code to it's intended place

tortoise/backends/base/schema_generator.py
waketzheng: @abondar No need to raise RuntimeError, the `Index.__init__` prevent the case that both fields and expressions are empty.
image

I prefer one line style: `fields = list(index.fields) if index.fields else [f"({exp.get_sql()})" for exp in index.expressions]`

abondar: I understand that it is enforced there, but it is not guaranteed by anything but this code. If this code changes for some reason - this `else` clause can become wider and not guarantee existence of `expressions`

Although it is quite hypothetical situation and I do not insist that it should be changed, just explicit check of existence of `expressions` seems cleaner to me

waketzheng:
Looks good.

@henadzit Cloud you add more test cases? For example, add `class Meta: Index(fields=['name'])` to the Node class in tests/testmodels.py

henadzit:
> Looks good.
> 
> 
> 
> @henadzit Cloud you add more test cases? For example, add `class Meta: Index(fields=['name'])` to the Node class in tests/testmodels.py

Will do!

tortoise/backends/base/schema_generator.py
henadzit: I added an extra check.

henadzit:
I added tests and fixed the issue with using `Model.describe` when an `Index` is in `Model.Meta.indexes`.

abondar (approved):

PR #1854, opened 2025-01-17
filter datetime objects by date using __date

xaptronic:
Add functionality to allow filtering for `datetime` type fields by date using `{field_name}__date`

## Description
I have fields in my models/schema that are datetime objects, but I wish to be able to filter for them using only a date (e.g.; filter all entries on a particular date).

This PR adds `__date` operator that allows for such filtering.

## Motivation and Context
There does not appear to be a way to filter datetime fields by a date object. Perhaps it can be done using `lte` / `gte` but I would rather specify one filter.. it seems more obvious and less of a work around a missing filter type.

## How Has This Been Tested?
Added unit tests and used functionality in my use-case.

## Checklist:

### DIFF
--- a/tests/test_filtering.py
+++ b/tests/test_filtering.py
@@ -164,6 +164,15 @@ class TestFiltering(test.TestCase):
         self.assertEqual(len(tournaments), 2)
         self.assertSetEqual({t.name for t in tournaments}, {"0", "1"})
 
+    async def test_filter_date(self):
+        await DatetimeFields.create(
+            datetime=datetime.datetime(
+                year=2020, month=5, day=20, hour=0, minute=0, second=0, microsecond=0
+            )
+        )
+        date = datetime.date(2020, 5, 20)
+        self.assertEqual(await Tournament.filter(created__date=date).count(), 1)
+
     @test.requireCapability(dialect="mysql")
     @test.requireCapability(dialect="postgres")
     async def test_filter_exact(self):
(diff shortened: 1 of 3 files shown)


tests/test_filtering.py
```diff
@@ -164,6 +164,15 @@ async def test_filter_not_with_or(self):
...
         self.assertSetEqual({t.name for t in tournaments}, {"0", "1"})
 
+    async def test_filter_date(self):
+        await DatetimeFields.create(
+            datetime=datetime.datetime(
+                year=2020, month=5, day=20, hour=0, minute=0, second=0, microsecond=0
+            )
+        )
+        date = datetime.date(2020, 5, 20)
+        self.assertEqual(await DatetimeFields.filter(datetime__date=date).count(), 1)
```
henadzit, line 174: We should have more tests for these: checking that this filter modifier actually converts a date time to a date. The provided test case doesn't actually test anything. For instance, in Postgres

```
SELECT TIMESTAMP '2024-01-20 00:00:00' = DATE '2024-01-20';
> true
```

henadzit (changes requested):

henadzit:
Thanks for looking into this!

Here's a few links to how Django implements this:
- https://github.com/django/django/blob/22fc151bb86a553d84c62d7effd289356e9b6c6c/django/db/models/functions/datetime.py#L402
- [Postgres implementation](https://github.com/django/django/blob/main/django/db/backends/postgresql/operations.py#L120) of `datetime_cast_date_sql`

In Django they implement the conversion for each database, so this might be more complicated than just using the Date function.

PR #1859, opened 2025-01-22
fix `_get_dialects` - properties

markus-96:
## Description

properties in `_db_{dialect}` classes were not taken into account.

## Motivation and Context

see #1858 

## How Has This Been Tested?

`make build test_sqlite test_asyncpg`

also, I am not able to test against an actual oracle database, and as far as I can tell is that it is currently not tested at all in CICD. So please, someone check it for me? :)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added the changelog accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1862, opened 2025-01-23
add benchmarks for `get_for_dialect`

markus-96:
these are high-level-benchmarks that will somehow call `Field.get_for_dialect`

## Description

The above mentioned method has some potential to be optimized. 

## Motivation and Context

I want to optimize the method, like in this comment: https://github.com/tortoise/tortoise-orm/pull/1859#issuecomment-2607651773

It is not used very often, but for example if you call any tortoise.expressions.Function, filter through/retrieve values of related models,...

## How Has This Been Tested?

there seams to be a slight improvement with my optimizations

## Checklist:


CHANGELOG.rst
```diff
@@ -16,6 +16,7 @@ Fixed
 - Rename pypika to pypika_tortoise for fixing package name conflict (#1829)
 - Concurrent connection pool initialization (#1825)
 - `_get_dialects`: support properties (#1859)
+- add benchmarks for `get_for_dialect` (#1862)
```
henadzit: `0.24.0` has been released. Can you please add `0.25.0 (unreleased)` to the file?

tests/benchmarks/test_annotate.py
```diff
@@ -0,0 +1,39 @@
...
+from tortoise.functions import Count
+
+
+def test_function_count(benchmark, few_fields_benchmark_dataset):
+    loop = asyncio.get_event_loop()
+
+    @benchmark
+    def bench():
+        async def _bench():
+            await BenchmarkFewFields.annotate(text_count=Count("text")).all()
```
henadzit: No need to call `.all()`

tests/benchmarks/test_annotate.py
```diff
@@ -0,0 +1,39 @@
...
+
+        loop.run_until_complete(_bench())
+
+
+def test_values_related_m2m(benchmark, create_team_with_participants):
```
henadzit: This one doesn't have `.annotate` but it is located in the `test_annotate`. Maybe, move to `test_filter.py`?

tests/benchmarks/test_annotate.py
```diff
@@ -0,0 +1,39 @@
+import asyncio
+from decimal import Decimal
+
+from tests.testmodels import Event, BenchmarkFewFields, DecimalFields
+from tortoise.expressions import F
+from tortoise.functions import Count
+
+
+def test_function_count(benchmark, few_fields_benchmark_dataset):
```
henadzit: All the functions in this file are going to be the names of the benchmarks, so it would be nice to name them well:
- `test_annotate_with_count`
- `test_annotate_with_decimal`
- etc.

tests/benchmarks/test_annotate.py
markus-96: moved it to `test_relations.py` for now. Maybe we can add more tests benchmarking relational behaviour in the future

tests/benchmarks/test_annotate.py
markus-96: `test_annotate.py` was a working title I did not change yet, sorry for that. I simply wanted some functions that call `get_for_dialect`, so I went ahead and looked where the function is used, placed a `1/0` before/after the function call, ran `make ci` and hoped that an error is thrown somewhere. Then, I copied these "failing" tests and adopted them as a benchmark. Duriung this, I skipped everything in `schema_generator.py` because this is only executed once during startup and would not benchmark any real-world scenario.

henadzit (approved):
Looks good to me!

Issue #1853, opened 2025-01-17
IntField: constraints not taken into account


markus-96:
**Describe the bug**
The IntField has the property constraints. It contains the constraints `ge` and `le`. These constraints are not taken into account (in any place as far as I can tell) or they should be reflected by a validator.

**To Reproduce**
```py
from tortoise import models, fields, run_async
from tortoise.contrib.test import init_memory_sqlite


class Acc(models.Model):
    id = fields.IntField(pk=True)
    some = fields.IntField()


async def main():
    constraints_id = Acc._meta.fields_map.get('id').constraints  # {'ge': -2147483648, 'le': 2147483647}
    too_high_id = constraints_id.get('le') + 1
    constraints_some = Acc._meta.fields_map.get('some').constraints  # {'ge': -2147483648, 'le': 2147483647}
    too_low_some = constraints_some.get('ge') - 1
    acc = Acc(id=too_high_id, some=too_low_some)  # this should throw an error
    await acc.save()  # or maybe this


if __name__ == '__main__':
    run_async(init_memory_sqlite(main)())
```

**Expected behavior**
Either the constraints should match the capabilities of the DB or the constraints should be checked beforehand. And an error should be thrown.

**Additional context**
The produced Pydantic Model by pydantic_model_creator would check those constraints.

PR #1868, opened 2025-01-30
IntField: add MinMaxValidator

markus-96:
## Description

Before sending anything to the DB, IntField values are checked beforehand now. As far as I could test, the added overhead is not much, lets see what Codspeed says.

Will add documentation later.

## Motivation and Context

Using sqlite, you could send values greater than defined in `IntField.constraints.get('le')` to the DB and nothing was raised. If you for example want to migrate to another DBMS, you have a problem.

Also, have a look at the Issue I raised: #1853

## How Has This Been Tested?

Existent tests and new tests

## Checklist:

### DIFF
--- a/tortoise/fields/base.py
+++ b/tortoise/fields/base.py
@@ -326,12 +326,17 @@ class Field(Generic[VALUE], metaclass=_FieldMeta):
         """
         if not self.has_db_field:  # pragma: nocoverage
             return None
+        default = getattr(self, "SQL_TYPE")
         return {
-            "": getattr(self, "SQL_TYPE"),
+            "": default,
             **{
-                dialect: _db["SQL_TYPE"]
-                for dialect, _db in self._get_dialects().items()
-                if "SQL_TYPE" in _db
+                dialect: sql_type
+                for dialect, sql_type in (
+                    (key[4:], self.get_for_dialect(key[4:], "SQL_TYPE"))
+                    for key in dir(self)
+                    if key.startswith("_db_")
+                )
+                if sql_type != default
             },
         }
 
@@ -342,8 +347,18 @@ class Field(Generic[VALUE], metaclass=_FieldMeta):
         :param dialect: The requested SQL Dialect.
         :param key: The attribute/method name.
         """
-        dialect_data = self._get_dialects().get(dialect, {})
-        return dialect_data.get(key, getattr(self, key, None))
+        try:
+            dialect_cls = getattr(self, f"_db_{dialect}")  # throws AttributeError if not present
+            dialect_value = getattr(dialect_cls, key)  # throws AttributeError if not present
+        except AttributeError:
+            pass
+        else:  # we have dialect_cls and dialect_value, so lets use it
+            # it could be that dialect_value is a computed property, like in CharField._db_oracle.SQL_TYPE,
+            # and therefore one first needs to instantiate dialect_cls
+            if isinstance(dialect_value, property):
+                return getattr(dialect_cls(self), key)
+            return dialect_value
+        return getattr(self, key, None)  # there is nothing special defined, return the value of self
 
     def describe(self, serializable: bool) -> dict:
         """


tortoise/fields/data.py
```diff
@@ -71,21 +71,57 @@ class IntField(Field[int], int):
...
+            self.ge = ge
+        if le is None:
+            self.le = self.LE
+        else:
+            if le > self.LE:
+                raise ConfigurationError(f"'le' must be <= {self.LE}")
+            self.le = le
+
+    def to_db_value(self, value: Any, instance: "Union[Type[Model], Model]") -> Any:
+        if value is not None:
```
henadzit: I think we should be adding a validator instead of customizing `to_db_values` similarly how it is done in [CharField](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/fields/data.py#L185)

tortoise/fields/data.py
```diff
@@ -71,21 +71,57 @@ class IntField(Field[int], int):
...
     SQL_TYPE = "INT"
     allows_generated = True
+    GE = -2147483648
+    LE = 2147483647
 
-    def __init__(self, primary_key: Optional[bool] = None, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        primary_key: Optional[bool] = None,
+        ge: Optional[int] = None,
```
henadzit: Extending the field configuration with ge and le introduces the second way of setting up validation. I think we should have a single way of doing it to keep things simple - through validators. If some one needs specific validation, they can add a validator. So I suggest to remove these options from the constructor.

tortoise/fields/base.py
```diff
@@ -269,7 +269,7 @@ def validate(self, value: Any) -> None:
         """
         for v in self.validators:
             if self.null and value is None:
-                continue
+                return
```
henadzit: I don't think this is correct.

tortoise/fields/data.py
markus-96: that's what I did first, but it adds far too much overhead decreasing the benchmarks like 10% (see the edit history of the comment of codspeed..)

I think that's because it will actually enter the for each loop in self.validate instead of doing nothing. Also I think that overriding to_do_value actually increases speed because the interpreter does not have to go that much up the MRO.

tortoise/fields/data.py
markus-96: ge and le are currently baked in in the constraints property. Instead of that I wanted them to be customizable, a bit like with max_char in CharField where this is reflected during validation and also by the constraints property.

tortoise/fields/base.py
markus-96: If you have like ten validators in self.validators, you will always have to perform the check whether the field may be null and if the given value is None/null. Both informations do not change during itering other the validators. So instead of performing the same check over and over again, we can instead directly say that nothing else will happen here and that the value is valid.

waketzheng:
@markus-96 I push another solution: #1872

markus-96:
> @markus-96 I push another solution: #1872

Sorry if I ask, but where is this different from mine? I put much effort into optimizing my solution, could you tell what is wrong about that? I think we all have an interest in tortoise-orm being the fastest orm for python, so something that crucial like an int field has to be as fast as possible as it can be, so I think hard-coding default validators is a completly valid solution.

tortoise/fields/base.py
henadzit: Oh, gotcha, I see what you are saying after having a second look.

tortoise/fields/data.py
henadzit: > ge and le are currently baked in in the constraints property. Instead of that I wanted them to be customizable, a bit like with max_char

The int field constraints are related to the nature of the data type. I think it is different from business logic validation and we already have a way of doing it. We shouldn't have two ways of doing the same thing.

tortoise/fields/data.py
```diff
@@ -71,21 +71,57 @@ class IntField(Field[int], int):
...
+            self.le = self.LE
+        else:
+            if le > self.LE:
+                raise ConfigurationError(f"'le' must be <= {self.LE}")
+            self.le = le
+
+    def to_db_value(self, value: Any, instance: "Union[Type[Model], Model]") -> Any:
+        if value is not None:
+            if not isinstance(value, self.field_type):
+                value = self.field_type(value)  # pylint: disable=E1102
```
henadzit: In which cases is it required?

tortoise/fields/data.py
henadzit: Gotcha, that makes sense!

henadzit:
I think we should be able to merge it if we remove the ability to customize constraints.

waketzheng:
> Sorry if I ask, but where is this different from mine? I put much effort into optimizing my solution, could you tell what is wrong about that? I think we all have an interest in tortoise-orm being the fastest orm for python, so something that crucial like an int field has to be as fast as possible as it can be, so I think hard-coding default validators is a completly valid solution.

@markus-96 Your issue #1853 request a feature to check constraints for int fields before saving to db. There are several ways to do that. We can compare the two PRs and choose one or merge both. Code speed is important, and compatibility also counts. I use Django for a long time before migrating to tortoise-orm, and do not want it to be too much different from Django.

CHANGELOG.rst
```diff
@@ -17,6 +17,8 @@ Fixed
 Changed
 ^^^^^^^
 - add benchmarks for `get_for_dialect` (#1862)
+- optimize `get_for_dialect` (#1863)
+- add validator to `IntField` (#1866)
```
henadzit: Can you please extend the note so it is easier to understand for people reading the changelog? I think something like "Added default range validation to `IntField`, `BigIntField`, `SmallIntField`" would be better.

Could you please rebase on develop? I cleaned up the changelog to remove internal notes from the changelog.

markus-96: i am so bad at rebasing, i create a new pr...

PR #1872, opened 2025-02-05
feat: validate int value range for IntField

### LINKED ISSUE #1853: IntField: constraints not taken into account

waketzheng:
## Description

Close #1853

## Motivation and Context

Add a new validator for int fields to check value range when saving.

## How Has This Been Tested?

make ci

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,21 +6,19 @@ Changelog
 
 .. rst-class:: emphasize-children
 
-0.25
+0.24
 ====
 
-0.25.0 (unreleased)
+0.24.1 (unreleased)
 ------
 Fixed
 ^^^^^
+- IntField: constraints not taken into account (#1861)
 - Fixed asyncio "no current event loop" deprecation warning by replacing `asyncio.get_event_loop()` with modern event loop handling using `get_running_loop()` with fallback to `new_event_loop()` (#1865)
 
 Changed
 ^^^^^^^
 - add benchmarks for `get_for_dialect` (#1862)
-
-0.24
-====
 
 0.24.0
 ------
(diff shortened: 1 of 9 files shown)


waketzheng:
@henadzit CodSpeed check failed. Int fields validator increases bulk_create by 2ms. Which one is better:
- [ ] 1. Ignore it as 2ms is acceptable
- [ ] 2. Refactor `bulk_create` to skip value validating for int fields

henadzit:
@waketzheng I think @markus-96 makes a good point about performance in [his PR](https://github.com/tortoise/tortoise-orm/pull/1868). I think it should be fine to have the int field constraints in `to_db_value` if it doesn't harm performance.

PR #1873, opened 2025-02-05
fix: update pk field only raises unfriendly error

### LINKED ISSUE #1870: Bug/Feature: Allow updating primary key of an object
**Describe the bug**
If I do this:
```python
user.id = 123
await user.save(update_fields=['id'])
```
the operation fails with a database error:
```
asyncpg.exceptions._base.InterfaceError: the server expects 0 arguments for this query, 1 was passed
HINT:  Check the query against the passed list of arguments.  Note that parameters are supported only in SELECT, INSERT, UPDATE, DELETE, MERGE and VALUES statements, and will *not* work in statements  like CREATE VIEW or DECLARE CURSOR.
```

**To Reproduce**
See above.

**Expected behavior**
The object's primary key should be updated.

**Additional context**
This if-statement
https://github.com/tortoise/tortoise-orm/blob/960b1c1b81e44a06ab177fedfb2fbb878cabe0a4/tortoise/backends/base/executor.py#L270
removes the `id` from the update field list, and this results in an empty SQL statement being generated.

I can update the primary key with hand-written SQL just fine.

waketzheng:
## Description

Close #1870 

## Motivation and Context

Explicit raises OptionalError when executing `await model.save(update_fields=['id'])`

## How Has This Been Tested?

make ci

## Checklist:


tests/test_model_methods.py
```diff
@@ -121,6 +123,8 @@ async def test_save_partial(self):
         n_mdl = await self.cls.get(id=self.mdl.id)
         self.assertEqual(n_mdl.name, "Test")
         self.assertEqual(n_mdl.desc, "Something")
+        with pytest.raises(OperationalError):
```
henadzit: I would suggest to have a separate test case for this os it is documented with tests and it is not just a check in an existing test. 

I would also suggest another test case:

```
o.save()
o.id = 123
with pytest.raises(OperationalError):
  o.save()
```

tortoise/backends/base/executor.py
```diff
@@ -264,16 +265,20 @@ def get_update_sql(
     async def execute_update(
         self, instance: "Union[type[Model], Model]", update_fields: Optional[Iterable[str]]
     ) -> int:
+        fields: "dict[str, Field]" = {
```
henadzit: We don't have performance benchmarks for updates but I'm pretty sure that two loops is slower than a single loop. Why not just check for the pk field in the existing loop?

tests/test_model_methods.py
waketzheng: I tried, but it's hard to do that -- because can't get the original id if it changed which make it difficult to know whether object only changed pk field value.

waketzheng: @henadzit Cloud your review again?

henadzit: Sorry, I was traveling and was quite busy recently, so wasn't able to participate much. I'll have a look in coming days, thanks!

henadzit:
@waketzheng [I added benchmarks for updates](https://github.com/tortoise/tortoise-orm/pull/1881), let's merge them first so we have a baseline for performance and can be sure that this changes doesn't impact it significantly.

tortoise/backends/base/executor.py
```diff
@@ -267,13 +267,15 @@ async def execute_update(
         values = []
         expressions = {}
         for field in update_fields or self.model._meta.fields_db_projection.keys():
```
henadzit, line 269: The current logic assumes that if no values and no expressions, then the user must have been passed pk as part of `update_fields`. That is a bit hard to follow. What do you think about the version below?

```python
for field in update_fields or self.model._meta.fields_db_projection.keys():
  field_obj = self.model._meta.fields_map[field]
  if field_obj.pk:
    if update_fields:
      raise OperationalError(
                f"Can't update pk field, use `{self.model.__name__}.create` instead."
            )
    continue
```

tortoise/backends/base/executor.py
```diff
@@ -267,13 +267,15 @@ async def execute_update(
...
-            if not self.model._meta.fields_map[field].pk:
-                instance_field = getattr(instance, field)
-                if isinstance(instance_field, Expression):
+            if not (field_obj := self.model._meta.fields_map[field]).pk:
```
henadzit: Personally I find this syntax hard to read, having `field_obj` outside of the if statement feels better to me but it might be a personal preference.

henadzit (commented):
Can you please rebase on top of the latest changes in develop?

tests/test_model_methods.py
```diff
@@ -121,6 +123,8 @@ async def test_save_partial(self):
...
+            await self.mdl.save(update_fields=["id"])
```
henadzit, line 128: Can you make this check a separate test so it is documented behavior? I would also suggest checking the exception message, e.g.`with pytest.raise(OperationalError, match="...")`

tortoise/backends/base/executor.py
waketzheng: I think we should allow `.update_fields=['id', 'xxx']` as the Django do. And show warning to make sure user want to do that.

henadzit: > I think we should allow .update_fields=['id', 'xxx'] as the Django do. And show warning to make sure user want to do that.

Why introduce a behavior that triggers a warning? This doesn't seem like a good design. You usually introduce warnings about deprecation or something you want users to stop doing.

tortoise/backends/base/executor.py
```diff
@@ -267,13 +268,23 @@ async def execute_update(
...
-                    expressions[field] = instance_field
-                else:
-                    value = self.model._meta.fields_map[field].to_db_value(instance_field, instance)
-                    values.append(value)
+            field_obj = self.model._meta.fields_map[field]
+            if field_obj.pk:
+                if update_fields:
+                    if len(update_fields) == 1:  # type:ignore
+                        raise OperationalError(
+                            f"Can't update pk field only, use `{self.model.__name__}.create` instead."
```
henadzit:
```suggestion
                            f"Can't update pk field, use `{self.model.__name__}.create` instead."
```

tests/test_model_methods.py
henadzit: By introducing a separate test I meant 

```
async def test_save_partial_with_id_update(...):
```

tortoise/backends/base/executor.py
```diff
@@ -267,13 +268,23 @@ async def execute_update(
...
+                        )
+                    else:
+                        warnings.warn(
+                            "Do not add pk field to `update_fields`! It may change exists record by mistake.",
+                            RuntimeWarning,
+                        )
+                continue
+            if isinstance(instance_field := getattr(instance, field), Expression):
```
henadzit: This is quite hard to read. I can see why you would want to use the := operator in a simple if-else like

```
if a := something:
  ...
```

but in this case the assignment is hidden within a function call which make it quite hard to read.  I would rather have  the variable outside of the `if` check.

waketzheng: Updated.

Look at this: https://peps.python.org/pep-0572/#examples

To keep easy to read:
- This is ok:
```py
match1 = pattern1.match(data)
if match1:
   print(match1)

# Change to be 
if match1 := pattern1.match(data):
   print(match1)
```
- This ok too
```py
group = re.match(data).group(1) if re.match(data) else None

# Change to be
group = m.group(1) if (m := re.match(data)) else None
```
----
This is not recommended:
```py
if (match1 := pattern1.match(data)) and len(match1.groups()) > 2:
   print(match1)
```
This not recommended either:
```py
if isinstance(m := get_object(), str):
    return m
````

waketzheng:
@henadzit All done.

> Why introduce a behavior that triggers a warning? This doesn't seem like a good design. You usually introduce warnings about deprecation or something you want users to stop doing.

But I don't quite get the point that why do you say `This doesn't seem like a good design.`
Deprecation warning is important for compatibility, for example:
If we want to remove the `index` parameter for Field class, we should triggers a warning for 0.24.x and then delete it at 0.25.0 or later version.

henadzit (approved):
LGTM!

Issue #937, opened 2021-10-04
Support real enums


numberZero:
Both MySQL and PostgreSQL support real type-safe enums, while TortoiseORM only provides some sugar to access a numerical or text field as if it was enum.

alex-klepa:
+1

Chr0nos:
https://www.postgresql.org/docs/current/datatype-enum.html

i've tried to implement it on my own but i'm facing 2 issues:
1: i was not able to find documentation on aerich to learn where the migration part of an implemented "Field" is made so i cannot implement my own, as a workaround i've just changed the migration manualy to create the field
2: pydantic_model_creator don't like it and throw a `pydantic.errors.PydanticUserError: Field 'seniority' requires a type annotation` error to me

the code that i've used:
```python
from enum import Enum
from typing import Type

from tortoise.fields import Field


def enum_type_factory(name: str, enum: Enum) -> Type[Field]:
    # https://www.postgresql.org/docs/current/datatype-enum.html
    class EnumField(Field):
        SQL_TYPE = name

        @classmethod
        def to_db_value(cls, value, instance) -> str:
            if isinstance(value, Enum):
                return value.value
            return value

        @classmethod
        def to_python_value(cls, value):
            if isinstance(value, Enum):
                return value
            return enum(value)

        def describe(self, serializable=False):
            description = super().describe(serializable)
            description.update({"enum": [item.value for item in enum]})
            return description

    return EnumField
```

it was able to interact with the models but not to use pydantic_model_creator

d3QUone:
+1

XaviTorello:
+1

Has this request been prioritized? Any open PR where we can contribute?

Many thanks

PR #1883, opened 2025-02-14
add real enum support for mysql/mariadb

Abdeldjalil-H:
## Description

Added MySQL-specific implementation for CharEnumField to properly support native ENUM types in MySQL databases. This implementation:
- Introduces a new `db_mysql` class for CharEnumFieldInstance
- Converts enum values to MySQL ENUM type definition
- Properly escapes enum values to prevent SQL injection
- Maintains type safety between Python Enum and MySQL ENUM

## Motivation and Context

Currently, CharEnumField doesn't utilize MySQL's native ENUM type, instead storing enum values as regular VARCHAR fields. This change improves:
- Data integrity by enforcing enum constraints at the database level
- Query performance by using MySQL's optimized ENUM type
- Schema clarity by explicitly showing valid enum values in database structure

## How Has This Been Tested?

`make test`

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,10 @@ Changelog
 
 0.25.0 (unreleased)
 ------
+Added
+^^^^^
+- Added MySQL-specific ENUM type support for CharEnumField (#1883)
+
 Fixed
 ^^^^^
 - Fixed asyncio "no current event loop" deprecation warning by replacing `asyncio.get_event_loop()` with modern event loop handling using `get_running_loop()` with fallback to `new_event_loop()` (#1865)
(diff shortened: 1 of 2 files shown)


Abdeldjalil-H:
Partially addresses #937

henadzit (changes requested):
Thanks for the contribution! However, I have multiple concerns about this change:

- This is a breaking change
- This is not supported by aerich which is used by a lot of people 
- This change will make the MySQL integration standout

If we decide to use the database enums, I think it should be a more complex change: 
- Implement this for all databases that support enums
- Update aerich to support enums 
- Make sure aerich can handle the migration to enums

Potentially we might have a configuration option on a field to regulate whether to use database enums to smooth the migration.

PR #1884, opened 2025-02-14
refactor: apply future type hints style

waketzheng:
## Description

1. Does not change any code logic
2. Add extend-select for ruff lint
```
    "FA",     # https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa
    "UP",     # https://docs.astral.sh/ruff/rules/#pyupgrade-up
    "RUF100", # https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf
```
3. Run the following script to auto add `from __future__ import annotations`:
```py
#!/usr/bin/env python
"""
Auto add 'from __future__ import annotations' to py files that imports `typing`
"""

import os
from pathlib import Path


def main() -> None:
    t = "from __future__ import annotations"
    package = Path(__file__).parent.name.replace("tortoise-orm", "tortoise")
    total = updated = 0
    for name in (package, "tests", "examples", "conftest.py"):
        if not (d := Path(name)).exists():
            print(f"Skip {d} as it not exists")
            continue
        for p in [d] if d.is_file() else d.rglob("*.py"):
            total += 1
            if t in (s := p.read_text()) or ("from typing " not in s and "import typing" not in s):
                continue
            updated += 1
            lines = s.splitlines()
            index = 0
            doc = False
            for idx, one in enumerate(lines):
                if one.startswith("#"):
                    continue
                if one.startswith('"""') or one.startswith("'''"):
                    if len(one) > 3:
                        # One line document
                        continue
                    if not doc:
                        doc = True
                        continue
                    doc = False
                    index = idx + 1
                    break
                elif not doc:
                    index = idx
                    break
            lines.insert(index, t)
            p.write_text(os.linesep.join(lines + [""]))
            print(f"Inject '{t}' into {p} at line {index + 1}")
    if not total:
        print("No target found, please check current directory!")
    elif not updated:
        print(f"None of file need to update({total} files found)")
    else:
        print(f"Summary: {total = }; {updated = }")


if __name__ == "__main__":
    main()
```
4. Execute `ruff check --fix --unsafe-fixes tortoise examples tests conftest.py` to reformat
5. Install autoflake and run its shell command to remove unused imports: 
    `autoflake --remove-all-unused-imports -i -r tortoise examples tests conftest.py` 
7. Reformat by `make _style`
8. Change `make lint` to be `make style`+`make check`
9. Refactor `tortoise/backends/base/schema_generator.py::BaseSchemaGenerator::_get_table_sql` to reduce complexity (which was complained by codacy)

## Motivation and Context

To improve contributing experience

## How Has This Been Tested?

make ci

## Checklist:


waketzheng:
@henadzit Cloud you review this?

CHANGELOG.rst
```diff
@@ -21,6 +21,7 @@ Fixed
 
 Changed
 ^^^^^^^
+- Apply future style type hints (#1878)
```
henadzit: Does this change anything from the package user perspective? If not, I don't think we need to have it in the CHANGELOG.

pyproject.toml
```diff
@@ -198,6 +198,11 @@ show_missing = true
 line-length = 100
 [tool.ruff.lint]
 ignore = ["E501"]
+extend-select = [
```
henadzit, line 201: Nice!

tortoise/backends/base/schema_generator.py
```diff
@@ -449,16 +364,104 @@ def _get_table_sql(self, model: "type[Model]", safe: bool = True) -> dict:
...
-            "table": model._meta.db_table,
+            "table": table_name,
             "model": model,
             "table_creation_string": table_create_string,
             "references": references,
             "m2m_tables": m2m_tables_for_create,
         }
 
-    def _get_models_to_create(self, models_to_create: "list[type[Model]]") -> None:
+    def _get_models_to_create(self, models_to_create: list[type[Model]]) -> None:
```
henadzit: I think this function should just return the list of models instead of modifying the argument. I find it hard to understand why it should be modifying the argument instead...

tortoise/backends/base/schema_generator.py
```diff
@@ -449,16 +364,104 @@ def _get_table_sql(self, model: "type[Model]", safe: bool = True) -> dict:
...
+                        column_name,
+                        self._escape_default_value(default),
+                        auto_now_add,
+                        auto_now,
+                    )
+                except NotImplementedError:
+                    pass
+        return ""
+
+    def _get_table_sql(self, model: type[Model], safe: bool = True) -> dict:
```
henadzit, line 394: There are two name patterns in this file:
- _get_*_sql
- _generate_*

What is the difference? Can we just use a single pattern?

tortoise/backends/base/schema_generator.py
```diff
@@ -449,16 +364,104 @@ def _get_table_sql(self, model: "type[Model]", safe: bool = True) -> dict:
...
+            field_object = model._meta.fields_map[field_name]
+            comment = self._generate_field_comment(field_object, table_name, column_name)
+            default = self._generate_field_default(field_object, table_name, column_name, model)
+
+            # TODO: PK generation needs to move out of schema generator.
+            if create_pk_field := self._generate_pk_create_sql(field_object, column_name, comment):
+                fields_to_create.append(create_pk_field)
+                continue
+
+            field_creation_string, related_table_name = self._generate_field_creation(
```
henadzit: `_generate_field_creation` doesn't seem to be valid English. Should it be something like `_get_field_sql`?

tortoise/backends/base/schema_generator.py
```diff
@@ -144,18 +144,13 @@ def _make_hash(*args: str, length: int) -> str:
...
         # characters (Oracle limit).
         # That's why we slice some of the strings here.
         table_name = model if isinstance(model, str) else model._meta.db_table
-        index_name = "{}_{}_{}_{}".format(
-            prefix,
-            table_name[:11],
-            field_names[0][:7],
-            self._make_hash(table_name, *field_names, length=6),
-        )
+        index_name = f"{prefix}_{table_name[:11]}_{field_names[0][:7]}_{self._make_hash(table_name, *field_names, length=6)}"
```
henadzit: This is quite hard to read. I think it would be better to move at least the hash from the f-string:

```
h = self._make_hash(table_name, *field_names, length=6)
index_name = f"{prefix}_{table_name[:11]}_{field_names[0][:7]}_{h}"
```

For instance, Django's code style disallows having expressions in f-strings, see https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/. Personally, I think simple expressions are fine but here we have quite a lot of complexity.

tortoise/backends/base/schema_generator.py
```diff
@@ -164,16 +159,12 @@ def _generate_fk_name(
         # NOTE: for compatibility, index name should not be longer than 30
         # characters (Oracle limit).
         # That's why we slice some of the strings here.
-        index_name = "fk_{f}_{t}_{h}".format(
-            f=from_table[:8],
-            t=to_table[:8],
-            h=self._make_hash(from_table, from_field, to_table, to_field, length=8),
-        )
+        index_name = f"fk_{from_table[:8]}_{to_table[:8]}_{self._make_hash(from_table, from_field, to_table, to_field, length=8)}"
```
henadzit: same

henadzit (commented):
Hey, i reviewed probably a half of this PR and left a few comments. Hopefully, I'll check out the other half in the coming days.

tortoise/backends/base/schema_generator.py
waketzheng: Use `get_xxx` for sync function, while `generate_*` for async function.

tortoise/backends/base/schema_generator.py
waketzheng: Updated.
Use template `{prefix}_{table}_{field}_{hash}` and `pk_{f}_{t}_{h}`

tortoise/backends/base/schema_generator.py
waketzheng: Changed to be `_get_field_sql_and_related_table`

tortoise/backends/base/schema_generator.py
```diff
@@ -143,54 +144,48 @@ def _make_hash(*args: str, length: int) -> str:
...
         # That's why we slice some of the strings here.
+        template = "{prefix}_{table}_{field}_{hash}"
         table_name = model if isinstance(model, str) else model._meta.db_table
-        index_name = "{}_{}_{}_{}".format(
-            prefix,
-            table_name[:11],
-            field_names[0][:7],
-            self._make_hash(table_name, *field_names, length=6),
+        hash_ = self._make_hash(table_name, *field_names, length=6)
+        return template.format(
```
henadzit: I think it would be much more readable if the template wasn't separated by multiple lines from actual formatting.

tortoise/backends/base/schema_generator.py
```diff
@@ -143,54 +144,48 @@ def _make_hash(*args: str, length: int) -> str:
...
         # That's why we slice some of the strings here.
-        index_name = "fk_{f}_{t}_{h}".format(
-            f=from_table[:8],
-            t=to_table[:8],
-            h=self._make_hash(from_table, from_field, to_table, to_field, length=8),
-        )
-        return index_name
+        template = "fk_{f}_{t}_{h}"
+        hash_ = self._make_hash(from_table, from_field, to_table, to_field, length=8)
+        return template.format(f=from_table[:8], t=to_table[:8], h=hash_)
```
henadzit: same

henadzit (commented):
Left a few minor comments but apart from that I think it's good to go!

henadzit (approved):
LGTM!

Issue #1521, opened 2023-11-30
Initializer doesn't support multiple apps


primal100:
**Describe the bug**
Initalizer only supports one app label in the arguments to passed to it. In trying to run it multiple times it fails at the first app giving an error for the other apps.

```python
    raise ConfigurationError(
tortoise.exceptions.ConfigurationError: default_connection for the model <class 'app2.models.Category'> cannot be None
```
 This makes it difficult to test when there are multiple apps.

**To Reproduce**
```app1/models.py```:

```python
from tortoise.models import Model
from tortoise import fields

class Users(Model):
    id = fields.IntField(pk=True)
    password = fields.CharField(max_length=128, blank=True, null=True)
    email = fields.CharField(max_length=120, unique=True, index=True)
```

```app2/models.py```:

```python
from tortoise.models import Model
from tortoise import fields

class Category(Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(max_length=128)
    user = fields.ForeignKeyField('app1.Users')
```

```main.py```:

```python
import asyncio
import os
from tortoise import Tortoise
from tortoise.contrib.test import finalizer, initializer

db_url = "sqlite://:memory:?cache=shared"

def setup_db():
    Tortoise.init_models(["app1.models"], "app1")
    Tortoise.init_models(["app2.models"], "app2")
    print('Initializing app1')
    initializer(['app1.models'], db_url=db_url, app_label="app1")
    print('Initializing app2')
    initializer(['app2.models'], db_url=db_url, app_label="app2")
    print('DB initialized')
    finalizer()
    print('finalized')

try:
    setup_db()
    print('setup_db done')
finally:
    asyncio.run(Tortoise.close_connections())
```

At the line where app1 is run through initializer the following error is given:

```python
    raise ConfigurationError(tortoise.exceptions.ConfigurationError: default_connection for the model <class 'app2.models.Category'> cannot be None
```

The following also fails with the same error. Here both apps ```models.py``` are provided as modules, but there is still the restriction of only one app label.

```python
def setup_db():
    Tortoise.init_models(["app1.models"], "app1")
    Tortoise.init_models(["app2.models"], "app2")
    initializer(['app1.models', 'app2.models'], db_url=db_url, app_label="app1")
    print('DB initialized')
    finalizer()
    print('finalized')
```

```python
tortoise.exceptions.ConfigurationError: default_connection for the model <class 'app2.models.Category'> cannot be Non
```

**Expected behavior**
[...]

PR #1895, opened 2025-02-25
Support multiple apps in testing initializer

primal100:
## Description
Allows the intializer to accept a mapping of app names to module list, allowing more flexibility when testing, and allowing tests to run on models across multiple apps. This would also allow models to work where they have a foreign key to a model in another app. The previous modules list and app label still works as normal so compatibility will not be affected. If a dict of app names to modules is provided, the app_label kw argument is not used.

## Motivation and Context
Particularly useful when integrating models from a third party library into your own application. 
Issue created a while back: https://github.com/tortoise/tortoise-orm/issues/1521

## How Has This Been Tested?

This is something that needs to be discussed as at the moment intializer and finalizer is run as an autouse session fixture, so difficult to override for a single test. Coverage is not an issue as there are no new lines.

Possibilities include:
* Changing `initialize_tests` fixture to use the new apps/module dict format in which case the older single list format would not be tested.
* Create a completely separate tests directory which ignores the main conftest.py. Probably the main conftest.py would have to be moved into the current tests directory. The new tests directory would have it's own conftest.py using the new style initializer argument format

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,12 @@ Changelog
 0.24
 ====
 
+0.24.2
+------
+Added
+^^^^^
+- Allow mapping of modules for multiple apps to be provided to tortoise.contrib.test.initializer (#1895)
+
 0.24.1
 ------
 Added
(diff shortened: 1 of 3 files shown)


primal100:
Some checks are failing due to 

```
UP007 Use `X | Y` for type annotations
```

However, this newer way of writing Union types was introduced in Python 3.10 and tortoise-orm still allows support for 3.9, so that's why I still used typing.Union.

waketzheng:
> Some checks are failing due to
> 
> ```
> UP007 Use `X | Y` for type annotations
> ```
> 
> However, this newer way of writing Union types was introduced in Python 3.10 and tortoise-orm still allows support for 3.9, so that's why I still used typing.Union.

With this line `from __future__ import annotations` at the beginning of the file, you are able to use `X | Y` for Python3.9

PR #1903, opened 2025-02-28
refactor: use 'unique' instead of 'create_unique_index' for m2m field

waketzheng:
## Description

Related issue:  https://github.com/tortoise/aerich/issues/406


tortoise/fields/relational.py
```diff
@@ -374,6 +374,7 @@ def describe(self, serializable: bool) -> dict:
         desc["backward_key"] = self.backward_key
         desc["through"] = self.through
         desc["on_delete"] = str(self.on_delete)
+        desc["create_unique_index"] = self.create_unique_index
```
henadzit: Oh, why is it called `create_unique_index` if we use just `unique` for other fields? It feels not clean to have different names for the same configuration options.

Before we propagate this naming to aerich, I suggest we rename `create_unique_index` to `unique` but keep `create_unique_index` deprecated for backwards compatibility.

tortoise/models.py
```diff
@@ -1535,46 +1535,46 @@ def describe(cls, serializable: bool = True) -> dict:
 
             Each field is specified as defined in :meth:`tortoise.fields.base.Field.describe`
         """
+        fields: dict[str, list[dict]] = {
```
henadzit: Does it really improve performance? What is the impact? I think if you want to make a point that this improves performance, you need to add a benchmark as a separate PR and then provide the comparison. 

My honest opinion is that this change makes very readable code unreadable.

henadzit (changes requested):

tortoise/models.py
waketzheng: Readable is more important.

tortoise/fields/relational.py
waketzheng: Agree. Will do it later.

tortoise/fields/relational.py
```diff
@@ -582,10 +582,18 @@ def ManyToManyField(
...
         The default is True. If you want to allow repeat records, set this to False.
     """
+    if "create_unique_index" in kwargs:
+        create_unique_index = kwargs.pop("create_unique_index")
+        msg = "Parameter `create_unique_index` is deprecated! Use `unique` instead."
+        if create_unique_index is False and unique is True:  # For backwards compatibility
+            warnings.warn(msg, DeprecationWarning, stacklevel=2)
+            unique = False
+        elif create_unique_index is True and unique is False:  # Invalid usage
+            raise ConfigurationError(msg)
```
henadzit: And if both `create_unique_index` and `unique` are True?

I think a simpler version should be fine:

```
if "create_unique_index" in kwargs:
        unique = kwargs.pop("create_unique_index")
        warnings.warn("Parameter `create_unique_index` is deprecated! Use `unique` instead.", DeprecationWarning, stacklevel=2)
```

tortoise/fields/relational.py
```diff
@@ -350,12 +351,12 @@ def __init__(
...
         on_delete: OnDelete = CASCADE,
         field_type: type[MODEL] = None,  # type: ignore
-        create_unique_index: bool = True,
+        unique: bool = True,
         **kwargs: Any,
     ) -> None:
         # TODO: rename through to through_table
         # TODO: add through to use a Model
-        super().__init__(field_type, **kwargs)
+        super().__init__(field_type, unique=unique, **kwargs)
```
henadzit, line 366: Why do you need to pass unique explicitly here?

waketzheng: Because ManyToManyFieldInstance inherit from `class Field`, which receive the `unique` parameter.

henadzit (approved):
LGTM! Although we should also update CHANGELOG, however, we can do it when doing a release.

PR #1904, opened 2025-02-28
[WIP] Fix connection from router edgecase

i-am-grub:
## Description

Fix for preventing the `_router_func` method from passing a `NoneType` to `connections.get`.

## Motivation and Context

I was receiving a `TypeError: 'NoneType' object is not iterable` when using a configuration based on what is described in the [Two Databases](https://tortoise.github.io/examples/basic.html#two-databases) documentation. The exception was occurring when using `Model.create` (in a subclass) without specifying a database to use. The exception was traced up though `Model._choose_db()`.

## How Has This Been Tested?

No additional tests have been conducted besides verifying that my use case is now working as expected.

## Checklist:

### DIFF
--- a/tortoise/router.py
+++ b/tortoise/router.py
@@ -30,10 +30,14 @@ class ConnectionRouter:
                     return chosen_db
 
     def _db_route(self, model: type[Model], action: str) -> BaseDBAsyncClient | None:
-        try:
-            return connections.get(self._router_func(model, action))
-        except ConfigurationError:
-            return None
+        conn_alias = self._router_func(model, action)
+        if conn_alias is not None:
+            try:
+                return connections.get(conn_alias)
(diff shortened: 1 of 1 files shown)


henadzit:
@i-am-grub, thanks for the contribution. Can you please do the following things to get your change merged?

- Provide a test case exposing the issue
- Provide a example of the configuration that caused the issue for you

Thanks

i-am-grub:
Turns out I'm the problem. I was using the toolkit library for parsing a config file in my application and then passing it directly to Tortoise.init

I originally thought that the object generated from parsing the file was a dict, but it isn't. The object's interface nearly matches a dict but not completely. This was likely why the original error was occurring in such an odd location.

This PR can be closed without merging unless there is another reason to have it merged.

PR #1911, opened 2025-03-01
on_delete should be required in all type relations #1801.

### LINKED ISSUE #1801: on_delete should be required

FluckyGo:
My changes remove the default CASCADE behavior for the on_delete parameter in all relationship types, making it mandatory and the behavior obvious and predictable.
This will help prevent accidental data loss when implicitly cascading deletes.
All tests have been updated to explicitly specify the on_delete value.

Fixes https://github.com/tortoise/tortoise-orm/issues/1801

Description
This helps prevent accidental data loss caused by implicit cascading deletes, which developers might overlook due to the default CASCADE behavior. By requiring on_delete to be explicitly defined, we ensure that deletion behavior is always clear and intentional..

Motivation and Context
Issue [#1801](https://github.com/tortoise/tortoise-orm/issues/1801) highlighted the default on_delete=CASCADE behavior as a potential risk for data loss. Users reported that unintended cascading deletions could occur, particularly when developers were unaware of the default setting. To enhance data safety, this change now requires explicit configuration of the deletion behavior.

How Has This Been Tested?
Ran the full test suite using pytest.
Tested on Python 3.9, 3.10, 3.11, 3.12, 3.13 with SQLite/PostgreSQL.
## Checklist:
* [x]  My code follows the code style of this project.
* [x]  My change requires a change to the documentation.
* [ ]  I have updated the documentation accordingly.
* [x]  I have added the changelog accordingly.
* [x]  I have read the **CONTRIBUTING** document.
* [x]  I have added tests to cover my changes.
* [x]  All new and existing tests passed.


FluckyGo:
This PR is designed to get rid of the implicit behavior when dealing with any type of relations.

PR #1897, opened 2025-02-25
fix: model with multi m2m fields generates wrong references name

waketzheng:
## Description

Fixes #1893 

## Motivation and Context

The new function `_get_m2m_tables` has argument `table_name` which is conflict with variable defined at https://github.com/tortoise/tortoise-orm/blob/0.24.1/tortoise/backends/base/schema_generator.py#L327

## How Has This Been Tested?

make ci

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added the changelog accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1885, opened 2025-02-15
chore: use project section

waketzheng:
## Description

Use the standard `[project]` section, and move dependencies from `[tool.poetry.dependencies]` section to `project.dependencies`.

## Motivation and Context

Fixes #1898 

## How Has This Been Tested?

make ci

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added the changelog accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1914, opened 2025-03-03
Release 0.24.2

henadzit:
Fixed
^^^^^
- Fix model with multi m2m fields generates wrong references name (#1897)
- Fix using reserved words in order_by (#1900)
- Fix installing tortoise-orm with poetry 2 (#1885)

Changed
^^^^^^^
- Use 'unique' instead of 'create_unique_index' for m2m field (#1903)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,13 +9,18 @@ Changelog
 0.24
 ====
 
-0.24.2 (Unreleased)
+0.24.2
 ------
 
 Fixed
 ^^^^^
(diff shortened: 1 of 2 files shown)


waketzheng (changes requested):
Changed
^^^^^
- Use 'unique' instead of 'create_unique_index' for m2m field (#1903)

henadzit:
@waketzheng updated!

waketzheng (approved):

PR #1917, opened 2025-03-04
fix: sanic example raises no such table

### LINKED ISSUE #1909: Error creating tables in the sqlite database
Starting with tortoise 0.22, the documentation examples do not work correctly: tables are not created in the sqlite database.
Tested with Sanic and FastAPI examples.
Error: tortoise.exceptions.OperationalError: no such table: users.
After manually creating the table, the code works correctly.
Tested on tortoise ORM versions 0.24, 0.23, 0.22, 0.21.

waketzheng:
## Description

Fixes #1909

## Motivation and Context

1. Use `db.sqlite3` instead of memory sqlite for sanic example to fix no such table error
2. Add unittest for sanic example

## How Has This Been Tested?

make ci

## Checklist:

### DIFF
--- a/examples/sanic/main.py
+++ b/examples/sanic/main.py
@@ -17,16 +17,16 @@ async def list_all(request):
     return response.json({"users": [str(user) for user in users]})
 
 
-@app.route("/user")
+@app.post("/user")
 async def add_user(request):
     user = await Users.create(name="New User")
     return response.json({"user": str(user)})
 
 
 register_tortoise(
-    app, db_url="sqlite://:memory:", modules={"models": ["models"]}, generate_schemas=True
+    app, db_url="sqlite://db.sqlite3", modules={"models": ["models"]}, generate_schemas=True
 )
 
 
 if __name__ == "__main__":
-    app.run(port=5000)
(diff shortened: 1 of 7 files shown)


examples/sanic/main.py
```diff
@@ -17,16 +17,16 @@ async def list_all(request):
...
-@app.route("/user")
+@app.post("/user")
 async def add_user(request):
     user = await Users.create(name="New User")
     return response.json({"user": str(user)})
 
 
 register_tortoise(
-    app, db_url="sqlite://:memory:", modules={"models": ["models"]}, generate_schemas=True
+    app, db_url="sqlite://db.sqlite3", modules={"models": ["models"]}, generate_schemas=True
```
henadzit, line 27: Why was it failing with `sqlite://:memory:`? I see `generate_schemas`, so the schema should have been created, right?

waketzheng: I don't know why it failed. It is hard for me to find out the reason. As this is an example, I believe that `db.sqlite3` is better that `:memory:`.

henadzit (approved):

PR #1923, opened 2025-03-25
Support relations in .only()

### LINKED ISSUE #916: use only for nested model fields
Hi There!

How to make `only` for nested model fields
Example: 

```python
class A:
    id = fields.IntField()
    name = fields.CharField()
    other_field = fields.CharField()

class B:
    id = fields.IntField()
    b_name = fields.CharField()
    a = fields.ForeignKeyField('models.A', related_name='Bs', on_delete=fields.CASCADE)
    
    
B.all().select_related('a').only('id', 'b_name', 'a__name')
```

This example will throw 
```python
KeyError: 'a__name'
```

### LINKED ISSUE #1918: 如何同时使用only和select_related
怎么在only中指定select_related中表的字段
是否目前还未实现该功能
class A(Model):
    id = fields.IntField(primary_key=True, description="ID")
    name = fields.CharField(description="name", max_length=50)

class B(Model):
    id = fields.IntField(primary_key=True, description="ID")
    a = fields.ForeignKeyField("models.A", description="外键A", on_delete=fields.CASCADE)

eg:
  await B.filter().select_related("a").only("id", "a__name") # just secelt B.id and A.name

henadzit:
## Description
- `.only` supports selecting related fields, e.g. `.only("related__field")`
- Add a check that `.values()`, `.values_list()` and `.only()` cannot be used together
- Add `expand_lookup_expression` that can be used to convert the lookup expression (`a__b__c`) to a list of fields. Currently it is used in `_join_select_related` and `_resolve_only` but it can be reused in other places that support lookup expressions, for instance, in filters and ordering.

## Motivation and Context
- Will resolve https://github.com/tortoise/tortoise-orm/issues/1918
- Will resolve https://github.com/tortoise/tortoise-orm/issues/916

## How Has This Been Tested?
`make ci` + a lot of new tests

## Checklist:

### DIFF
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -114,7 +114,7 @@ class BaseExecutor:
                 for model, index, *__, full_path in self.select_related_idx[1:]:
                     (*path, attr) = full_path
                     related_items = row_items[current_idx : current_idx + index]
-                    if any((v for _, v in related_items)):
+                    if any(v for _, v in related_items):
                         obj = model._init_from_db(**{k.split(".")[1]: v for k, v in related_items})
                     elif index == 0:
                         # 0 signals that an empty "filler" object should be created in the case
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -114,10 +114,12 @@ class BaseExecutor:
                 for model, index, *__, full_path in self.select_related_idx[1:]:
                     (*path, attr) = full_path
                     related_items = row_items[current_idx : current_idx + index]
-                    if not any((v for _, v in related_items)):
-                        obj = None
-                    else:
+                    if any((v for _, v in related_items)):
                         obj = model._init_from_db(**{k.split(".")[1]: v for k, v in related_items})
+                    elif index == 0:
+                        obj = model._init_from_db()
+                    else:
+                        obj = None
                     target = instances.get(tuple(path))
                     if target is not None:
                         setattr(target, f"_{attr}", obj)
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -117,6 +117,9 @@ class BaseExecutor:
                     if any((v for _, v in related_items)):
                         obj = model._init_from_db(**{k.split(".")[1]: v for k, v in related_items})
                     elif index == 0:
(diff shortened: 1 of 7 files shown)


tortoise/queryset.py
```diff
@@ -1087,8 +1166,7 @@ def _make_query(self) -> None:
                 self.model,
                 (None,),
             )
-            if append_item not in self._select_related_idx:
```
henadzit, line 1090: This check is not needed since `self._select_related_idx` is always empty at this point.

tortoise/query_utils.py
```diff
@@ -78,6 +79,23 @@ def get_joins_for_related_field(
...
+    fields: list[Field | RelationalField] = []
+    model = root_model
+    for field_name in field_names[:-1]:
+        if field_name not in model._meta.fetch_fields:
+            raise FieldError(f"{lookup_expression} not resolvable")
+        field = cast(RelationalField, model._meta.fields_map[field_name])
+        fields.append(field)
+        model = field.related_model
+    try:
+        fields.append(model._meta.fields_map[field_names[-1]])
```
waketzheng, line 94: Cloud you leave some comment for why put the last element out of the for loop?

tortoise/backends/base/executor.py
```diff
@@ -114,10 +114,15 @@ async def execute_select(
                 for model, index, *__, full_path in self.select_related_idx[1:]:
                     (*path, attr) = full_path
                     related_items = row_items[current_idx : current_idx + index]
-                    if not any((v for _, v in related_items)):
-                        obj = None
-                    else:
+                    if any((v for _, v in related_items)):
```
waketzheng: I think
```py
if any(v for _, v in related_items):
```
is more readable

henadzit: `if any(v for _, v in related_items):` should be faster too because a new tuple isn't created. Updated!

tortoise/query_utils.py
henadzit: updated!

waketzheng (approved):

tortoise/backends/base/executor.py
waketzheng: Actually, `(v for _, v in related_items)` is a generator.

PR #1949, opened 2025-05-14
Add SELECT ... FOR NO KEY UPDATE support for Postgres

Pankrat:
## Description

Allow slightly reducing the lock level when using `select_for_update` on Postgres DBs. In contrast to the classic `SELECT FOR UPDATE` this lock level allows concurrent inserts and deletes on tables that have foreign key references to the locked rows.

The parameter follow the Django ORM design:

https://code.djangoproject.com/ticket/30375

This depends on these changes in pypika-tortoise: https://github.com/tortoise/pypika-tortoise/pull/29

Please see the linked PR for an alternative approach that exposes lock strength directly (which would allow supporting `SELECT FOR SHARE` and `SELECT FOR KEY SHARE` as well but lose the "compatibility" with Django ORM). 

## Motivation and Context

I'm currently working on a project that suffers from lock contention issues blocking INSERTs and DELETEs where SELECT FOR UPDATE is involved. The SELECT FOR UPDATE is often used in a transaction where certain columns are modified for the locked rows but the primary key is untouched and the rows are not deleted. Since I'm using Postgres, `SELECT FOR NO KEY UPDATE` would be a simple way to reduce lock contention slightly. However, this is not supported by Tortoise so I'd have to fall back to raw queries which is somewhat cumbersome.

## How Has This Been Tested?

I've used the test suite of tortoise-orm and pypika-tortoise and ran test against PostgreSQL and MySql and MacOS with docker:

```
# Start test databases
docker run --name tortoise-test-postgres -e POSTGRES_PASSWORD=123456 -p 5432:5432 -d postgres
docker run --name tortoise-test-mysql -e MYSQL_ROOT_PASSWORD=123456 -p 3306:3306 -d mysql

# Use modified version of pypika-tortoise
pip install git+https://github.com/Pankrat/pypika-tortoise.git@select-for-no-key-update

# Run tests (seems to reset pypika-tortoise, need to re-run pip install command above before each run)
pip install cryptography
make test_mysql
make test_postgres_asyncpg
make test_postgres_psycopg
```

## Checklist:

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-from collections import defaultdict
 import types
+from collections import defaultdict
 from collections.abc import AsyncIterator, Callable, Collection, Generator, Iterable
 from copy import copy
 from typing import TYPE_CHECKING, Any, Generic, Optional, TypeVar, cast, overload
@@ -326,6 +326,7 @@ class QuerySet(AwaitableQuery[MODEL]):
         "_select_for_update_nowait",
         "_select_for_update_skip_locked",
         "_select_for_update_of",
+        "_select_for_update_no_key",
         "_select_related",
         "_select_related_idx",
         "_use_indexes",
@@ -351,6 +352,7 @@ class QuerySet(AwaitableQuery[MODEL]):
         self._select_for_update_nowait: bool = False
         self._select_for_update_skip_locked: bool = False
         self._select_for_update_of: set[str] = set()
+        self._select_for_update_no_key: bool = False
         self._select_related: set[str] = set()
         self._select_related_idx: list[
             tuple[type[Model], int, Table | str, type[Model], Iterable[str | None]]
@@ -385,6 +387,7 @@ class QuerySet(AwaitableQuery[MODEL]):
         queryset._select_for_update_nowait = self._select_for_update_nowait
         queryset._select_for_update_skip_locked = self._select_for_update_skip_locked
         queryset._select_for_update_of = self._select_for_update_of
+        queryset._select_for_update_no_key = self._select_for_update_no_key
         queryset._select_related = self._select_related
         queryset._select_related_idx = self._select_related_idx
         queryset._force_indexes = self._force_indexes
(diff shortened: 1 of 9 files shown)


tortoise/queryset.py
```diff
@@ -572,7 +575,11 @@ def distinct(self) -> QuerySet[MODEL]:
         return queryset
 
     def select_for_update(
-        self, nowait: bool = False, skip_locked: bool = False, of: tuple[str, ...] = ()
+        self,
```
henadzit, line 578: It isn't a blocker in any way but if you have time, could you please document the meaning of the arguments?

henadzit (commented):
The change looks good to me! Thanks for the contribution. 

The CI is failing though, can you please have a look into it?

tortoise/queryset.py
Pankrat: Thanks for the review! I've added some documentation for the arguments.

I think the build is currently failing because it requires the pypika-tortoise changes and the test run installs v0.5.0 which doesn't have those changes yet. I've updated the version requirements in the PR, but it gets blocked and won't build until a new pypika-tortoise version is released.

Alternatively, we could add version checks to provide backwards compatibility with older versions of pypika-tortoise, but that doesn't seem very attractive to me.

henadzit: I'll release the new version of pypika to day, and will let you know! Thanks!

henadzit: I just released pypika-tortoise 0.6.0 with your changes.

Pankrat: Thanks! I've updated the version requirement and lock file and re-ran the tests locally. Not sure if/how I can re-trigger the CI tests.

waketzheng: @henadzit Let's clean the PRs of pypika-tortoise and release a new version. Do you want to continue this https://github.com/tortoise/pypika-tortoise/pull/21, or let me pick it up?

henadzit (commented):
Looks good! Thank you for the contribution!

henadzit (dismissed):

tortoise/queryset.py
henadzit: @Pankrat sorry, can you please update the poetry.lock file? There was another PR that updated dependencies and caused a conflict.

@waketzheng please feel free to pick it up!

Pankrat: @henadzit I've rebased the PR on the develop branch. If you roll out additional changes for pypika and there are conflicts again, one could also drop the version update commit here if the pyproject.toml is updated in the develop branch, or just cherry-pick the feature commit.

henadzit (approved):

PR #1958, opened 2025-05-29
docs: Update Getting Started installation commands

HSM95:
## Description
zsh misinterprets square brackets, but adding quotation marks resolves that.

## Motivation and Context
This is a minor change to avoid slight friction when someone wants to install the package with a non-default database.

## How Has This Been Tested?
Tested by successfully running the installation command with quotation marks.

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,6 @@ Changelog
 Changed
 ^^^^^
 - Force async task switch every 2000 rows when converting db objects to python objects to avoid blocking the event loop (#1939)
-- Update installation commands to be zsh-friendly (#1958)
 
 Added
 ^^^^^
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ Changelog
 Changed
 ^^^^^
 - Force async task switch every 2000 rows when converting db objects to python objects to avoid blocking the event loop (#1939)
+- Update installation commands to be zsh-friendly (#1958)
```
henadzit: Change log is intended primarily for reviewing the changes when upgrading to a new version of the package. We don't really need documentation changes in the change log.

henadzit (commented):
Thank you for the contribution! I left a comment.

CHANGELOG.rst
henadzit:
```suggestion
```

henadzit (approved):

CHANGELOG.rst
HSM95: Yes, makes sense. I was unsure of whether I should add it but erred on the side of caution.

I think in that case the pull request template checklist should mention that doc-only changes don't require updating the changelog.

PR #1963, opened 2025-06-06
refactor: migrate lint tool from isort+black to ruff

waketzheng:
## Description

1. ruff is faster that isort+black when `make style`, and we should reduce dev deps.
2. remove coveralls from dev deps as it is only used by ci and it's installed by https://github.com/tortoise/tortoise-orm/blob/cb72b8afd0204d71bf851c22d90557581766a79c/.github/workflows/ci.yml#L85
3. replace `assert ...` by `raise ConfigurationError` in `tortoise/__init__.py` to fix bandit complaint
4. reformatted code by `make style`
5. move some deps from dev to new groups, and then lock and upgrade deps

## Motivation and Context

Clean dependencies

## Checklist:


henadzit (approved):
LGTM!

Issue #1969, opened 2025-06-16
Incorrect JOIN aliasing when filtering on multiple ManyToMany relations to the same table


sietr0s:
**Title:** Incorrect JOIN aliasing when filtering on multiple ManyToMany relations to the same table

**Description:**
When making a query that filters on two different ManyToMany relations pointing to the same model (through different through tables), Tortoise ORM fails to properly alias the JOINs, which leads to incorrect SQL generation.

**Steps to Reproduce:**
1. Define models with two ManyToMany relations to the same model:
```python
class Category(Model):
    name = fields.CharField(max_length=255)

class Profile(Model):
    benefits = fields.ManyToManyField("models.Category", through="profile_benefits")
    intentions = fields.ManyToManyField("models.Category", through="profile_intentions")
```

2. Try to query profiles that have both a benefit with name "1" and an intention with name "2":
```python
Profile.filter(benefits__name="1").filter(intentions__name="2")
```

**Expected Behavior:**
Tortoise ORM should generate proper SQL with aliases for each JOIN to distinguish between the two relations to the same table, something like:
```sql
SELECT ... 
FROM profile
JOIN profile_benefits ON profile.id = profile_benefits.profile_id
JOIN category AS benefits_category ON profile_benefits.category_id = benefits_category.id
JOIN profile_intentions ON profile.id = profile_intentions.profile_id
JOIN category AS intentions_category ON profile_intentions.category_id = intentions_category.id
WHERE benefits_category.name = '1' AND intentions_category.name = '2'
```

**Actual Behavior:**
Tortoise ORM generates SQL without proper table aliasing, causing ambiguity in the JOINs. The generated SQL might look like:
```sql
SELECT ... 
FROM profile
JOIN profile_benefits ON profile.id = profile_benefits.profile_id
JOIN category ON profile_benefits.category_id = category.id
JOIN profile_intentions ON profile.id = profile_intentions.profile_id
JOIN category ON profile_intentions.category_id = category.id
WHERE category.name = '1' AND category.name = '2'
```

sietr0s:
## Root Cause
The `get_joins_for_related_field` function didn't apply table aliasing for `ManyToManyFieldInstance` cases, unlike other relation types which already had proper aliasing.

## Solution
Modified the function to add proper table aliasing for ManyToMany relations:

```python
def get_joins_for_related_field(
    table: Table, 
    related_field: RelationalField, 
    related_field_name: str
) -> list[TableCriterionTuple]:
[...]

PR #1972, opened 2025-06-24
Added m2m related table alias

sietr0s:
Fix incorrect JOIN aliasing when filtering on multiple ManyToMany relations to the same table

## Description
This PR fixes an issue where Tortoise ORM fails to properly alias JOINs when filtering on two different ManyToMany relations pointing to the same model. The problem occurred because `get_joins_for_related_field` function didn't apply table aliasing for ManyToManyFieldInstance cases.

The main change modifies the `get_joins_for_related_field` function to add proper table aliasing for ManyToMany relations by using the pattern `{table_name}__{field_name}` for the alias.

## Motivation and Context
When making queries that filter on two different ManyToMany relations to the same model (through different through tables), Tortoise ORM was generating SQL without proper table aliasing, causing ambiguity in the JOINs. This made such queries impossible to execute correctly.

Fixes issue #1969

### DIFF
--- a/tortoise/query_utils.py
+++ b/tortoise/query_utils.py
@@ -125,6 +125,8 @@ def resolve_nested_field(
             related_table = related_table.as_(
                 f"{table.get_table_name()}__{iter_field.model_field_name}"
             )
+        if isinstance(related_field, ManyToManyFieldInstance):
+            related_table = related_table.as_(f"{table.get_table_name()}__{iter_field.model_field_name}")
         table = related_table
 
     last_field = fields[-1]
@@ -142,6 +144,10 @@ def resolve_nested_field(
                 related_table = related_table.as_(
                     f"{table.get_table_name()}__{related_field.model_field_name}"
                 )
+        if isinstance(related_field, ManyToManyFieldInstance):
+            related_table = related_table.as_(
+                f"{table.get_table_name()}__{related_field.model_field_name}"
+            )
 
         term = related_table[related_field_meta.db_pk_column]
     else:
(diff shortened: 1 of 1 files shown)


henadzit:
@sietr0s thank you for the contribution! Can you please add a test exposing the issue? This requirement was a part of the PR template :) 

Update: there are failing tests.

tortoise/query_utils.py
```diff
@@ -124,6 +125,10 @@ def resolve_nested_field(
             related_table = related_table.as_(
                 f"{table.get_table_name()}__{iter_field.model_field_name}"
             )
+        if isinstance(related_field, ManyToManyFieldInstance):
+            related_table = related_table.as_(
```
henadzit, line 129: The code duplicate the code of the check above. You can rewrite it as 

```python
if isinstance(related_field, (ForeignKeyFieldInstance, ManyToManyFieldInstance)):
```

tortoise/query_utils.py
```diff
@@ -141,6 +146,10 @@ def resolve_nested_field(
                 related_table = related_table.as_(
                     f"{table.get_table_name()}__{related_field.model_field_name}"
                 )
+        if isinstance(related_field, ManyToManyFieldInstance):
```
henadzit, line 149: `elif` should prevent a redundant check.

henadzit (commented):
Can you please add a test exposing the original issue? Meaning that the new test should fail without your changes. This is a requirement for a PR to be merged. Thanks!

abondar:
Repacked fix as https://github.com/tortoise/tortoise-orm/pull/2083

PR #2020, opened 2025-11-23
fix: sqlite decimal filter error with `__gt`

### LINKED ISSUE #2017: The result of querying the Decimal field through the filter interface does not meet expectations
**Describe the bug**

There are two questions in total:
1. Query results include Scientific notation
2. Based on the query results, it can be evaluated that the filter interface filters based on char

**To Reproduce**

```python
from decimal import Decimal
from typing import Callable, Coroutine
from tortoise import models, fields, run_async
from decimal import Decimal


class Item(models.Model):
    dec = fields.DecimalField(max_digits=100, decimal_places=2)


def run_in_memory_sqlite(func: Callable[..., Coroutine]) -> None:
    run_async(init_memory_sqlite(func)())


async def main():
    item1 = Item(dec=Decimal("400.00"))
    await item1.save()

    item2 = Item(dec=Decimal("90.01"))
    await item2.save()

    gte_100 = await Item.filter(dec__gte=Decimal("100")).all()

    print("get_100 result:")
    for _ in gte_100:
        print(Decimal(_.dec))

    lte_50 = await Item.filter(dec__lte=Decimal("50")).all()
    print("lte_50 result:")
    for _ in lte_50:
        print(Decimal(_.dec))


if __name__ == "__main__":
    run_in_memory_sqlite(main)

```

**Expected behavior**

**Additional context**
Execution result：
```cmd 
get_100 result:
4E+2
90.01
lte_50 result:
4E+2
```

waketzheng:
## Description

Fixes #2017 

## Motivation and Context

image

`function_cast` of DecimalField for sqlite does not used when querying. BUG fixed when this function  was called as expected.

## How Has This Been Tested?

make ci

## Checklist:

### DIFF
--- a/tortoise/expressions.py
+++ b/tortoise/expressions.py
@@ -387,6 +387,7 @@ class Q:
         else:
             filter_info = model._meta.get_filter(key)
 
+        field_object = None
         if "table" in filter_info:
             # join the table
             join = (
@@ -405,7 +406,12 @@ class Q:
                 else field_object.to_db_value(value, model)
             )
         op = filter_info["operator"]
-        criterion = op(table[filter_info.get("source_field", filter_info["field"])], value)
+        term: Term = table[filter_info.get("source_field", filter_info["field"])]
+        if func := field_object and field_object.get_for_dialect(
+            model._meta.db.capabilities.dialect, "function_cast"
+        ):
+            term = func(field_object, term)
+        criterion = op(term, value)
         return criterion, join
 
     def _resolve_regular_kwarg(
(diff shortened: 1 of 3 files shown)


tortoise/expressions.py
```diff
@@ -405,7 +406,12 @@ def _process_filter_kwarg(
                 else field_object.to_db_value(value, model)
             )
         op = filter_info["operator"]
-        criterion = op(table[filter_info.get("source_field", filter_info["field"])], value)
+        term: Term = table[filter_info.get("source_field", filter_info["field"])]
+        if func := field_object and field_object.get_for_dialect(
```
henadzit: Can you please help me to understand why do we need to assign field_object to func?

waketzheng: OK, updated. Code refactored for readability.

henadzit (approved):

PR #2021, opened 2025-11-23
Fix grouping by in subqueries

### LINKED ISSUE #1999: Selecting a subquery with a GROUP BY clause produces invalid SQL
**Describe the bug**
Starting in version 0.24.0, selecting a subquery containing a GROUP BY clause produces invalid SQL. The field names in the GROUP BY clause are given aliases, which causes a SQL error when the query is executed.

**To Reproduce**
This example script illustrates the problem:

```python
from tortoise import fields, run_async, Tortoise


class A(Model):
    name = fields.TextField()


class B(Model):
    name = fields.TextField()


async def main():

    query = A.annotate(b_name=Subquery(B.all().group_by("name").order_by("name").limit(1).values("name")))
    print(query.sql())

run_async(main())
```

Starting in version 0.24.0, the output of the script contains an invalid field alias `b__name` in the GROUP BY clause of the subquery:

```
SELECT "id","name",(SELECT "name" "name" FROM "b" GROUP BY "name" "b__name" ORDER BY "name" ASC LIMIT ?) "b_name" FROM "a"
```

**Expected behavior**
Versions 0.23.0 and earlier produce a valid GROUP BY clause without field aliases:

```
SELECT "id","name",(SELECT "name" "name" FROM "b" GROUP BY "name" ORDER BY "name" ASC LIMIT ?) "b_name" FROM "a"
```

henadzit:
## Description
- Removes redundant alias from GROUP BY in a Subquery.
- Updates pypika-tortoise to 0.6.3

## Motivation and Context
Fixes https://github.com/tortoise/tortoise-orm/issues/1999.

## How Has This Been Tested?
ci

## Checklist:


Makefile
```diff
@@ -21,25 +21,25 @@ help:
...
 
 deps_with_asyncmy:
-	@uv sync --all-groups --extra asyncpg --extra accel --extra psycopg --extra asyncodbc --extra asyncmy $(options)
+	uv sync --all-groups --extra asyncpg --extra accel --extra psycopg --extra asyncodbc --extra asyncmy $(options)
 
 check: build _check
 _check:
-	ruff format --check $(checkfiles) || (echo "Please run 'make style' to auto-fix style issues" && false)
-	ruff check $(checkfiles)
+	uv run ruff format --check $(checkfiles) || (echo "Please run 'make style' to auto-fix style issues" && false)
```
waketzheng: `uv run` will update the uv.lock
I think it's better to run with `--no-sync` or set [activate-environment](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#inputs) to be true, e.g.: https://github.com/waketzheng/tortoise-orm/commit/a7cad2e35a8112d9c8967f43bc40f66fb62e23cc

henadzit: I don't think it will. The `run` should use the existing lock file.

waketzheng: > I don't think it will. The `run` should use the existing lock file.

Imagine that you are using another mirror instead of pypi.org, for example:
```bash
export UV_DEFAULT_INDEX=https://mirrors.cloud.tencent.com/pypi/simple/
make _check
```

henadzit: You're right! I didn't realize `uv sync` would update the lock file by default. I added `--frozen` which seems to be a good way to keep using the lock.

waketzheng (approved):

PR #2027, opened 2025-12-01
feat: foreignkey to model class

waketzheng:
## Description

- Before this PR:
```
from tortoise import Model, fields

class Group(Model):
    name = fields.CharField(20)

class Role(Model):
    name = fields.CharField(20)

class Member(Model):
    name = fields.CharField(20)
    user = fields.ForeignkeyField('models.User')

class User(Model):
    group = fields.ForeignkeyField('models.Group')
    roles = fields.ManyToManyField('models.Role')
```
- After this PR merged:
```
from tortoise import Model, fields

class Group(Model):
    name = fields.CharField(20)

class Role(Model):
    name = fields.CharField(20)

class Member(Model):
    name = fields.CharField(20)
    user = fields.ForeignkeyField('models.User') # User defined after Member, so use string instead of class

class User(Model):
    group = fields.ForeignkeyField(Group)
    roles = fields.ManyToManyField(Role)
```

## Motivation and Context

To support Django style foreignkey/o2o/m2m

## How Has This Been Tested?

make ci

## Checklist:


tortoise/fields/relational.py
```diff
@@ -306,16 +306,24 @@ def describe(self, serializable: bool) -> dict:
         return desc
 
     @classmethod
-    def validate_model_name(cls, model_name: str) -> None:
-        if len(model_name.split(".")) != 2:
+    def validate_model_name(cls, model_name: str | type[Model]) -> None:
+        if not isinstance(model_name, str):
+            model_class: type[Model] = model_name
+            if not hasattr(model_class, "_meta"):
```
henadzit: You can check for that more precisely: 

```
issubclass(child_type, parent_type)
```

tortoise/fields/relational.py
```diff
@@ -396,9 +404,16 @@ def __init__(
...
         self.validate_model_name(model_name)
-        self.model_name: str = model_name
+        self.model_name = model_name
         self.related_name: str = related_name
-        self.forward_key: str = forward_key or f"{model_name.split('.')[1].lower()}_id"
+        if not forward_key:
+            if not isinstance(model_name, str):
+                forward_key = f"{model_name.__name__.lower()}_id"
+            elif model_name == "self":
+                forward_key = "self_id"  # TODO: compatible
```
henadzit: Did you mean to leave the TODO? It would be nice to expand on what exactly needs to be done or create an issue.

henadzit (commented):
Looks like a great addition!

Does it actually work though? Can you please add tests with "self" and classes passed to relations. Thanks!

tortoise/fields/relational.py
waketzheng: parent_type is tortoise.models.Model, which will cause a import cycle if I use issubclass

tortoise/fields/relational.py
waketzheng: Remove support for 'self' as it requires a symmetrical argument for m2m, which have to change too many codes. I think it is better to do it in another PR.

henadzit (approved):
LGTM! Sorry for not looking at this for so long.

PR #2059, opened 2026-01-20
Fix annotations being selected in ValuesListQuery despite not specified in `.values_list` fields list

### LINKED ISSUE #2058: Annotations are selected in sql despite not being specified in `.values_list` fields list
**Describe the bug**
In tortoise-orm >0.21.7, when `.values_list` is used after `.annotate`, generated sql selects annotations even if they were not specified in values_list (and even if flat=True is passed). In regular queries that use values_list, everything works fine (because, if i understand correctly, ValuesListQuery only selects first field when flat=True, ignoring everything else), but if such query is used inside Subquery - it throws exception `tortoise.exceptions.OperationalError: sub-select returns 2 columns - expected 1`.

**To Reproduce**
```python
from tortoise import fields, run_async
from tortoise.functions import Length

class SomeModel(Model):
    id: int = fields.BigIntField(pk=True)
    name: str = fields.TextField()


@init_memory_sqlite
async def run() -> None:
    await SomeModel.create(name="asdqwe")
    await SomeModel.create(name="asdqweasd")
    await SomeModel.create(name="asdqweasd123")

    subquery = SomeModel.annotate(
        name_length=Length("name"),
    ).filter(
        name_length__gt=6,
    ).values_list(
        "id", flat=True,
    )

    subquery_result = await subquery
    print(f"Subquery result: {subquery_result}")
    print(f"Subquery sql: {subquery.sql()}")

    query = SomeModel.filter(id__in=Subquery(subquery))
    print(f"Query sql: {query.sql()}")

    await query

if __name__ == "__main__":
    run_async(run())
```

If code above is run with tortoise-orm 0.21.7, everything works, subquery sql is `SELECT "id" "0" FROM "somemodel" WHERE LENGTH("name")>6`. But on 0.22.0+, `name_length` is being added to SELECT clause (`SELECT "id" "0",LENGTH("name") "name_length" FROM "somemodel" WHERE LENGTH("name")>6`).

**Expected behavior**
ValuesListQuery selects only specified fields.

**Additional context**
After bisecting, i found out that bug was introduced in [this commit](https://github.com/tortoise/tortoise-orm/commit/9e84736472ceeccfa86a39d74db1636b98a234dc). If i understand correctly - bug is caused by overwriting annotations dict with extra annotations dict in FieldSelectQuery on line 1341.

RuslanUC:
## Description
Fixes #2058 

## How Has This Been Tested?
Added test case in test_queryset.py, ran `make test`.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Fixed
 ^^^^^
 - Fix ``AttributeError`` when using ``tortoise-orm`` with Nuitka-compiled Python code (#2053)
 - Fix 'Self' in python standard library typing.py, but tortoise/model.py required it in 'typing_extensions' (#2051)
+- Fix annotations being selected in ValuesListQuery despite not specified in `.values_list` fields list (#2059)
 
 Changed
 ^^^^^
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -24,7 +24,6 @@ from tortoise.exceptions import (
 from tortoise.expressions import F, RawSQL, Subquery
 from tortoise.functions import Avg, Length
 
-
 # TODO: Test the many exceptions in QuerySet
 # TODO: .filter(intnum_null=None) does not work as expected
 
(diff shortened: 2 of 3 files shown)


waketzheng:
Before this PR:
- Work
```
await MyModel.filter(id__in=Subquery(MyModel.filter().values('id')))
```
- Not work
```
await MyModel.filter(id__in=Subquery(MyModel.filter().values_list('id', flat=True)))
```
After this PR, both of them work.

waketzheng (approved):

PR #2061, opened 2026-01-25
Add native migrations solution

abondar:
## Overview

This PR implements a full migrations solution for Tortoise, providing a native way to manage database state.

It's based on [my old PR](https://github.com/tortoise/tortoise-orm/pull/406), which was never finished and was abandoned.
Building on that PR, it is heavily AI-generated, but the process was curated and tested by me, so it should be safe to use.

@waketzheng @henadzit
I don't expect a meaningful code review of such a large PR, but it would be great if you could test the behavior and provide feedback.

## Changes in main code

- Moved the CLI into the main repository so it can be used for migrations.
- Introduced the `Apps` registry, providing discovery of apps and their models, replacing the old dict-style registry.
- Allowed initializing `Tortoise` without opening connections.
- Added backend-specific changes.
- Introduced optional typing for tortoise config

## Migrations highlights

- Full Django-like offline migrations with state reconstruction.
- Multi-app support with automatic dependency resolution between apps.
- Django-like migrations CLI.
- Data migrations using historical models.

## Implementation overview

- `State`, `ModelState`, and `StateApps` represent model state at a point in time, enabling diff building and state reconstruction.
- `MigrationAutodetector` builds an "old state" from migrations on disk and a "current state" from registered apps, then uses the operation generator to produce schema operations.
- `Migration` is the base class for all migrations created by users.
- `MigrationLoader` resolves and imports migration modules.
- `MigrationGraph` builds the dependency graph across apps and computes execution plans.
- `MigrationRecorder` reads and writes migration history to the database.
- `MigrationExecutor` runs migrations using plans from `MigrationGraph` and tracking progress with `MigrationRecorder`.
- Schema editors provide backend-specific schema operations.


waketzheng:
I like it~
Three years ago, when the first time I used Tortoise ORM, it made me puzzled why the migration commands were `migrate/upgrade` instead of django style `makemigrations/migrate`.

## Suggestion

- Add `tortoise/__main__.py` to support `python -m tortoise`

## Discussion
Sometimes I use tortoise-orm for mini project and expect it to not have too many dependencies.

1. How about move `asyncclick/ptpython` to optional dependencies?
2. Or move this PR to tortoise/tortoise-cli.git?

abondar:
> Add tortoise/__main__.py to support python -m tortoise

Updated

> How about move asyncclick/ptpython to optional dependencies?
> Or move this PR to tortoise/tortoise-cli.git?

I was thinking about these, but I am not sure
Having migrations in separate repo seems like it could bring compatibility problems, as we and user would have to keep them in sync and probably debug issues of people using mismatching versions, which doesn't sound fun

Regarding making them optional - I am in doubt. I get not wanting extra dependencies at base level, but migrations seem like  important part of the package and I am not sure I want people to try to run, realize they don't have cli for some reason and wasting time trying to understand that they need some kind of special packaged tortoise, on top of them already installing tortoise with their db backend definition

abondar:
@waketzheng I removed asyncclick dependency, rewritten it in standard argparse

Bit less colorful, but works fine

Issue #1110, opened 2022-04-21
Problems with pytest supporting in 0.19.0


Firezi:
I want to use pytest via fixture like this https://tortoise-orm.readthedocs.io/en/latest/contrib/unittest.html#py-test, but i get `ConfigurationError`

in `conftest.py`:
```py
@pytest.fixture(scope="function")
def initialize_db(request):
    initializer(["src.models"], db_url="sqlite://:memory:", app_label="models")
    request.addfinalizer(finalizer)
```
simple test:
```py
@pytest.mark.asyncio
async def test_user(initialize_db):
    users = await User.all()
```

But after running test i get:
```
During handling of the above exception, another exception occurred:

initialize_db = None

    @pytest.mark.asyncio
    async def test_user(initialize_db):
>       users = await User.all()

tests/test_models.py:20: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/lib/python3.10/site-packages/tortoise/models.py:1246: in all
    return cls._meta.manager.get_queryset()
venv/lib/python3.10/site-packages/tortoise/manager.py:15: in get_queryset
    return QuerySet(self._model)
venv/lib/python3.10/site-packages/tortoise/queryset.py:297: in __init__
    super().__init__(model)
venv/lib/python3.10/site-packages/tortoise/queryset.py:94: in __init__
    self.capabilities: Capabilities = model._meta.db.capabilities
venv/lib/python3.10/site-packages/tortoise/models.py:286: in db
    return connections.get(self.default_connection)
venv/lib/python3.10/site-packages/tortoise/connection.py:115: in get
    connection: BaseDBAsyncClient = self._create_connection(conn_alias)
venv/lib/python3.10/site-packages/tortoise/connection.py:90: in _create_connection
    db_info = self._get_db_info(conn_alias)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tortoise.connection.ConnectionHandler object at 0x102505a50>
conn_alias = 'models'

    def _get_db_info(self, conn_alias: str) -> Union[str, Dict]:
        try:
            return self.db_config[conn_alias]
        except KeyError:
>           raise ConfigurationError(
[...]

PR #2069, opened 2026-02-02
Introduce TortoiseContext, remove global state

abondar:
Since start of Tortoise - it's state, with connections and apps, was stored globally, as class vars on "Tortoise". 

This brought number of problems, including #1110 , because in tests we were forced to use hacks with global variables to initiate tests, while also limiting users in how they can initialise tortoise

This PR introduces new TortoiseContext, which encapsulates state of tortoise, allowing you to have several of those simultaneously

Also removed old test infrastructure, replacing it with pytest style one. I hope it should ease adding of new tests as now they are less polluting each other

This PR also introduces a lot of breaking changes, so it will be released as part of 1.0

## AI Summary:

This pull request introduces major changes as part of the context-first architecture migration for Tortoise ORM, including breaking changes to test infrastructure, connection handling, and documentation. The update replaces legacy test helpers and fixtures with modern, context-aware pytest fixtures, revises connection access patterns, and updates documentation to reflect the new API and migration guidance.

**Breaking changes and migration to context-first architecture:**

* All ORM state now lives in `TortoiseContext` instances, and legacy test classes/helpers (`test.TestCase`, `initializer()`, etc.) have been removed. `Tortoise.init()` now returns a `TortoiseContext` instead of `None`. Multiple `asyncio.run()` calls in sequence now require explicit context management.
* The `connections` singleton is deprecated; new functions `get_connection()` and `get_connections()` provide connection access from the current context. [[1]](diffhunk://#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43cR15-L24) [[2]](diffhunk://#diff-22610db9239c5a71087e85ed7146da70604c52eba27747a664999729d6797076L7-R211)

**Testing infrastructure overhaul:**

* Legacy test initialization/finalization is replaced by new pytest fixtures (`db`, `db_simple`, `db_isolated`, `db_truncate`) using `tortoise_test_context` for robust, context-based test isolation. A fixture factory (`make_db_fixture`) is provided for custom test modules.
* The test configuration now uses function-scoped fixtures for true isolation and improved performance, with clear documentation for each fixture's use case.

**Documentation and migration guidance:**

* The `connections.rst` documentation is rewritten to describe new connection access patterns, context usage, and migration from the deprecated singleton. It now includes practical examples, API reference, and a migration guide table.
* The changelog is updated to warn about breaking changes and provide a detailed migration summary.

**Minor improvements:**

* Minor formatting and comment changes in documentation and configuration files for clarity.

These changes modernize Tortoise ORM's architecture, improve test reliability and speed, and provide clear migration paths for users upgrading from earlier versions.


uv.lock
waketzheng, line 1: Should run `make up` to upgrade deps

tortoise/cli/cli.py
```diff
@@ -13,8 +13,10 @@
 
 from ptpython.repl import embed
```
waketzheng, line 14: Django shell support ipython(which is much more popular than ptpython) as optional deps and will use the built-in Python REPL if ipython not installed.
Is it possible for us to support both REPL/ipython/ptpython?

abondar: Problem with default REPL is that it doesn't natively support "await ..." 

ipython doesn't have such problem 

Only advantage ptpython has that it is smaller dependency than ipython

Probably we could remove ptpython from dependencies and optionally allow ipython and ptpython, depending on what's available 

Not sure if it worth it to implement fallback to default REPL

waketzheng: 1. No need to fallback REPL
2. Optional ipython/ptpython is OK

abondar: I'll update it in separate PR

PR #2053, opened 2026-01-09
Fix AttributeError when using Nuitka-compiled models

shkey:
# Fix AttributeError when using Nuitka-compiled models

## Description

This PR fixes an `AttributeError` that occurs when using tortoise-orm with Nuitka-compiled Python code. The issue happens during model initialization when the `_get_comments` function attempts to retrieve source code from compiled modules.

**The Problem:**
When code is compiled with Nuitka, the original Python source files are not available at runtime. The `_get_comments` function calls `inspect.getsource(cls)`, which internally calls `inspect.getsourcefile()`. For Nuitka-compiled modules, this returns `None` instead of a file path. Subsequently, when `inspect.findsource()` tries to call `filename.endswith()` on this `None` value, it raises an `AttributeError`.

**The Solution:**
Add `AttributeError` to the exception handling in the `_get_comments` function (line 154 of `tortoise/models.py`), alongside the existing `TypeError` and `OSError` exceptions. This allows the function to gracefully return an empty dictionary when source code cannot be retrieved from compiled modules.

**Code Changes:**

Before:
```python
except (TypeError, OSError):  # pragma: nocoverage
```

After:
```python
except (AttributeError, TypeError, OSError):  # pragma: nocoverage
```

**Error Stack Trace:**
```
Process granian-worker:
Traceback (most recent call last):
  File "/opt/pyenv/versions/3.11.8/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/opt/pyenv/versions/3.11.8/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/var/app/enabled/.venv/lib/python3.11/site-packages/granian/server/mp.py", line 73, in wrapped
    callback = callback_loader()
               ^^^^^^^^^^^^^^^^^
  File "/var/app/enabled/.venv/lib/python3.11/site-packages/granian/_internal.py", line 65, in load_target
    module = load_module(path)
             ^^^^^^^^^^^^^^^^^
  File "/var/app/enabled/.venv/lib/python3.11/site-packages/granian/_internal.py", line 47, in load_module
    __import__(module_name)
......
  File "/var/app/enabled/.venv/lib/python3.11/site-packages/tortoise/models.py", line 516, in __new__
    for fname, comment in _get_comments(new_class).items():  # type: ignore
                          ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #2048, opened 2025-12-22
Examples should not be installed


moubctez:
**Describe the bug**
`examples` directory gets installed in `site-packages`.

**To Reproduce**
Build the wheel.

**Expected behaviour**
Well… `examples` are not installed.

**Additional context**
My patch

```
--- pyproject.toml.orig	2025-12-22 09:00:06.415451379 +0000
+++ pyproject.toml
@@ -101,6 +101,7 @@ path = "tortoise/__init__.py"
 excludes = [
     "./**/.git",
     "./**/.*_cache",
+    "examples",
 ]
 include = [
     "CHANGELOG.rst",
```

PR #2070, opened 2026-02-02
Release 0.25.4

waketzheng:
# 0.25.4

## Fixed

- Fix ``AttributeError`` when using ``tortoise-orm`` with Nuitka-compiled Python code (#2053)
- Fix 'Self' in python standard library typing.py, but tortoise/model.py required it in 'typing_extensions' (#2051)
- Fix examples should not be installed (#2050)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,13 +12,19 @@ Changelog
 
 0.26.0 (unreleased)
 -------------------
+
+0.25
+====
+
+0.25.4
+------
+
 Fixed
 ^^^^^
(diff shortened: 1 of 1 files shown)


abondar:
Do you plan to release it like that? Right now there are migration and tsvector changes in develop 
And I am planning to merge big refactoring in #2069 and was thinking to release it all as 1.0 release

waketzheng:
It seems that 1.0 will be a big change and does not support python3.9
I think we can release a micro version to fix the #2048 which was introduced by version 0.25.3

abondar:
You are probably right, we just need to be careful about our merge strategy 

But probably your current flow with separate branch that we will merge directly to main should work

abondar (approved):

PR #2080, opened 2026-02-08
Fix pydantic_model_creator incorrectly marking fields as Optional

### LINKED ISSUE #1977: Bug: `pydantic_model_creator` incorrectly marks fields as `Optional` when default is not `None`
Hi! I believe I’ve found a small issue in the `pydantic_model_creator` helper function.

**Describe the bug**
**Source**:
https://github.com/tortoise/tortoise-orm/blob/4b4e4a5a016cf327d795e1097d907526aa9d74f7/tortoise/contrib/pydantic/creator.py#L527-L530

In line 528, when generating field annotations, the logic currently considers a field optional if its default is anything other than `None` (`field.default is not None`). However, this can be misleading—for example, an integer field might have a default value of `0`, but `None` is still not a valid input.

In such cases, the generated Pydantic model uses `Optional[int]`, so the schema accepts `None`, but that later causes an `ValueError` at the ORM level bypassing schema validation at the HTTP level.

Let me know if this makes sense, or if there's another reason for this behavior I might have missed.

**Stack versions:**

* pydantic: `2.10.4`
* tortoise-orm: `0.21.3` (also reproduced in the latest version)

**To Reproduce**

```python
class Altitude(models.Model):
    altitude = fields.FloatField(default=0, null=False)

AltitudeSchema = pydantic_model_creator(
    Altitude,
    name="AltitudeSchema",
)
```

```bash
# Generated field annotation
$ AltitudeSchema.__fields__["altitude"]
> FieldInfo(annotation=Union[float, NoneType], required=False, default=0, title='Altitude', json_schema_extra={})
```

Later, when using this schema with a `None` value, it will be accepted by the schema but will fail when inserting into the database:

```python
payload = AltitudeSchema(id=1, altitude=None)
await Altitude.create(**payload.model_dump())
```

```text
  ...
  File "<string>", line 2, in <module>
  File "packages/tortoise/models.py", line 697, in _set_kwargs
    raise ValueError(f"{key} is non nullable field, but null was passed")
ValueError: altitude is non nullable field, but null was passed
```

**Expected behavior**

I expected the Pydantic schema generated by the `pydantic_model_creator` helper function to type as `Optional` only the fields explicitly passed to the `optional` parameter or those declared at the ORM level with `null=True`.

**Additional context**

If I'm not missing anything, I think this issue could be fixed by changing the current `if` block to:

```python
if not field.pk and (
    field_name in self._optional or field.default is None or field.null
):
```

Or even simpler:

```python
if not field.pk and (
    field_name in self._optional or field.null
)
```
The generated schema:

```python
AltitudeSchema.__fields__["altitude"]
# FieldInfo(annotation=float, required=False, default=0, title='Altitude', json_schema_extra={})

AltitudeSchema(id=1, altitude=None)
# Traceback (most recent call last):
#   File "<string>", line 1, in <module>
#  /pydantic/main.py", line 214, in __init__
#     validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
#                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# pydantic_core._pydantic_core.ValidationError: 1 validation error for AltitudeSchema
# altitude
#   Input should be a valid number [type=float_type, input_value=None, input_type=NoneType]
#   For further information visit https://errors.pydantic.dev/2.10/v/float_type
```

veeceey:
## Summary
Fixes #1977 

Fields with default values but `null=False` were incorrectly marked as `Optional` in the Pydantic schema generated by `pydantic_model_creator`. This caused schema validation to accept `None` values that would later fail at the ORM level with a `ValueError`.

## Changes
- Removed the `field.default is not None` condition from the Optional check in `tortoise/contrib/pydantic/creator.py`
- Fields are now only Optional if:
  - Explicitly listed in the `optional` parameter, OR
  - Have `null=True` set on the field
- Updated test expectations to reflect correct behavior
- Added test case demonstrating the fix

## Test Plan
- Added new test `test_pydantic_model_with_default_not_optional` that verifies fields with `default=0` and `null=False` reject `None` values
- All existing pydantic tests pass with updated expectations
- Manually verified that the generated schema no longer incorrectly marks such fields as Optional

## Before
```python
# Field with default=0 and null=False
altitude = fields.FloatField(default=0, null=False)

# Generated schema INCORRECTLY allowed None
Schema.model_fields["altitude"].annotation  # Union[float, NoneType]
Schema(altitude=None)  # Accepted by Pydantic, but fails at ORM level
```

## After
```python
# Same field
altitude = fields.FloatField(default=0, null=False)

# Generated schema correctly rejects None
Schema.model_fields["altitude"].annotation  # float
Schema(altitude=None)  # Rejected by Pydantic validation
```

### DIFF
--- a/tests/contrib/test_pydantic.py
+++ b/tests/contrib/test_pydantic.py
@@ -262,7 +262,7 @@ async def test_event_schema(db, pydantic_setup):
                 "title": "Modified",
                 "type": "string",
             },
-            "token": {"anyOf": [{"type": "string"}, {"type": "null"}], "title": "Token"},
+            "token": {"title": "Token", "type": "string"},
             "alias": {
                 "anyOf": [
                     {"maximum": 2147483647, "minimum": -2147483648, "type": "integer"},
@@ -300,7 +300,7 @@ async def test_eventlist_schema(db, pydantic_setup):
     Event_Pydantic_List = pydantic_setup["Event_Pydantic_List"]
     assert Event_Pydantic_List.model_json_schema() == {
         "$defs": {
-            "Event_bx6qaa": {
+            "Event_rgfzbr": {
                 "additionalProperties": False,
                 "description": "Events on the calendar",
                 "properties": {
@@ -336,8 +336,8 @@ async def test_eventlist_schema(db, pydantic_setup):
                         "type": "string",
                     },
                     "token": {
-                        "anyOf": [{"type": "string"}, {"type": "null"}],
                         "title": "Token",
+                        "type": "string",
                     },
                     "alias": {
                         "anyOf": [
@@ -483,7 +483,7 @@ async def test_eventlist_schema(db, pydantic_setup):
             },
         },
         "description": "Events on the calendar",
-        "items": {"$ref": "#/$defs/Event_bx6qaa"},
+        "items": {"$ref": "#/$defs/Event_rgfzbr"},
         "title": "Event_list",
         "type": "array",
     }
@@ -494,7 +494,7 @@ async def test_address_schema(db, pydantic_setup):
     Address_Pydantic = pydantic_setup["Address_Pydantic"]
     assert Address_Pydantic.model_json_schema() == {
         "$defs": {
-            "Event_lkfyxk_leaf": {
+            "Event_bhjepe_leaf": {
                 "additionalProperties": False,
                 "description": "Events on the calendar",
                 "properties": {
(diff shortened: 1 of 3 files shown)


abondar:
Linters seem to fail

abondar:
Also please consider finding proper place for the test, without referencing issue name in file name, and remove excessive comments on each line in it

abondar:
Merged in https://github.com/tortoise/tortoise-orm/pull/2082

PR #2081, opened 2026-02-08
Fix expand_db_url to handle special characters in passwords

### LINKED ISSUE #404: Tortoise orm uses expand_db_url() throws raise ValueError("Invalid IPv6 URL") for my db_url
**Describe the bug**

tortoise-orm's [expand_db_url](https://github.com/tortoise/tortoise-orm/blob/72f84f0848dc68041157f03e60cd1c92b0ee5137/tortoise/backends/base/config_generator.py#L62) function in [config_generator.py](https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/base/config_generator.py) 
throws  *ValueError("Invalid IPv6 URL")* when trying to parse db_url.

expand_db_url function has,
`    url = urlparse.urlparse(db_url)`

Source code of urlparse (which uses urlsplit, see this line of [code](https://github.com/python/cpython/blob/fa0a66e62d087765dbc5c1b89d6149a23ecfb0a6/Lib/urllib/parse.py#L474))

When given a db url with either one of these special characters '['  or ']', urllib's parse tries to parse and raises it's invalid IPv6 URL error.  

**To Reproduce**
```
from urllib.parse import urlparse

cs_db1 = 'mysql://some_user:ADM[r$VIS]@test-rds.somedata.net:3306/mydb?charset=utf8mb4'
sms_db2 = 'mysql://fail_user:DMK_15[ZWIN6@test-rds.somedata.net:3306/mydb2?charset=utf8mb4'

print("This runs...")
urlparse(cs)
print("Fails...")
urlparse(sms)
```

**Expected behavior**
Expected is to allow this password with special characters. 

https://github.com/tortoise/tortoise-orm/blob/72f84f0848dc68041157f03e60cd1c92b0ee5137/tortoise/backends/base/config_generator.py#L62
https://github.com/tortoise/tortoise-orm/blob/72f84f0848dc68041157f03e60cd1c92b0ee5137/tortoise/backends/base/config_generator.py#L63

This shouldn't be happening. I tried the same with sqlalchemy ORM please have a look at the below code,
mydatabase.py
```
from sqlalchemy import create_engine
from sqlalchemy.orm import scoped_session, sessionmaker
from sqlalchemy.ext.declarative import declarative_base

_db_url = 'mysql://fail_user:DMK_15[ZWIN6@test-rds.somedata.net:3306/mydb2?charset=utf8mb4'
engine = create_engine(_db_url, convert_unicode=True)
db_session = scoped_session(sessionmaker(autocommit=False,
                                         autoflush=False,
                                         bind=engine))
Base = declarative_base()
Base.query = db_session.query_property()

def init_db():
    # import all modules here that might define models so that
    # they will be registered properly on the metadata.  Otherwise
    # you will have to import them first before calling init_db()
    import models
    Base.metadata.create_all(bind=engine)
```
models.py
```
from sqlalchemy import Column, Integer, String
from mydatabase import Base

class User(Base):
    __tablename__ = 'users'
    id = Column(Integer, primary_key=True)
    name = Column(String(50), unique=True)
    email = Column(String(120), unique=True)

    def __init__(self, name=None, email=None):
        self.name = name
        self.email = email

    def __repr__(self):
        return '<User %r>' % (self.name)
``` 
Terminal:
```
>>> from mydatabase import init_db
>>> a = init_db()
```
Above works without any issue. 

After digging through the source code for sqlalchemy they use this [make_url](https://github.com/sqlalchemy/sqlalchemy/blob/dabc226af6d1396dca702a933f64cd7caa3aa126/lib/sqlalchemy/engine/url.py#L221) following [RFC 1738] (https://tools.ietf.org/html/rfc1738). 

Ideally the above password **should be allowed**, it is an unexpected error here. Sqlalchemy connects with db, whereas when using tortoise orm error occurs.
  
I'm working on using and modifying sqlalchemy's db url parsing logic for tortoise orm. Wanted to contrib. Can the dev's have review if I do that change? or is this error expected in tortoise orm and let it be. Please advice. :) 

**Additional context**
Please note, I understand that this is because of the password in db url and just posting this for further discussion. Also somewhere down the line, somebody hopefully stumbles across this discussion when they face this IPV6 error using tortoise orm.

veeceey:
## Summary
Fixes #404

`urlparse.urlparse()` raises `ValueError` when database URLs contain special characters like `[` or `]` in passwords, because it tries to interpret them as IPv6 address brackets.

## Changes
- Added `_quote_url_userinfo()` helper function that percent-encodes the userinfo part (username:password) before URL parsing
- The function preserves already-encoded sequences (like `%25`) to avoid double-encoding
- Only encodes characters that cause parsing issues, leaving properly formatted URLs intact
- Added comprehensive tests for passwords with brackets

## Test Plan
- Added test cases for passwords with balanced brackets: `ADM[r$VIS]`
- Added test for unbalanced brackets: `DMK_15[ZWIN6`
- All existing URL parsing tests pass, including tests for already percent-encoded passwords
- Verified that the fix works with MySQL, PostgreSQL, and other database backends

## Before
```python
db_url = 'mysql://user:DMK_15[ZWIN6@host:3306/db'
expand_db_url(db_url)  # ValueError: Invalid IPv6 URL
```

## After
```python
db_url = 'mysql://user:DMK_15[ZWIN6@host:3306/db'
config = expand_db_url(db_url)
# Returns correctly parsed config with password 'DMK_15[ZWIN6'
```

### DIFF
--- a/tests/test_issue_404.py
+++ b/tests/test_issue_404.py
@@ -1,0 +1,34 @@
+"""Test for issue #404 - expand_db_url fails with special characters in password."""
+import pytest
+from tortoise.backends.base.config_generator import expand_db_url
+
+
+def test_expand_db_url_with_brackets_in_password():
+    """Test that passwords with square brackets work correctly."""
+    # Password with square brackets (as mentioned in issue #404)
+    db_url = 'mysql://some_user:ADM[r$VIS]@test-rds.somedata.net:3306/mydb?charset=utf8mb4'
+
+    config = expand_db_url(db_url)
+
+    assert config["engine"] == "tortoise.backends.mysql"
+    assert config["credentials"]["user"] == "some_user"
+    assert config["credentials"]["password"] == "ADM[r$VIS]"
+    assert config["credentials"]["host"] == "test-rds.somedata.net"
+    assert config["credentials"]["port"] == 3306
+    assert config["credentials"]["database"] == "mydb"
+    assert config["credentials"]["charset"] == "utf8mb4"
+
+
+def test_expand_db_url_with_unbalanced_brackets_in_password():
+    """Test that passwords with unbalanced square brackets work correctly."""
+    # Password with unbalanced bracket (causes IPv6 parsing error)
+    db_url = 'mysql://fail_user:DMK_15[ZWIN6@test-rds.somedata.net:3306/mydb2?charset=utf8mb4'
+
+    config = expand_db_url(db_url)
+
+    assert config["engine"] == "tortoise.backends.mysql"
+    assert config["credentials"]["user"] == "fail_user"
+    assert config["credentials"]["password"] == "DMK_15[ZWIN6"
+    assert config["credentials"]["host"] == "test-rds.somedata.net"
+    assert config["credentials"]["port"] == 3306
+    assert config["credentials"]["database"] == "mydb2"
--- a/tests/test_issue_404.py
+++ b/tests/test_issue_404.py
@@ -1,12 +1,12 @@
 """Test for issue #404 - expand_db_url fails with special characters in password."""
-import pytest
+
 from tortoise.backends.base.config_generator import expand_db_url
 
 
 def test_expand_db_url_with_brackets_in_password():
     """Test that passwords with square brackets work correctly."""
     # Password with square brackets (as mentioned in issue #404)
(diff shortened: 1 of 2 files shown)


abondar:
Please fix "make check"

abondar (approved):

PR #2093, opened 2026-02-10
Improve QuerySet integer-indexing error guidance

sam-dawnops:
## Description
Add a dedicated `ParamsError` message for integer indexing on `QuerySet` and cover it with a regression test.

## Motivation and Context
`QuerySet` is lazy and supports slicing, but `queryset[5]` currently raises a generic message. This change makes that failure mode clearer and tells users how to evaluate/query for a single item correctly.

## How Has This Been Tested?
- `uv run --frozen --group test --group contrib --extra psycopg pytest -q tests/test_queryset.py`
- `uv run --frozen --group dev ruff check tests/test_queryset.py tortoise/queryset.py`

## Checklist:

### DIFF
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -129,6 +129,20 @@ def test_slicing_negative_values(db):
         _ = IntFields.all()[:-1]
 
 
+def test_integer_indexing_not_supported(db):
+    with pytest.raises(
+        ParamsError,
+        match=(
+            r"QuerySet indices must be slices, not integers\. "
+            r"QuerySets are lazy and do not support random access\. "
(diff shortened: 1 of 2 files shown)


waketzheng:
You can add `# type:ignore[unreachable]` at the line of line 480 to fix lint issue

waketzheng:
Thanks for this improvement — the current "QuerySet indices must be slices." message is indeed unhelpful, and the new message gives users a clear path forward. The implementation looks correct: the `isinstance(key, int)` check is properly nested inside the not `isinstance(key, slice)` branch, so non-integer non-slice keys still get the original error.

I've updated the code to fix the lint issues and will merge this once CI passes.

waketzheng (approved):

PR #2116, opened 2026-02-24
Fix sql(params_inline=True) not inlining subquery filter values

### LINKED ISSUE #1800: `sql(params_inline=True)` doesn't include filter values from subqueries
**Describe the bug**
When using `.sql(params_inline=True)` on a query that contains filters with subqueries, the filter values from the subqueries are not included in the generated raw SQL.

### Current Behavior
When calling `.sql(params_inline=True)` on a query with subquery filters, the filter values from the subquery are not included in the final SQL string.

### Expected Behavior
The filter values from subqueries should be included in the generated SQL when using `params_inline=True`.

### Minimal Reproducible Example
```python
from tortoise import Model, fields, Tortoise, run_async


class OtherModel(Model):
    id = fields.IntField(pk=True)
    field = fields.CharField(max_length=255)

    class Meta:
        table = "other_model"


class MyModel(Model):
    id = fields.IntField(pk=True)
    name = fields.CharField(max_length=255)

    class Meta:
        table = "my_model"


async def main():
    config = {
        "apps": {
            "models": {
                "models": ["__main__"],
            }
        },
    }
    await Tortoise.init(config=config)

    # The query
    query = MyModel.filter(id__in=Subquery(OtherModel.filter(field="test_value").values_list("id")))

    # Generate SQL with inline params
    sql = query.sql(params_inline=True)
    print(sql)  # The value "test_value" is not included in the output


if __name__ == "__main__":
    run_async(main())
```

### Additional Context
- Tortoise-ORM version: `0.22.1`

worksbyfriday:
## Summary

Fixes #1800

`Subquery.get_sql()` unconditionally called `get_parameterized_sql()`, which creates a new `Parameterizer` even when the outer context expects inline values. This caused subquery filter values to be rendered as `?` placeholders instead of being inlined.

The fix checks whether the `SqlContext` has an active parameterizer:
- If yes (parameterized mode): use `get_parameterized_sql()` so values are collected into the outer query's parameterizer
- If no (inline mode): use `get_sql()` so values are rendered directly

### Before
```python
query = MyModel.filter(id__in=Subquery(OtherModel.filter(field="test_value").values_list("id")))
query.sql(params_inline=True)
# SELECT ... WHERE "id" IN (SELECT "id" FROM "other_model" WHERE "field"=?)
#                                                                        ^ value missing
```

### After
```python
query.sql(params_inline=True)
# SELECT ... WHERE "id" IN (SELECT "id" FROM "other_model" WHERE "field"='test_value')
#                                                                        ^ value inlined
```

## Changes

- `tortoise/expressions.py`: Check `ctx.parameterizer` before choosing between `get_sql()` and `get_parameterized_sql()`
- `tests/test_query_api.py`: Added regression test verifying subquery values are inlined

## Test plan
- [x] New regression test passes (`test_subquery_params_inline`)
- [x] Existing subquery tests pass (`test_queryset.py -k subquery`)
- [x] Existing `params_inline` tests pass (`test_case_when.py`)
- [x] All query API tests pass (16 passed, 2 skipped)

### DIFF
--- a/tests/test_query_api.py
+++ b/tests/test_query_api.py
@@ -17,6 +17,7 @@ from tortoise.connection import connections
 from tortoise.context import TortoiseContext, tortoise_test_context
 from tortoise.contrib import test
 from tortoise.exceptions import ParamsError
+from tortoise.expressions import Subquery
 from tortoise.models import Model
 from tortoise.query_api import QueryResult, execute_pypika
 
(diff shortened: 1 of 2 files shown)


abondar:
@worksbyfriday please check linter

waketzheng:
For Linux/Mac, run `make lint` to auto fix lint  issue, for windows run `prek run --all-files` to do that.

PR #2118, opened 2026-02-25
fix: ensure AddIndex, AddConstraint, RemoveIndex and RemoveConstraint are in correct order

tsifrer:
## Description

Ensures migration generator now correctly orders `AddIndex` , `AddConstraint`, `RemoveIndex`, `RemoveConstraint` when adding/removing a new indexed field or constraint field to a model.

## Motivation and Context

When adding a new field on an existing model, and using that field in as an index or a constraint, the migration creates operations in the wrong order.
Going from this model:
```python
class OldWidget(models.Model):
    id = fields.IntField(pk=True)
```
to this model:
```python
class OldWidget(models.Model):
    id = fields.IntField(pk=True)
    value = fields.IntField()
    
    class Meta:
        indexes = [Index(fields=["value"])]
```

would generate a migration file with operations in order: `[AddIndex, AddField]` causing migration to fail as it can't add an index to a non existing field.
This PR fixes this order.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,13 @@ Changelog
 1.1
 ===
 
+1.1.6
+-----
+
+Fixed
+^^^^^
+- Migration generator now correctly orders ``AddField`` before ``AddIndex`` or ``AddConstraint`` when adding a new indexed field or constraint field to a model.
+
 1.1.5
 -----
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ Changelog
 
 Fixed
 ^^^^^
-- Migration generator now correctly orders ``AddField`` before ``AddIndex`` or ``AddConstraint`` when adding a new indexed field or constraint field to a model.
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -8,6 +8,13 @@ Changelog
 1.1
 ===
 
+1.1.6
+-----
+
+Fixed
+^^^^^
+- Migration generator now correctly orders ``AddIndex``, ``RemoveIndex``, ``AddConstraint``, ``RemoveConstraint`` operatins when adding/removing a field to a model and is used in an index or constraint.
```
abondar, line 16: It says that it fixes remove operations too - but changes don't seem to do anything about them? 

Is there some implicit guarantee for that? Or maybe we should add "reverse" logic for removing operations there, to guarantee that they are always first?

tsifrer: There is an implicit guarantee for this as 
```python
operations.extend(self._generate_index_operations())
operations.extend(self._generate_constraint_operations())
```
are always called before 
```python
operations.extend(StateFieldDiff(self.old_state, self.new_state).generate_operations())
```

Changelog entry text corresponds to the added tests.

abondar (approved):

PR #2121, opened 2026-03-01
Implement json_contains filter for SQLite backend

### LINKED ISSUE #1961: json_contains NotImplementedError with SQLite - only works with PostgreSQL
**Describe the bug**
Using field__contains on JSONField with SQLite throws NotImplementedError: must be overridden in each executor. Works fine with PostgreSQL but SQLite just crashes.

**To Reproduce**
https://gist.github.com/zkanda/b6151db1bbe984772be037e431a14604

**Expected behavior**
Either implement it for SQLite or give a proper error message saying "JSON contains not supported with SQLite". Right now it's just a confusing NotImplementedError.

**Additional context**

- Error happens in filters.py:230 where json_contains isn't implemented for SQLite
- Makes it impossible to write db-agnostic code with JSON fields
- No docs mention this limitation

Br1an67:
## Description

Implement the `json_contains` filter for the SQLite backend by registering a custom `json_contains` Python function on SQLite connections. The function implements JSON containment semantics matching PostgreSQL's `@>` operator:

- **Arrays**: every element of the candidate appears in the target
- **Objects**: every key-value pair of the candidate exists in the target
- **Scalars**: equality check

## Motivation and Context

Closes #1961.

Using `data__contains` on a `JSONField` with SQLite throws `NotImplementedError: must be overridden in each executor`. This makes it impossible to write database-agnostic code that filters JSON fields. PostgreSQL and MySQL both support this filter; this PR adds SQLite support.

## How Has This Been Tested?

- Updated `test_list_contains` in `tests/fields/test_json.py` to also run on SQLite (was previously restricted to MySQL and PostgreSQL)
- All 15 JSON field tests pass (5 skipped are for contained_by/filter which remain unimplemented for SQLite)
- Ran ruff lint and format checks — all pass

## Checklist:


tortoise/contrib/sqlite/json_functions.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def sqlite_json_contains(field: Term, value: str) -> Criterion:
+    return SQLiteJSONContains(field, ValueWrapper(value))
+
+
+def _json_contains_impl(target_str: str | None, candidate_str: str | None) -> bool:
+    """Check if target JSON value contains the candidate JSON value.
+
+    Semantics match PostgreSQL's @> operator:
```
abondar, line 22: I don't think we should be referencing postgresql implementation in sqlite contrib

PR #2122, opened 2026-03-01
Fix auto_now field not updated with save(update_fields=[...])

### LINKED ISSUE #1512: `auto_now` field does not get updated on model.save(update_fields=[]) if it is not included in update_fields
Hi there,
I am a bit new to tortoise-orm and was playing with things where I encountered a problem where an `auto_now=True` datetimefield does not get updated if I call `model.save(update_fields=['updated_field'])`. It only gets updated if I pass the `auto_now` field in `update_fileds`. I don't know if this was intentional or a miss or just that's how things work.

Br1an67:
## Description

When calling `model.save(update_fields=['some_field'])`, fields with `auto_now=True` were not automatically included in the update, so their timestamps were not refreshed.

## Motivation and Context

Fixes #1512

Django's ORM automatically includes `auto_now` fields when `update_fields` is specified. Tortoise-ORM should behave the same way — `auto_now` fields should always be updated on save regardless of whether they are explicitly listed in `update_fields`.

## How Has This Been Tested?

Added `test_update_auto_now_with_update_fields` in `tests/test_update.py` that:
1. Creates an Event, sets its `modified` timestamp to the past
2. Updates only the `name` field via `save(update_fields=['name'])`
3. Verifies the `modified` (auto_now) field was also updated to today

## Checklist:

### DIFF
--- a/tortoise/models.py
+++ b/tortoise/models.py
@@ -1149,6 +1149,12 @@ class Model(metaclass=ModelMeta):
                 raise IncompleteInstanceError(
                     f"{self.__class__.__name__} is a partial model, can only be saved with the relevant update_field provided"
                 )
+        if update_fields:
+            update_fields = list(update_fields)
+            for field_name, field_obj in self._meta.fields_map.items():
+                if field_name not in update_fields and getattr(field_obj, "auto_now", False):
+                    update_fields.append(field_name)
+
         await self._pre_save(db, update_fields)
 
         if force_create:
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -218,6 +218,26 @@ async def test_bulk_update_auto_now(db):
 
 
 @pytest.mark.asyncio
+async def test_queryset_update_auto_now(db):
+    tournament = await Tournament.create(name="1")
+    event = await Event.create(name="original", tournament=tournament)
+
+    # Set modified to the past (explicit modified= won't be overridden by auto_now)
+    past = timezone.now() - timedelta(days=1)
+    await Event.filter(pk=event.pk).update(modified=past)
+
+    event = await Event.get(pk=event.pk)
+    assert event.modified.date() == past.date()
+
+    # queryset.update() should auto-include auto_now fields
+    await Event.filter(pk=event.pk).update(name="updated")
+
+    event = await Event.get(pk=event.pk)
+    assert event.name == "updated"
+    assert event.modified.date() == timezone.now().date()
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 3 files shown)


tortoise/models.py
```diff
@@ -1149,6 +1149,12 @@ async def save(
                 raise IncompleteInstanceError(
                     f"{self.__class__.__name__} is a partial model, can only be saved with the relevant update_field provided"
                 )
+        if update_fields:
+            update_fields = list(update_fields)
+            for field_name, field_obj in self._meta.fields_map.items():
+                if field_name not in update_fields and getattr(field_obj, "auto_now", False):
+                    update_fields.append(field_name)
```
itzNikshit, line 1156: what about queryset `.update()` and `.bulk_update()` methods?

itzNikshit (changes requested):

Br1an67 (commented):
Good catch! I've added  support to `bulk_update()` as well — the new commit auto-includes `auto_now` fields in `BulkUpdateQuery`, with a corresponding test.

For `QuerySet.update()`, this intentionally does **not** auto-set `auto_now` — it's a raw SQL operation where the caller provides explicit values. This matches Django's documented behavior.

itzNikshit:
> Good catch! I've added support to `bulk_update()` as well — the new commit auto-includes `auto_now` fields in `BulkUpdateQuery`, with a corresponding test.
> 
> For `QuerySet.update()`, this intentionally does **not** auto-set `auto_now` — it's a raw SQL operation where the caller provides explicit values. This matches Django's documented behavior.

Yes, that makes sense. The PR looks good to me. This will now solve the ‘auto_now’ fields issue without explicitly passing them in ‘updated_fields’

Br1an67:
Thanks for the review! Good point about `queryset.update()`.

I've updated the PR to handle all three cases:

1. **`save(update_fields=[...])`** — already handled (injects auto_now fields into update_fields)
2. **`bulk_update()`** — already handled (injects auto_now fields into BulkUpdateQuery)
3. **`queryset.update()`** — **now added**: UpdateQuery automatically injects `timezone.now()` for auto_now fields not explicitly specified in kwargs

Added `test_queryset_update_auto_now` to verify the new behavior.

abondar:
I am not sure right now if we should include this fix in that way, at least regarding querset.update

On one hand that seems logical, but at the same time it is implicit and differs from django behaviour. 
Maybe we should promote more using db_default in case user wants that field to be always updated? Although it hard to imagine cases where user don't want "auto_now" field to not be updated on .update(..) queries

PR #2124, opened 2026-03-01
test: verify LIKE filter in update queries uses parameterized values

### LINKED ISSUE #1225: Fuzzy filter function translation to sql not escaping "%" in `LIKE` clause for `update` operations
**Describe the bug**
Fuzzy filter function translation (like `contains`, `icontains`, `startswith`, `istartswith`, `endswith`, `iendswith`) to raw sql is not escaping `%` in the `LIKE` clause for `update` operations. This leads to error "not enough arguments for format string".

**To Reproduce**
Down below is a demo to reproduce this bug:
```Python
from tortoise.expressions import Q


class Person(Model):
    id = fields.IntField(pk=True)
    id_num = fields.CharField(max_length=18, unique=True, index=True)
    name = fields.CharField(max_length=30, index=True)
    occupation = fields.CharField(max_length=40, index=True)
    alive = fields.BooleanField(index=True, default=True)


async def run():
    await Tortoise.init(db_url="mysql://localhost:3306/", modules={"models": ["__main__"]})

    sql = Person.filter(
        Q(name__startswith="John") &
        Q(id_num__contains="123") &
        Q(occupation__endswith="Doctor")
    ).update(alive=False).sql()
    print(sql)


if __name__ == "__main__":
    run_async(run())
```
The raw sql string is as below:
```'UPDATE `person` SET `alive`=%s WHERE CAST(`name` AS CHAR) LIKE \'John%\' AND CAST(`id_num` AS CHAR) LIKE \'%123%\' AND CAST(`occupation` AS CHAR) LIKE \'%Doctor\''```
The printed output of the above code is as below:
```UPDATE `person` SET `alive`=%s WHERE CAST(`name` AS CHAR) LIKE 'John%' AND CAST(`id_num` AS CHAR) LIKE '%123%' AND CAST(`occupation` AS CHAR) LIKE '%Doctor'```

Tortoise-orm will automatically format the sql with value of `alive` field, but there are other unescaped `%` characters.

**Expected behavior**
The `%` sign in above fuzzy filter translation in `update` operations is escaped correctly.

**Additional context**
None

Br1an67:
## Summary

Adds a regression test for https://github.com/tortoise/tortoise-orm/issues/1225.

The LIKE pattern generated by `contains()` / `startswith()` / `endswith()` filters is passed as a **parameterized query value** (not rendered inline in the SQL template). This means the `%` wildcards in the LIKE pattern never conflict with the MySQL driver's `%s` parameter substitution.

The test verifies the correct parameterized SQL for all four backends:
- **MySQL**: `LIKE %s`
- **PostgreSQL asyncpg**: `LIKE $2 ESCAPE '\\'`
- **PostgreSQL psycopg**: `LIKE %s ESCAPE '\\'`
- **SQLite**: `LIKE ? ESCAPE '\\'`

Fixes #1225

### DIFF
--- a/tests/test_sql.py
+++ b/tests/test_sql.py
@@ -2,7 +2,7 @@ import re
 
 import pytest
 
-from tests.testmodels import CharPkModel, Drink, Event, IntFields
+from tests.testmodels import CharPkModel, Drink, Event, IntFields, Tournament
 from tortoise import connections
 from tortoise.backends.psycopg.client import PsycopgClient
 from tortoise.expressions import F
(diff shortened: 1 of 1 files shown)


abondar:
Test seems to be failing

abondar:
I lost track of this PR's sense - it now just adds test? Logic that it should have fixed actually works properly? Or why issue originally arised?

PR #2126, opened 2026-03-01
Fix bulk_update raising syntax error for ForeignKey fields

### LINKED ISSUE #1335: bulk_update raises syntax error when fields contains a foreign key
**Describe the bug**
Returns the following syntax error when calling bulk_update with foreign key field. 
```

Exception has occurred: OperationalError
near "<": syntax error
sqlite3.OperationalError: near "<": syntax error
```

**To Reproduce**
`

from tortoise import run_async
import tortoise

class Post(tortoise.models.Model):
    id = tortoise.fields.IntField(pk=True)
    title = tortoise.fields.CharField(max_length=255)
    content = tortoise.fields.TextField()

class Comment(tortoise.models.Model):
    id = tortoise.fields.IntField(pk=True)
    post = tortoise.fields.ForeignKeyField('models.Post', related_name='comments')
    content = tortoise.fields.TextField()

run_async(tortoise.Tortoise.init(
    db_url='sqlite://dbtest.sqlite3',
    modules={'models': ['__main__']},
))
run_async(tortoise.Tortoise.generate_schemas())

async def check_bulk_update():
    post1 = Post(title="My first post", content="Hello world!")
    await post1.save()  # will create id in post1
    post2 = Post(title="My first post", content="Hello world!")
    await post2.save()  # will create id in post2
    comment1 = Comment(post=post1, content="Nice post!")
    comment2 = Comment(post=post1, content="I did not like it.")
    await Comment.bulk_create([comment1, comment2])  # works fine, but comment1 and comment2 don't have ids
    c1 = await Comment.get(id=1)  # fetch comment1 properly with id
    c2 = await Comment.get(id=2)  # fetch comment2 properly with id

    # Update both comments with bulk_update
    c1.content = "Nice post! Updated"
    c2.content = "I did not like it. Updated"
    c1.post = post2
    c2.post = post2

    await Comment.bulk_update([c1, c2], fields=['content', 'post'])  # throws syntax error
    await Comment.bulk_update([c1, c2], fields=['content', 'post_id'])  # Works!

run_async(check_bulk_update())
run_async(tortoise.Tortoise.close_connections())
`

**Expected behavior**
Either it should have worked or given a more helpful error.

**Additional context**
Only tested with sqlite. The generated sql looks like this:

`> Comment.bulk_update([c1, c2], fields=['content', 'post']).sql()
UPDATE "comment" SET "content"=CASE WHEN "id"=1 THEN 'Nice post! Updated' WHEN "id"=2 THEN 'I did not like it. Updated' END,"post"=CASE WHEN "id"=1 THEN <Post> WHEN "id"=2 THEN <Post> END WHERE "id" IN (1,2) AND "id" IN (1,2)
`

Br1an67:
## Description

When `bulk_update` was called with a ForeignKey field name (e.g., `fields=['post']`), the generated SQL used the Python object repr (e.g., `<Post>`) instead of the FK column value, causing a syntax error like `near "<": syntax error`.

The fix resolves FK fields to their `source_field` column name and extracts the related object's PK value, matching the behavior of the regular `UpdateQuery`.

## Motivation and Context

Fixes #1335

Using `fields=['post_id']` worked as a workaround, but `fields=['post']` (the FK field name) should also work, as it does for regular `.update()` queries.

## How Has This Been Tested?

Added `test_bulk_update_foreign_key` that:
1. Creates events linked to tournament1
2. Reassigns them to tournament2
3. Calls `bulk_update(events, fields=['tournament'])`
4. Verifies the FK was correctly updated in the database

All 8 existing bulk_update tests continue to pass.

## Checklist:

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -1959,25 +1959,39 @@ class BulkUpdateQuery(UpdateQuery, Generic[MODEL]):
             for field in self.fields:
                 case = Case()
                 pk_list = []
+                field_obj = self.model._meta.fields_map[field]
+                is_fk = isinstance(field_obj, (ForeignKeyFieldInstance, OneToOneFieldInstance))
+                if is_fk:
+                    fk_field = field_obj.source_field
+                    underlying_field_obj = self.model._meta.fields_map[fk_field]
+                    db_column = underlying_field_obj.source_field
+                else:
+                    underlying_field_obj = field_obj
+                    db_column = self.model._meta.fields_db_projection[field]
                 for obj in objects_item:
                     pk_value = self.model._meta.fields_map[pk_attr].to_db_value(obj.pk, None)
-                    field_obj = obj._meta.fields_map[field]
-                    field_value = field_obj.to_db_value(getattr(obj, field), obj)
-                    case.when(
-                        pk == pk_value,
-                        (
-                            Cast(
-                                self.query._wrapper_cls(field_value),
-                                field_obj.get_for_dialect(
-                                    self._db.schema_generator.DIALECT, "SQL_TYPE"
-                                ),
-                            )
-                            if self._db.schema_generator.DIALECT == "postgres"
-                            else self.query._wrapper_cls(field_value)
-                        ),
-                    )
+                    if is_fk:
+                        related_obj = getattr(obj, field)
+                        self.model._validate_relation_type(field, related_obj)
+                        field_value = underlying_field_obj.to_db_value(
+                            getattr(related_obj, field_obj.to_field_instance.model_field_name),
+                            None,
+                        )
+                    else:
+                        field_value = underlying_field_obj.to_db_value(getattr(obj, field), obj)
+                    value_expr: Term
+                    if self._db.schema_generator.DIALECT == "postgres":
+                        value_expr = Cast(
+                            self.query._wrapper_cls(field_value),
+                            underlying_field_obj.get_for_dialect(
+                                self._db.schema_generator.DIALECT, "SQL_TYPE"
+                            ),
+                        )
+                    else:
+                        value_expr = self.query._wrapper_cls(field_value)
+                    case.when(pk == pk_value, value_expr)
                     pk_list.append(pk_value)
-                query = query.set(field, case)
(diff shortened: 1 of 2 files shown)


abondar:
@Br1an67 Also check linter here, pls

tortoise/queryset.py
```diff
@@ -1959,25 +1959,39 @@ def _make_queries(self) -> list[tuple[str, list[Any]]]:
...
                 pk_list = []
+                field_obj = self.model._meta.fields_map[field]
+                is_fk = isinstance(field_obj, (ForeignKeyFieldInstance, OneToOneFieldInstance))
+                if is_fk:
+                    fk_field = field_obj.source_field
+                    underlying_field_obj = self.model._meta.fields_map[fk_field]
+                    db_column = underlying_field_obj.source_field
+                else:
+                    underlying_field_obj = field_obj
+                    db_column = self.model._meta.fields_db_projection[field]
```
waketzheng, line 1970: You changed the non-FK field case from `field` to `self.model._meta.fields_db_projection[field]`. It would be better to add tests for this change.

tortoise/queryset.py
```diff
@@ -1959,25 +1959,39 @@ def _make_queries(self) -> list[tuple[str, list[Any]]]:
...
+                    if is_fk:
+                        related_obj = getattr(obj, field)
+                        self.model._validate_relation_type(field, related_obj)
+                        field_value = underlying_field_obj.to_db_value(
+                            getattr(related_obj, field_obj.to_field_instance.model_field_name),
+                            None,
+                        )
+                    else:
+                        field_value = underlying_field_obj.to_db_value(getattr(obj, field), obj)
+                    value_expr: Term
```
waketzheng, line 1982: Keep the origin code and only change `field_obj` to `underlying_field_obj`, or use the following code to reduce duplicated:
```
                    value_expr: Term = self.query._wrapper_cls(field_value)
                    if self._db.schema_generator.DIALECT == "postgres":
                        value_expr = Cast(
                            value_expr,
                            underlying_field_obj.get_for_dialect(
                                self._db.schema_generator.DIALECT, "SQL_TYPE"
                            ),
                        )
```

PR #2127, opened 2026-03-01
fix: use configured timezone for MySQL session time_zone

### LINKED ISSUE #2114: MySQL session.time_zone is always forced to +00:00 when use_tz=True
**Describe the bug**

When `use_tz=True` is enabled in MySQL connections, the `@@session.time_zone` is incorrectly forced to `+00:00`, ignoring the timezone setting in the configuration. This causes **database-level features like ON UPDATE CURRENT_TIMESTAMP** to function incorrectly, as they generate UTC timestamps instead of the expected local time.
This happens because create_connection in the MySQL client calculates the offset using timezone.now(), which returns a UTC-aware datetime when use_tz=True, resulting in a zero offset.

**To Reproduce**

Initialize Tortoise with `use_tz=True` and timezone="Asia/Shanghai" (or any non-UTC timezone).

Execute a raw SQL query to check the session timezone:

```python
conn = Tortoise.get_connection("default")
res = await conn.execute_query_dict("SELECT @@session.time_zone AS tz")
print(res[0]['tz']) # Output: +00:00 (Expected: +08:00)
```

Root cause in tortoise/backends/mysql/client.py:[
(https://github.com/tortoise/tortoise-orm/blob/314273c0359e06233b4563b3e967c9974f420c00/tortoise/backends/mysql/client.py#L139)](https://github.com/tortoise/tortoise-orm/blob/314273c0359e06233b4563b3e967c9974f420c00/tortoise/backends/mysql/client.py#L112)

```python
# timezone.now() returns UTC when use_tz=True, so utcoffset() is always 0
hours = timezone.now().utcoffset().seconds / 3600 
tz = f"{int(hours):+d}:{int((hours % 1) * 60):02d}" # Results in "+00:00"
await cursor.execute(f"SET time_zone='{tz}';")
```

**Expected behavior**
The MySQL session timezone should reflect the configured timezone (e.g., +08:00), allowing **database-level features like ON UPDATE CURRENT_TIMESTAMP** to function correctly in the local timezone.

**Additional context**
This issue makes it impossible to sync MySQL's internal timestamp generation with the application's local timezone. A possible fix would be to calculate the offset from the actual configured timezone or tortoise.timezone.get_default_timezone() instead of timezone.now().

Br1an67:
## Description

When `use_tz=True`, the MySQL client sets the session `time_zone` using `timezone.now().utcoffset()`. Since `timezone.now()` returns a UTC-aware datetime, its `utcoffset()` is always `timedelta(0)`, causing the session to always be set to `+0:00` regardless of the configured timezone (e.g. `Asia/Shanghai`).

This PR replaces `timezone.now()` with `get_default_timezone()` to compute the correct UTC offset from the user-configured timezone.

## Motivation and Context

Fixes #2114

## How Has This Been Tested?

- Added `test_mysql_session_timezone_uses_configured_tz` in `tests/backends/test_connection_params.py`
- The test configures `USE_TZ=True` and `TIMEZONE=Asia/Shanghai`, then verifies the `SET time_zone` call uses `+8:00`
- Existing tests continue to pass

## Checklist:

### DIFF
--- a/tests/backends/test_connection_params.py
+++ b/tests/backends/test_connection_params.py
@@ -190,9 +190,7 @@ async def test_mysql_session_timezone_uses_configured_tz():
 
                 # Verify SET time_zone was called with Asia/Shanghai offset (+8:00)
                 tz_calls = [
-                    call
-                    for call in mock_cursor.execute.await_args_list
-                    if "time_zone" in str(call)
+                    call for call in mock_cursor.execute.await_args_list if "time_zone" in str(call)
                 ]
                 assert len(tz_calls) == 1, f"Expected 1 SET time_zone call, got {tz_calls}"
                 assert "+8:00" in str(tz_calls[0]), (
--- a/tests/backends/test_connection_params.py
+++ b/tests/backends/test_connection_params.py
@@ -1,8 +1,15 @@
-from unittest.mock import AsyncMock, patch
+import os
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import asyncpg
 import pytest
 
+try:
+    import asyncmy as mysql
+except ImportError:
+    import aiomysql as mysql
+
+from tortoise import timezone
 from tortoise.context import TortoiseContext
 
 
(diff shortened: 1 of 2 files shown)


abondar:
@Br1an67 Hi! Check linter here please

PR #2142, opened 2026-03-13
Migrations fixes for comments and db_default in MySQL

### LINKED ISSUE #2141: Problems with db_default and ineffective migrations when using native Migrations
**Describe the bug (2 bugs)**
1. When db_default="", it becomes ineffective when modifying other attributes of the field (such as max_length).
2. Modifying only the description of a field will generate a migration file, but the migration will not be executed.

**To Reproduce**
Assume the model has the following field:
`name = fields.CharField(max_length=10, default='', db_default='', description='user name')`
The first migration works fine.
Then modify this field (change the field length from 10 to 20):
`name = fields.CharField(max_length=20, default='', db_default='', description='user name')`
After migrating again, the original default value of the name field in the database disappears.

If you modify the description of this field alone, assuming the original description='user name' is now changed to description='user name short', then executing the migration will result in nothing happening.

**Expected behavior**
1. After setting the database-level default value of a field, it should remain unchanged unless it is modified.
2. Modifying the description of a field alone should update it in the database.

abondar:
Fixes #2141

Added comment changing handling for mysql and postgre
Fixed issue for mysql where db_default was lost on alter field

### DIFF
--- a/tests/migrations/test_operations_real_db.py
+++ b/tests/migrations/test_operations_real_db.py
@@ -587,3 +587,206 @@ async def test_alter_field_null_change(db_isolated):
             await conn.execute_script(f"DROP TABLE IF EXISTS {q('test_config', dialect)}")
         except Exception:
             pass
+
+
+# ---------------------------------------------------------------------------
+# Issue #2141 reproduction tests
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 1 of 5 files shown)


seejah:
efficient 💯

PR #2143, opened 2026-03-13
Return value generated by db_default on create

### LINKED ISSUE #2140: Value must not be None?
The table model is as follows:
```
class Devices(Model):
  name = fields.CharField(max_length=20)
  descr = fields.CharField(max_length=50, db_default='')
```
Using the native migrations CLI to generate the MySQL table was successful, and the default value of the descr field is EMPTY_STRING.

Running:
`row = await Devices.create(name='Watch')`

The error message is as follows:
**`tortoise.exceptions.ValidationError: descr: Value must not be None`**

There was no error before version 1.1.6, but after upgrading to version 1.1.6, this error message appears and the creation fails.

abondar:
Fixes #2140

Fetch db_default values for created rows
Turned on by default, works through RETURNING on sqlite and pg, extra query on mysql and odbc
Auto query can be turned off through Meta class of Model

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -286,6 +286,26 @@ The ``Meta`` class
 
             ordering = ["name", "-score"]
 
+    .. attribute:: fetch_db_defaults
+        :annotation: = True
+
+        When ``True`` (the default), after an INSERT on a non-RETURNING backend
+        (e.g. MySQL), Tortoise will issue a follow-up ``SELECT`` to fetch
+        database-applied default values for fields declared with ``db_default``.
+
+        Set to ``False`` to skip the extra query when you don't need the
+        database-generated values back on the Python instance immediately.
+        On RETURNING backends (PostgreSQL, SQLite) the values are always
+        returned in the INSERT response regardless of this setting.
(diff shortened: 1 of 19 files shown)


seejah:
Then, what are the rules for usage? `default` is a Python-level default value, while `db_default` is both a Python-level and a database-level default value?
If both exist and their values are different, which one has higher priority?

Just saw the document. So that's how the rules work:
> Mixing explicit values and database defaults for the same field  raises :exc:`~tortoise.exceptions.OperationalError`.

abondar:
> Mixing explicit values and database defaults for the same field raises :exc:~tortoise.exceptions.OperationalError.

this one is about mixing values in bulk_create, not about `default` and `db_default`
`default` has bigger priority, you will see it on your models if you use both
`db_default` is set in database, so it will be triggered only if nothing was passed as a value from python side.

Normally - it makes sense to use only one, but maybe you can use both if you are okay with values being generated on python side, and add db_default just as failsafe, in case row will be created in db not through orm, so that you can guarantee that value would be there

Issue #2056, opened 2026-01-14
The as_query method in the queryset.py file has been removed


jsonvot:
**Describe the bug**
The as_query() method of the AwaitableQuery class was removed in version 0.24.0, and I haven't found any alternative method to replace it. This method was extremely useful for converting a queryset to a QueryBuilder. I'm wondering if this removal was an intentional/planned change, or an accidental one.

Here are two historical changes to this method:
![Image](https://github.com/user-attachments/assets/bf7b8c90-e922-47aa-9f28-42858bf77543)

![Image](https://github.com/user-attachments/assets/5ff22396-f0c8-495e-8f0d-c2dcf12831f5)

waketzheng:
Could you show code snippet of your use case?

jsonvot:
For example, it is applied in some complex association query scenarios, but you do not want to use physical foreign keys.

```python
from tortoise.models import Model

class User(Model):
	id
	name
	dept_id
	role_id

class Dept(Model):
	id
	name
	
class Role(Model):
	id
	name


tb_user = User._meta.basetable
tb_dept = Dept._meta.basetable
tb_role = Role._meta.basetable

query = (
	User.all().only('email', 'nickname', 'avatar').as_query().
	join(Dept.all().as_query()).on(tb_user.dept_id==tb_dept.id)
	join(Role.all().as_query()).on(tb_user.role_id==tb_role.id)
	select(
		tb_dept.name.as_('dept_name'),
		tb_role.name.as_('role_name')
	)
)
db = User._choose_db()
result = await  db.execute_query_dict(query.get_sql())
```

waketzheng:
@henadzit Cloud you look at this

seladb:
@jsonvot why don't you want to use foreign keys in your models? 🤔

jsonvot:
> [@jsonvot](https://github.com/jsonvot) why don't you want to use foreign keys in your models? 🤔

@seladb I am merely citing a simple usage case. In practice, you may encounter problems that the tortoise model cannot solve directly. In such cases, you have no choice but to convert the tortoise model into a pypika object to implement complex dynamic SQL.

seladb:
> > [@jsonvot](https://github.com/jsonvot) why don't you want to use foreign keys in your models? 🤔
> 
[...]

Issue #1507, opened 2023-10-31
dose tortoise-orm support union query ? 


icecraft:
Hi there, I am new to tortoise-orm. I can use this orm to perform some simple query, but I do not know how to build complex query, such as, union multiple query ! 
   can someone give me some help !

zmievsa:
It supports union queries with `Q` objects -- same as django ORM

DuckTapeMan35:
@zmievsa can you give an example of how this is done? As far as I know django has a dedicated union method on querysets and tortoise doesn't

zmievsa-semasoftware:
Sadly, this was too long ago and I haven't worked with tortoise much since then :)

zmievsa:
Sadly, this was too long ago and I haven't worked with tortoise much since then :)

PR #2146, opened 2026-03-17
feat: add union query support

seladb:
Add union query support in an API similar to [Django's API](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#union)

Addresses these issues:
https://github.com/tortoise/tortoise-orm/issues/1507
https://github.com/tortoise/tortoise-orm/issues/2056 (partially)

## Description

Add union query support:
```python
qs1 = Tournament.filter(name="T1").only("id")
qs2 = Reporter.filter(name="R1").only("id")

result = await qs1.union(qs2)
```

As opposed to Django, this API supports result of multiple models (in Django all results will be instances of the first model).
However, similar to Django, the requested field names and types should be the same across all queried models. This is because of how DBs support union queries.

## Motivation and Context

It addresses the following issues:
https://github.com/tortoise/tortoise-orm/issues/1507
https://github.com/tortoise/tortoise-orm/issues/2056 (partially)

Also - it brings more feature parity with Django.

## How Has This Been Tested?

Added test coverage for all written code. It was tested with MySQL, Postgres and SQLite.

## Checklist:


tortoise/queryset.py
```diff
@@ -2223,3 +2232,211 @@ def sql(self, params_inline=False) -> str:
...
+        query_cls: type,
+    ) -> None:
+        super().__init__(model)
+        self._union_query = union_query
+        self._db = db
+        self._query_cls = query_cls
+
+    def _make_query(self) -> None:
+        ctx = self.query.QUERY_CLS.SQL_CONTEXT
+        self._query_string = f"SELECT COUNT(*) FROM ({self._union_query.get_sql(ctx)})"  # nosec: B608
```
seladb: There must be a better way to do it but I can't figure out how 🤔

tests/test_queryset.py
```diff
@@ -901,3 +901,174 @@ def test_multiple_objects_returned():
...
+
+    if r1.id > t1.id:
+        expected = [t2, t1, r1]
+    else:
+        expected = [t2, r1, t1]
+
+    assert result == expected
+
+
+@requireCapability(dialect=NotEQ("mssql"))
```
seladb, line 998: MSSql limit syntax is different but it's not currently supported in `_SetOperation` which has its own implementation of `_limit_sql()` instead of using the dialect SQL. I guess we can leave it as a TODO for later

https://github.com/tortoise/pypika-tortoise/blob/378f6145f7529d88fa58510851d80454b742406e/pypika_tortoise/queries.py#L672

tortoise/queryset.py
seladb: Nevermind, I managed to figure it out in https://github.com/tortoise/tortoise-orm/pull/2146/commits/e96065b73a7cadae8fd68ccc53b3b78ab3f253f1

tortoise/queryset.py
```diff
@@ -2223,3 +2232,206 @@ def sql(self, params_inline=False) -> str:
...
+        self._db = db
+        self._qs = querysets
+        self._all = all
+        self._orderings: list[tuple[str, Order]] | None = None
+        self._limit: int | None = None
+
+    @classmethod
+    def _get_selects(cls, qs: QuerySet[Model] | UnionQuery[Model]) -> list[str]:
+        return [
+            select.name
```
abondar, line 2313: That will probably fail on query with .annotate(..) - if we don't support them we would need some clean error about that, not attribute error

tortoise/queryset.py
```diff
@@ -2223,3 +2232,206 @@ def sql(self, params_inline=False) -> str:
...
+            return
+
+        if self._orderings:
+            for field_name, order in self._orderings:
+                if field_name not in self._selects:
+                    raise ParamsError("Order by field must be in the select list for union queries")
+
+                self._union_query = self._union_query.orderby(field_name, order=order)
+
+        if self._limit is not None:
```
abondar, line 2352: shouldn't we support offset too in such case? or there are some limitations?

tortoise/queryset.py
```diff
@@ -2223,3 +2232,206 @@ def sql(self, params_inline=False) -> str:
...
+        """
+        Return the union of QuerySets.
+
+        :param other_qs: Another QuerySet(s) to union with.
+        :return: A new UnionQuery representing the union of all QuerySets.
+        """
+        union = self._clone()
+        union._models = {*union._models, *(qs.model for qs in other_qs)}
+        union._qs = union._qs + other_qs
+        union._all = all
```
abondar: that would overwrite existing `_all` from previous union call - is it intended behavior?

tortoise/queryset.py
```diff
@@ -2223,3 +2232,206 @@ def sql(self, params_inline=False) -> str:
...
+
+    def order_by(self, *orderings: str) -> UnionQuery[MODEL]:
+        """
+        Accept args to filter by in format like this:
+
+        .. code-block:: python3
+
+            .order_by('name', '-id')
+
+        Supports ordering by related models too.
```
abondar: does it really support ordering by related models?

tortoise/queryset.py
```diff
@@ -2223,3 +2232,206 @@ def sql(self, params_inline=False) -> str:
...
+            }
+            qs = qs.annotate(**model_annotations)
+            qs._make_query()
+            qs.query.wrap_set_operation_queries = False
+            if not self._union_query:
+                self._union_query = qs.query
+                self._selects = self._get_selects(qs)
+            else:
+                if self._get_selects(qs) != self._selects:
+                    raise ValueError("Union queries must have the same select fields")
```
abondar: Probably should be ParamsError

tortoise/queryset.py
```diff
@@ -2223,3 +2232,206 @@ def sql(self, params_inline=False) -> str:
...
+        union = self._clone()
+        union._limit = limit
+        return union
+
+    def count(self) -> UnionCountQuery:
+        """
+        Return count of objects in union query.
+        """
+        self._choose_db_if_not_chosen()
+        self._make_query()
```
abondar: There is issue here, that count mutates query, and if you will have code like

asyncio.gather(union_query, union_query.count()) - it will apply _make_query two times on same, which could fail

tortoise/queryset.py
seladb: Good catch! Updated in https://github.com/tortoise/tortoise-orm/pull/2146/commits/8f6a10477df22ece4b73dd879f4ddabab1da0914: it'd so `union._all or all`

tortoise/queryset.py
seladb: You're right! Removed in https://github.com/tortoise/tortoise-orm/pull/2146/commits/8f6a10477df22ece4b73dd879f4ddabab1da0914

tortoise/queryset.py
seladb: Fixed in https://github.com/tortoise/tortoise-orm/pull/2146/commits/8f6a10477df22ece4b73dd879f4ddabab1da0914

tortoise/queryset.py
seladb: It's possible to support annotations, but I can probably do it in a separate PR. In the meantime we raise a proper exception here: https://github.com/tortoise/tortoise-orm/pull/2146/commits/709ddfa72a904032b88139062daa50f44d61706e

tortoise/queryset.py
seladb: Added in https://github.com/tortoise/tortoise-orm/pull/2146/commits/dbbed50c8416730239ca14d989bbee3dbd86280b

tests/test_queryset.py
```diff
@@ -901,3 +901,218 @@ def test_multiple_objects_returned():
...
+    await Tournament.create(name="C")
+
+    qs1 = Tournament.all().only("id", "name")
+    qs2 = Reporter.all().only("id", "name")
+
+    result = await qs1.union(qs2).order_by("name").limit(2)
+    assert list(result) == [t1, r1]
+
+
+@requireCapability(dialect=NotEQ("mssql"))
```
seladb, line 1013: ditto: https://github.com/tortoise/tortoise-orm/pull/2146#discussion_r2951938529

tortoise/queryset.py
seladb: Good catch! Fixed in https://github.com/tortoise/tortoise-orm/pull/2146/commits/211bbfe5fd39b26ba45cf7c76c180e69d65c5e64

seladb:
@abondar I addressed your comments, can you please re-review this PR?

seladb:
> @abondar I addressed your comments, can you please re-review this PR?

@abondar a gentle reminder to re-review this PR - I addressed all of your comments

tortoise/backends/base/executor.py
```diff
@@ -154,6 +154,28 @@ async def execute_select(
...
+                # Forcibly yield to the event loop to avoid blocking the event loop
+                # when selecting a large number of rows
+                await asyncio.sleep(0)
+
+            for model in models:
+                if (
+                    model._meta.app == row[app_field]
+                    and model._meta._model.__name__ == row[model_field]
+                ):
+                    instance_list.append(model._init_from_db(**row))
```
abondar, line 179: I think it is not good idea to pass row like that, with extra values of app field and model field. I think we should strip them. I understand that it works now like that, but it seems fragile and can interfere with actual fields declared on model. Maybe even consider making these special field names more unique, to lower chance of overlapping with actual fields (maybt "_*" prefix?)

tortoise/queryset.py
```diff
@@ -2223,3 +2232,226 @@ def sql(self, params_inline=False) -> str:
...
+
+    @classmethod
+    def _get_selects(cls, qs: QuerySet[Model] | UnionQuery[Model]) -> list[str]:
+        return [
+            select.name
+            for select in qs.query._selects
+            if getattr(select, "alias") not in [cls.TORTOISE_APP_FIELD, cls.TORTOISE_MODEL_FIELD]
+        ]
+
+    def _make_query(self) -> None:
```
abondar, line 2318: I feel we still should make it more idempotent, maybe just add self._union_query = None at start?

tortoise/queryset.py
```diff
@@ -2223,3 +2232,226 @@ def sql(self, params_inline=False) -> str:
...
+        return self._execute().__await__()
+
+    async def _execute(self) -> int:
+        _, result = await self._db.execute_query(self.query.get_sql())
+        if not result:
+            return 0
+        return list(dict(result[0]).values())[0]
+
+
+class UnionQuery(AwaitableQuery[MODEL]):
```
abondar: I am skeptical about inheretence here - we inherit AwaitableQuery, but don't use many of it's fields, also redefining `__slots__` and not setting those base fields in `_clone`

Is there reason why we want that inheretence at all? only for choose_db methods? Maybe we should extract choose_db methods to some mixing used by both?

seladb: Do you want me to implement it in this PR or move it to a separate PR?
I'm a bit concerned that this PR already contains quite a lot of changes, maybe we should avoid making even more changes?

abondar: Well, it's still change that concerns new class, it would be strange to ship it with bad inheritance model and start refactoring it right after

tortoise/backends/base/executor.py
seladb: I'm not sure what you mean. We add 2 annotation fields: `tortoise_app` and `tortoise_model` here:
https://github.com/seladb/tortoise-orm/blob/074a61e59e88bbe221aa835f7ba6aa8390e7a605/tortoise/queryset.py#L2316-L2319

I can change the field names to `__tortoise_app__` and `__tortoise_model__` if you think it's better, but there's always a (very small) chance it will interfere with existing fields.

> I think we should strip them

What do you mean by that?

tortoise/queryset.py
seladb: Added in https://github.com/tortoise/tortoise-orm/pull/2146/commits/7ddbc5e9f7f6dadd37a83876f73c6a9a86fa38bd

tortoise/backends/base/executor.py
abondar: You are passing these field values inside model init, I think we shouldn't do it, we should pop them from row dict

seladb: Oh ok, it makes sense. Addressed in https://github.com/tortoise/tortoise-orm/pull/2146/commits/9d541c21cbaea0ebd1ae211872cae2df25eb027d

tortoise/queryset.py
seladb: I made this change in https://github.com/tortoise/tortoise-orm/pull/2146/commits/5f4ad800285c7a3dfae67472c8c6c89877415182, please let me know what you think

seladb:
@abondar I addressed your PR comments, please let me know what you think

seladb:
Thank you @abondar for the thorough review and for merging this PR, much appreciated! 🙏 
I'll update my other PRs and would love if you can review them also

PR #2154, opened 2026-03-23
feat: add postgresql:// scheme as alias for asyncpg

### LINKED ISSUE #2147: Unknown DB scheme: postgresql
Is there a reason tortoise does not accept postgresql for it's DB scheme?
pydantic's PostgresDsn e.g. accepts both, postgres:// and postgresql://

postgresql specifies both: https://www.postgresql.org/docs/current/libpq-connect.html

/tortoise/backends/base/config_generator.py:129 creates the alias for postgres, could we add another alias for postgresql?

r266-tech:
## Summary
Adds `postgresql://` as a supported database URL scheme, aliasing to the asyncpg backend — matching the behavior of `postgres://`.

The [official PostgreSQL documentation](https://www.postgresql.org/docs/current/libpq-connect.html) lists both `postgres://` and `postgresql://` as valid connection URIs, and [pydantic's PostgresDsn](https://github.com/pydantic/pydantic) already accepts both. Previously only `postgres://` worked in Tortoise ORM.

## Changes
- `tortoise/backends/base/config_generator.py`: Added `postgresql` to `urlparse.uses_netloc`, `DB_LOOKUP` aliases, and the password-null check
- `tests/backends/test_db_url.py`: Added `postgresql` to `_postgres_scheme_engines` dict so all existing postgres scheme tests run against it

## Related Issue
Fixes #2147

### DIFF
--- a/tortoise/backends/base/config_generator.py
+++ b/tortoise/backends/base/config_generator.py
@@ -128,6 +128,7 @@ DB_LOOKUP: dict[str, dict[str, Any]] = {
 }
 # Create an alias for backwards compatibility
 DB_LOOKUP["postgres"] = DB_LOOKUP["asyncpg"]
+# "postgresql" is the scheme accepted by libpq and pydantic's PostgresDsn
 DB_LOOKUP["postgresql"] = DB_LOOKUP["asyncpg"]
 
 
--- a/tortoise/backends/base/config_generator.py
+++ b/tortoise/backends/base/config_generator.py
@@ -9,6 +9,7 @@ from typing import Any
 from tortoise.exceptions import ConfigurationError
 
 urlparse.uses_netloc.append("postgres")
+urlparse.uses_netloc.append("postgresql")
 urlparse.uses_netloc.append("asyncpg")
 urlparse.uses_netloc.append("psycopg")
 urlparse.uses_netloc.append("sqlite")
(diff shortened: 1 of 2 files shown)


tortoise/backends/base/config_generator.py
```diff
@@ -127,6 +128,7 @@
 }
 # Create an alias for backwards compatibility
 DB_LOOKUP["postgres"] = DB_LOOKUP["asyncpg"]
+DB_LOOKUP["postgresql"] = DB_LOOKUP["asyncpg"]
```
waketzheng, line 132: Please add a comment line before it:
```
# Create an alias for pydantic compatibility
```

seladb (approved):
LGTM

tortoise/backends/base/config_generator.py
waketzheng: @r266-tech Now that "postgresql" is not for backwards compatibility, it would be better to add a comment explaining why this line is added.

waketzheng (changes requested):

r266-tech:
@waketzheng Done — added a comment above the alias line explaining why it exists (the `postgresql` scheme is accepted by libpq and pydantic's PostgresDsn) in d61f458. Thanks for the review.

waketzheng (approved):

PR #2159, opened 2026-03-29
feat: add pre commit config and fix codespell issues

waketzheng:
## Description

1. Add `.pre-commit-config.yaml` to support  lint code by `prek run --all-files` without creating a virtual environment
2. Fix typos that detected by the codespell hook

## Motivation and Context

Improve contributor experience

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added the changelog accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2162, opened 2026-03-31
feat: expose classmethod to build tortoise config

waketzheng:
## Description

1. The `init` function of tortoise.context.Context is too long
2. Add some new classmethods to TortoiseConfig  which is used for sharing config between contexts

## Motivation and Context

This function has too many lines; shorten it.

## How Has This Been Tested?

make ci

## Checklist:


seladb:
@waketzheng this PR has many changes that are not related to the PR title and description. Maybe you should break them into multiple PRs?

abondar (dismissed):

seladb:
@waketzheng this PR still seems to have unrelated changes. We have [this PR](https://github.com/tortoise/tortoise-orm/pull/2159) to add pre-commit, and [this PR](https://github.com/tortoise/tortoise-orm/pull/2170) for Starlette v1 support. These changes can probably be removed from this PR

waketzheng:
> @waketzheng this PR still seems to have unrelated changes. We have [this PR](https://github.com/tortoise/tortoise-orm/pull/2159) to add pre-commit, and [this PR](https://github.com/tortoise/tortoise-orm/pull/2170) for Starlette v1 support. These changes can probably be removed from this PR

@seladb No need to do that. I expected this PR to be reviewed after the previous one merged, so I explicitly mentioned that it’s `based on #2159`. Actually, I refactored the code once when fixing mypy complaints after Starlette upgraded, and then separated it into three PRs.

seladb:
> > @waketzheng this PR still seems to have unrelated changes. We have [this PR](https://github.com/tortoise/tortoise-orm/pull/2159) to add pre-commit, and [this PR](https://github.com/tortoise/tortoise-orm/pull/2170) for Starlette v1 support. These changes can probably be removed from this PR
> 
> @seladb No need to do that. I expected this PR to be reviewed after the previous one merged, so I explicitly mentioned that it’s `based on #2159`. Actually, I refactored the code once when fixing mypy complaints after Starlette upgraded, and then separated it into three PRs.

@waketzheng is there a reason to open these changes as stacked PRs? It looks like these are separate changes that can each have their own PR based on `develop` 🤔

waketzheng:
1. Using stacked PRs avoids merge conflicts
2. I can `prek run --all-files` to check this PR ([prek](https://github.com/j178/prek) is a reimagined version of pre-commit, built in Rust.)
3. You can select the latest commit (instead of all commits) to compare the code between this PR and the previous one.

image


image

seladb:
> 1. Using stacked PRs avoids merge conflicts
> 

Yes, if the changes in the stacked PRs depend on each other, otherwise there shouldn't be merge conflicts. Here it looks like the [actual commit](https://github.com/tortoise/tortoise-orm/pull/2162/changes/b1e1093fa464522e7b58d8659b41e91a133bde1a) related to this PR doesn't depend on the changes in the other PR. What am I missing here? 🤔

waketzheng:
> Yes, if the changes in the stacked PRs depend on each other, otherwise there shouldn't be merge conflicts. Here it looks like the [actual commit](https://github.com/tortoise/tortoise-orm/pull/2162/changes/b1e1093fa464522e7b58d8659b41e91a133bde1a) related to this PR doesn't depend on the changes in the other PR. What am I missing here? 🤔

You forgot that I need the pre-commit config to enforce code style.

seladb:
> You forgot that I need the pre-commit config to enforce code style.

I'm not sure why pre-commit is needed for this PR, but anyway - the pre-commit PR was merged so this PR is simpler now 🙂

tortoise/config.py
```diff
@@ -202,3 +206,36 @@ def from_dict(cls, data: Mapping[str, Any]) -> TortoiseConfig:
             use_tz=data.get("use_tz"),
             timezone=data.get("timezone"),
         )
+
+    @classmethod
+    def merge_args(
```
seladb, line 219: I'd suggest a version that might be a bit simpler to read:
```python
if config is not None and config_file is not None:
    raise ConfigurationError("Cannot specify both 'config' and 'config_file'")

if config is None and config_file is None:
    if db_url is None or modules is None:
        raise ConfigurationError(
            "Must provide either 'config', 'config_file', or both 'db_url' and 'modules'"
         )
    config_dict = generate_config(db_url, app_modules=modules)
    typed_config = cls.from_dict(config_dict)

# Now we know that at least one of `config` and `config_file` is not `None`

...
```

tortoise/config.py
```diff
@@ -202,3 +206,36 @@ def from_dict(cls, data: Mapping[str, Any]) -> TortoiseConfig:
...
+        cls,
+        config: dict[str, Any] | TortoiseConfig | None = None,
+        config_file: str | None = None,
+        db_url: str | None = None,
+        modules: dict[str, Iterable[str | ModuleType]] | None = None,
+    ) -> TortoiseConfig:
+        # Handle config_file: load it as config dict
+        if config_file is not None:
+            from tortoise import Tortoise
```
seladb: This internal imports are to avoid circular dependencies, right?
I'm wondering if it's not better to just move it as a private/protected method of `TortoiseContext` 🤔

waketzheng: > This internal imports are to avoid circular dependencies, right? I'm wondering if it's not better to just move it as a private/protected method of `TortoiseContext` 🤔

Don't do that:
1. It's only one line (aside from the import and the check that `config is not None`).
2. Typically, we use `config` or `db_url/module`, not `config_file`.

tortoise/config.py
waketzheng: Your suggestion inspired me to make the code more readable：

```
    def merge_args(
        cls,
        config: dict[str, Any] | TortoiseConfig | None = None,
        config_file: str | None = None,
        db_url: str | None = None,
        modules: dict[str, Iterable[str | ModuleType]] | None = None,
    ) -> TortoiseConfig:
        if config is not None:  # Recommended way to configure Tortoise ORM
            ...
        elif db_url and modules:  # For simple cases, use these two arguments
            ...
        elif config_file is not None:  # We also support configuration via a config file
            ...
        else:
            raise ConfigurationError
```

seladb: I think your suggestion make sense, however it will change the existing behavior, for example:
- If both `config` and `config_file` are provided - we currently raise an exception, and this implementation will change it
- If `config_file` is provided together with `db_url` and `modules` - we currently prefer using `config_file`

I'm still not a Tortoise expert, but I'm not sure if we can/want to change the existing behavior...

tortoise/config.py
seladb: I was asking about the internal import:
```python
from tortoise import Tortoise
```

The convention in Python is to put imports at the top of the file. The only exception I know for this rule is if to avoid circular dependencies, which I believe is the case here...
I'm wondering if it's worth having these internal imports to avoid circular dependencies, or we can just move it back to `TortoiseContext` as a private/protected member

abondar: @waketzheng let's move _get_config_from_config_file somewhere else, probably there is no more reason for it to be private method of Tortoise, or maybe even private method of any class, as it doesn't rely on any logic bound to any class. Probably should be public helper function

tortoise/config.py
```diff
@@ -202,3 +208,44 @@ def from_dict(cls, data: Mapping[str, Any]) -> TortoiseConfig:
...
+                "Must provide either 'config', 'config_file', or both 'db_url' and 'modules'"
+            )
+        else:
+            from tortoise.backends.base.config_generator import generate_config
+
+            config_dict = generate_config(db_url, app_modules=modules)
+            return cls.from_dict(config_dict)
+
+    @classmethod
+    def _get_config_from_config_file(cls, config_file: str) -> TortoiseConfig:
```
seladb: Maybe we can make it a public method instead of private/protected?

seladb (approved):
Please see one comment, otherwise LGTM

tortoise/config.py
waketzheng: > Maybe we can make it a public method instead of private/protected?

If it to be a public method, should generating config from `db_url/modules` also be public? For example:
```
230         else:
231             return cls.generate_config(db_url, modules)
232
233     @classmethod
234     def generate_config(
235         cls, db_url: str, modules: dict[str, Iterable[str | ModuleType]]
236     ) -> TortoiseConfig:
237         from tortoise.backends.base.config_generator import generate_config
238
239         config_dict = generate_config(db_url, app_modules=modules)
240         return cls.from_dict(config_dict)
241
242     @classmethod
243     def get_config_from_config_file(cls, config_file: str) -> TortoiseConfig:
```

seladb: I guess we can add it...

tortoise/config.py
```diff
@@ -202,3 +214,48 @@ def from_dict(cls, data: Mapping[str, Any]) -> TortoiseConfig:
...
+        elif db_url is None or modules is None:
+            raise ConfigurationError(
+                "Must provide either 'config', 'config_file', or both 'db_url' and 'modules'"
+            )
+        else:
+            return cls.generate_config(db_url, modules)
+
+    @classmethod
+    def generate_config(cls, db_url: str, modules: dict[str, Iterable[str | ModuleType]]) -> Self:
+        from tortoise.backends.base.config_generator import generate_config
```
seladb, line 241: Is there a reason not to move this import to the top of the file? 🤔

tortoise/config.py
```diff
@@ -202,3 +214,48 @@ def from_dict(cls, data: Mapping[str, Any]) -> TortoiseConfig:
             use_tz=data.get("use_tz"),
             timezone=data.get("timezone"),
         )
+
+    @classmethod
+    def merge_args(
```
seladb, line 219: I'm not sure the method name is clear, `merge_args` seems to create a merged configuration from `config`, `config_file`, `db_url` and `modules`, but this is actually not the case.

Maybe `generate_config` could be a better name.

Moreover, maybe we need to normalize the classmethods in this class:

```python
@classmethod
def generate_config(
    cls,
    config: dict[str, Any] | Self | None = None,
    config_file: str | None = None,
    db_url: str | None = None,
    modules: dict[str, Iterable[str | ModuleType]] | None = None,
) -> Self:
    ...

@classmethod
def generate_config_from_db_url_and_modules(cls, db_url: str, modules: dict[str, Iterable[str | ModuleType]]) -> Self:
    ...

@classmethod
def generate_config_from_config_file(cls, config_file: str) -> Self:
    ...
```

seladb, line 219: Also - since there are public methods, can we add docstrings to each one?

tortoise/cli/cli.py
```diff
@@ -177,14 +177,13 @@ def _load_config(ctx: CLIContext) -> TortoiseConfig:
...
     if not config_value:
         config_value = utils.tortoise_orm_config()
-    if not config_value:
-        raise utils.CLIUsageError(
-            "You must specify TORTOISE_ORM in option or env, or pyproject.toml [tool.tortoise]",
-        )
+        if not config_value:
+            raise utils.CLIUsageError(
+                "You must specify TORTOISE_ORM in option or env, or pyproject.toml [tool.tortoise]",
+            )
```
seladb, line 186: We can do something like this:
```python
config_value = config_value or utils.tortoise_orm_config()
if not config_value:
    raise utils.CLIUsageError(
        "You must specify TORTOISE_ORM in option or env, or pyproject.toml [tool.tortoise]",
    )
return utils.get_tortoise_config(config_value)
```

tortoise/config.py
waketzheng: Make it a lazy import for performance reasons.

tortoise/config.py
waketzheng: Thanks for the suggestions. I’ve updated the naming:

1. To keep consistent with `from_dict`, I’ll use `from_db_url_and_modules` / `from_config_file`.

2. Since there’s already a parameter called `config`, I used `resolve_args` instead of `generate_config` to avoid confusion.

tortoise/cli/cli.py
waketzheng: > ```python
> config_value = config_value or utils.tortoise_orm_config()
> if not config_value:
>     raise utils.CLIUsageError(
>         "You must specify TORTOISE_ORM in option or env, or pyproject.toml [tool.tortoise]",
>     )
> return utils.get_tortoise_config(config_value)
> ```

If config_value is not empty, this will check it twice.

tortoise/config.py
seladb: What is the performance issue we're trying to solve? This file should be called once in the application startup and not be used during runtime. Am I missing something here?

tortoise/config.py
seladb: I like the `from_*` convention and it's consistent with the existing code. Maybe we can rename `resolve_args` to `from_args`? 🤔

tortoise/cli/cli.py
seladb: I'm not sure I understand... can we do this?
```python
def _load_config(ctx: CLIContext) -> TortoiseConfig:
    """Load Tortoise ORM configuration from various sources.

    Returns:
        TortoiseConfig: Validated configuration object
    """ 
    if config_file := ctx.config_file:
        return TortoiseConfig.from_config_file(config_file)

    config_value = ctx.config or utils.tortoise_orm_config()
    if not config_value:
        raise utils.CLIUsageError(
            "You must specify TORTOISE_ORM in option or env, or pyproject.toml [tool.tortoise]",
        )
    return utils.get_tortoise_config(config_value)
```

waketzheng: Nice~

tortoise/config.py
waketzheng: > Maybe we can rename `resolve_args` to `from_args`?

That is one solution. However, I strongly recommend using `resolve_*` instead of `from_*` for this function, because:
1. `resolve_*` checks whether arguments conflict, whereas `from_*` does not.
2. `resolve` indicates that this function will parse the arguments, not just load configuration from them.

tortoise/config.py
waketzheng: Yes, you are right. I initially thought only config.py imported it.

tortoise/config.py
seladb: I think `from_*` is clearer, but we can let @abondar decide...

seladb (approved):
LGTM

tortoise/config.py
```diff
@@ -202,3 +215,102 @@ def from_dict(cls, data: Mapping[str, Any]) -> TortoiseConfig:
...
+        Args:
+            config (dict[str, Any] | TortoiseConfig | None):
+            config_file (str | None): Path to a config YAML or JSON file.
+            db_url (str | None): Database URL for config generation.
+            modules (dict[str, Iterable[str | ModuleType]] | None): App modules for config generation.
+        Args:
+            config: A configuration dict or TortoiseConfig instance.
+            config_file: Path to config file.
+            db_url: Database URL for config generation.
+            modules: App modules for config generation.
```
abondar, line 298: args are duplicated

abondar (approved):
There is duplication of args in docstring
I think resolve_args is okay

We need to add tests regarding these new methods, even if simple ones

Otherwise - looks good to me, @waketzheng you can fix these issues and merge without seeking my explicit reapproval

tortoise/config.py
waketzheng: Fixed by #2184

tortoise/config.py
waketzheng: @seladb Cloud you review #2170

PR #2163, opened 2026-04-01
feat: add `contains()` operation to `QuerySet`

seladb:
Add `QuerySet.contains()` method for checking if object exists in queryset

## Description
Added `QuerySet.contains()` which checks if a specific model instance exists within a queryset. Implementation includes:
- `QuerySet.contains(obj)` method in `tortoise/queryset.py` that returns a `ContainsQuery` object
- `ContainsQuery` class that extends `ExistsQuery` and adds filtering by the object's primary key

## Motivation and Context
This feature enables checking if a specific model instance belongs to a filtered queryset, useful for membership checks without fetching the full object.

It's also in par with [Django API](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#contains).

## How Has This Been Tested?
- Added `test_contains` in `tests/test_queryset.py` covering all cases
- All existing tests pass

## Checklist:


seladb:
@abondar MSSQL tests fail randomly, but shouldn't be related to the changes in this PR

seladb:
@abondar this PR is ready for review, can you please take a look? 🙏

seladb:
> @abondar this PR is ready for review, can you please take a look? 🙏

@abondar @waketzheng a gentle reminder that this PR is ready for review. Do you think you can take a look?

seladb:
@abondar this PR is open for a while, can you please review it?

tortoise/queryset.py
```diff
@@ -822,6 +822,24 @@ def exists(self) -> ExistsQuery:
             use_indexes=self._use_indexes,
         )
 
+    def contains(self, obj: MODEL) -> ContainsQuery:
```
abondar, line 825: It seems like we should add validations around obj here, so that it is same type as queryset, and it's obj.pk is not None - in other case passing such objects can lead to strange behavior

tortoise/queryset.py
```diff
@@ -1473,6 +1491,23 @@ async def _execute(
...
+    ) -> None:
+        super().__init__(**kwargs)
+        self._obj = obj
+
+    def _make_query(self) -> None:
+        super()._make_query()
+        pk_attr = self.model._meta.pk_attr
+        source_pk_attr = self.model._meta.fields_map[pk_attr].source_field or pk_attr
+        pk = Field(source_pk_attr)
+        self.query = self.query.where(pk.eq(self._obj.pk))
```
abondar: `pk_value = self.model._meta.fields_map[pk_attr].to_db_value(self._obj.pk, None)` - we need to add conversion, to be better ready for custom id types

abondar: Also we have `self.model._meta.db_pk_column` - to use it directly instead of calculating

tortoise/queryset.py
seladb: @abondar I'm not sure we have to have to verify the object has the same type as the queryset because we have that in the type-hint, but I can add it if you think it's needed:

image

seladb: Checking if `obj.pk` exists was added in https://github.com/tortoise/tortoise-orm/pull/2163/commits/7ace03cec025c22530a2ca6a658a11e73e834ad5

tortoise/queryset.py
seladb: Updated in https://github.com/tortoise/tortoise-orm/pull/2163/commits/7ace03cec025c22530a2ca6a658a11e73e834ad5

abondar: You didn't changed it to use "to_db_value" - is it intentional?

tortoise/queryset.py
abondar: I am concerned about case where object of other model would be passed and it will silently try to check against that other model id, maybe even returning True, which is definetely not what we want

seladb: Sure, fixed in this commit: https://github.com/tortoise/tortoise-orm/pull/2163/commits/8d17c88e59809aaf954fc3b27dc030f0bbff0fba

tortoise/queryset.py
seladb: I hope I got it right in this commit: https://github.com/tortoise/tortoise-orm/pull/2163/commits/73b1f8a036fbd968cb13588aa5792ddc759c2d5d, please let me know what you think

abondar (approved):

PR #2167, opened 2026-04-04
feat: add domain, URL and email validators

seladb:
Added new validators: `DomainNameValidator`, `URLValidator`, and `EmailValidator` heavily inspired by the corresponding [Django validators](https://docs.djangoproject.com/en/6.0/ref/validators/#django.core.validators)

## Description

Added three new validator classes:
1. `DomainNameValidator` - Validates domain names according to RFC 1034 and RFC 1123
   - Supports internationalized domain names (IDN) via `accept_idna` parameter
   - Includes `InvalidDomainName` exception class
2. `URLValidator` - Validates URLs according to RFC 3986
   - Supports IPv4, IPv6 addresses, ports, and various schemes (http, https, ftp, ftps)
   - Includes `InvalidURL` and `InvalidScheme` exception classes
3. `EmailValidator` - Validates email addresses according to RFC 3696
   - Validates both local part and domain part
   - Supports domain allowlisting via `allowed_domains` parameter
   - Supports literal IPv4/IPv6 addresses in domain
   - Includes `InvalidEmailAddress` exception class

Also added:
- Pre-configured validator instances: `validate_domain_name`, `validate_url`, `validate_email`
- Comprehensive tests

## Motivation and Context

Provides built-in validators for common validation patterns (domain names, URLs, email addresses) that can be used in tortoise-orm field validators.

Also in par with [Django validators](https://docs.djangoproject.com/en/6.0/ref/validators/#django.core.validators).

## How Has This Been Tested?

- Added new test cases unser `pytest tests/test_validators.py` covering valid and invalid inputs for all three validators

## Checklist:


seladb:
@abondar MSSQL tests are failing again 😕 
Can we disable the failing tests for now so we merge PRs that are not related?
Or maybe give me permissions to re-run the failing tests?

seladb:
Also @abondar @long2ice I have 5 PRs ready for review, some of them are ready for quite some time. Do you accept PRs in this project and are available to review them? If yes, I'd really appreciate if you can review.

I really like this project and would love to contribute more! ❤️

seladb:
> @abondar MSSQL tests are failing again 😕 Can we disable the failing tests for now so we merge PRs that are not related? Or maybe give me permissions to re-run the failing tests?

@abondar I think I found the bug! I created a PR with a fix here: https://github.com/tortoise/tortoise-orm/pull/2169

tortoise/validators.py
```diff
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 import abc
+import functools
```
waketzheng: How about `from functools import cached_property`?

tests/test_validators.py
```diff
@@ -116,3 +128,147 @@ async def test_update(db):
...
         await record.save()
+
+
+@pytest.mark.parametrize(
+    "value",
+    [
+        "example.com",
+        "sub.example.com",
+        "example.co.uk",
+        "münchen.de",
```
waketzheng, line 139: Add more case, such as:
```
sub1.sub2.example.org
UPPER-CASE.is.ok.net
tortoise.github.io
example.space
💻.tech
❤️.website
```

tests/test_validators.py
```diff
@@ -116,3 +128,147 @@ async def test_update(db):
...
+def test_domain_name_validator_valid(value):
+    validate_domain_name(value)
+
+
+@pytest.mark.parametrize(
+    "value",
+    [
+        "",
+        "---.com",
+        "example-.com",
```
waketzheng, line 156: Is `under_line.com` valid or not?

tests/test_validators.py
seladb: Apparently `💻.tech` is not a valid domain name. I added the rest to the list of valid domain, and `💻.tech` to the list of non-valid domains in this commit: https://github.com/tortoise/tortoise-orm/pull/2167/commits/92fabb84db24b1230354cc14caf62d63f6ae028a

According the Claude:

> Technically, no — not under current standards, though it's a nuanced answer.
>
> Here's why:
>
> **Punycode / IDN basics:** Non-ASCII characters in domain names are handled via Internationalized Domain Names (IDN), where they get encoded into ASCII-compatible "Punycode." So `💻.tech` would become something like `xn--vi8h.tech` under the hood.
>
> **The emoji problem — IDNA2008**: The current standard governing which Unicode characters are allowed in domain labels is **IDNA2008 (RFC 5891)**. It explicitly **disallows** emoji because they fall into Unicode categories (like `So` — Symbol, Other) that the spec marks as `DISALLOWED`. A domain label must only contain letters, digits, and hyphens from permitted Unicode scripts.
>
> **So why do emoji domains exist at all? A few reasons:**
>
> - The older **IDNA2003** standard was more permissive.
> - Some registries (notably `.ws`) chose to accept emoji domains outside strict IDNA2008 compliance.
> - Browsers like Chrome will sometimes display an emoji domain but resolve it via its Punycode form behind the scenes.
>
> **For `.tech` specifically:** The `.tech` registry follows standard ICANN/IDNA2008 rules, so `💻.tech` would **not** be a registerable or technically valid domain under that TLD.
>
> **Bottom line:** `💻.tech` is not a valid domain name under the IDNA2008 standard. It's a fun idea, but emoji are formally disallowed in domain labels by the current specification.

tests/test_validators.py
seladb: It's not a valid domain. I added it to the list of invalid domains in this commit: https://github.com/tortoise/tortoise-orm/pull/2167/commits/92fabb84db24b1230354cc14caf62d63f6ae028a

tortoise/validators.py
seladb: Sure, updated in this commit: https://github.com/tortoise/tortoise-orm/pull/2167/commits/92fabb84db24b1230354cc14caf62d63f6ae028a

waketzheng (approved):

seladb:
Thank you @waketzheng for reviewing this PR! 🙏 
I don't have permissions to merge it, can you please do it?

PR #2169, opened 2026-04-05
fix: MSSQL connection left in busy state after insert

seladb:
## Description

We recently see a lot of these errors in MSSQL CI which relies on `pyodbc`:
```
[HY000] [Microsoft][ODBC Driver 18 for SQL Server]Connection is busy with results for another command (0) (SQLExecDirectW)
```
`bulk_create` routes through `execute_many`, which uses `cursor.executemany()`. By default, MSSQL sends a "rows affected" message after each statement executed by `executemany`. These messages accumulate as pending result sets on the connection, and since they are never consumed, the connection is left in a busy state when returned to the pool, causing the next operation to fail.

The fix is to override `execute_many` in `MSSQLClient` to replace `cursor.executemany()` with an explicit loop of individual `cursor.execute()` calls, each prefixed with `SET NOCOUNT ON`. This suppresses the row-count messages at the source, preventing them from accumulating on the connection.

**NOTE**: There is potentially a small performance cost from the extra Python overhead, but the inserted data will be identical.

## Motivation and Context

Fix failing CI in MSSQL.

## How Has This Been Tested?

Run MSSQL in CI and make sure the error doesn't show up.

## Checklist:

### DIFF
--- a/tortoise/backends/mssql/client.py
+++ b/tortoise/backends/mssql/client.py
@@ -135,3 +135,11 @@ class TransactionWrapper(ODBCTransactionWrapper, MSSQLClient):
             raise TransactionManagementError("No savepoint to rollback to")
         self._savepoint = None
         self._finalized = True
+
+    @translate_exceptions
+    async def execute_many(self, query: str, values: list) -> None:
+        async with self.acquire_connection() as connection:
+            self.log.debug("%s: %s", query, values)
+            cursor = await connection.cursor()
+            for row_values in values:
+                await cursor.execute(f"SET NOCOUNT ON; {query}", row_values)
--- a/tortoise/backends/mssql/client.py
+++ b/tortoise/backends/mssql/client.py
@@ -65,12 +65,8 @@ class MSSQLClient(ODBCClient):
         async with self.acquire_connection() as connection:
             self.log.debug("%s: %s", query, values)
             async with connection.cursor() as cursor:
-                await cursor.execute(f"SET NOCOUNT ON; {query}; SELECT @@IDENTITY", values)
-                row = await cursor.fetchone()
-                # Drain any remaining result sets so the connection isn't left busy
-                while await cursor.nextset():
-                    pass
-                return row[0]
+                await cursor.executemany(f"SET NOCOUNT ON; {query}; SELECT @@IDENTITY", values)
+                return (await cursor.fetchone())[0]
 
     async def db_delete(self) -> None:
         if not self.database:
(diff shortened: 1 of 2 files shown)


seladb:
@abondar I think this should fix the issue that causes MSSQL CI to fail

abondar:
Did you check that issueing "SET NOCOUNT ON" before execute many doesn't work? Is it limited only to individual inserts? 
Also - we probably need to change "ODBCTransactionWrapper" too, as it overrides execute_many too

seladb:
> Did you check that issueing "SET NOCOUNT ON" before execute many doesn't work? Is it limited only to individual inserts?

Yes, this was one of my attempts that didn't work: https://github.com/tortoise/tortoise-orm/pull/2169/changes/339cebe98f23dcacf63e9363de988c5ad5e733d6

According to Claude: 
```
The `SET NOCOUNT ON` prefix in `executemany` may not be reliable — some ODBC driver configurations don't support multi-statement strings in `executemany` the same way as execute. The safest fix for `MSSQLClient` is to override `execute_many` using a loop of individual execute calls, mirroring how `execute_insert` works
```

> Also - we probably need to change "ODBCTransactionWrapper" too, as it overrides execute_many too

Yes, you're right. ~~I'll address it shortly~~. Addressed in https://github.com/tortoise/tortoise-orm/pull/2169/commits/1a91ae36b78dff4f8fa4360c11f29812678d2f2a


@abondar can you re-review?

abondar:
Can you please try instead of your fix to both client and wrapper - to make such fix to transaction wrapper?

```
  @translate_exceptions                                                                                                                                                            
  async def execute_many(self, query: str, values: list) -> None: 
          self.log.debug("%s: %s", query, values)                                                                                                                                  
          async with connection.cursor() as cursor:  # ← was missing
              await cursor.executemany(query, values)  
```

There is a chance that all problems we are having are just because transactional wrapper doesn't have proper cursor closing, and as tests are running in heavy transaction enviroment - tests are flaking because of it

seladb:
> Can you please try instead of your fix to both client and wrapper - to make such fix to transaction wrapper?
> 
> ```
>   @translate_exceptions                                                                                                                                                            
>   async def execute_many(self, query: str, values: list) -> None: 
>       async with self.acquire_connection() as connection:
>           self.log.debug("%s: %s", query, values)                                                                                                                                  
>           async with connection.cursor() as cursor:  # ← was missing
>               await cursor.executemany(query, values)  
> ```
> 
> There is a chance that all problems we are having are just because transactional wrapper doesn't have proper cursor closing, and as tests are running in heavy transaction enviroment - tests are flaking because of it

@abondar I think you're right! I opened this PR and it seems to be working: https://github.com/tortoise/tortoise-orm/pull/2171
I think we can close this PR and merge the other one?

seladb:
We implemented a better solution in https://github.com/tortoise/tortoise-orm/pull/2171

PR #2171, opened 2026-04-06
fix: MSSQL connection left in busy state after insert

seladb:
Close cursor properly in `ODBCTransactionWrapper.execute_many`.

## Description

We recently see a lot of these errors in MSSQL CI which relies on `pyodbc`:
```
[HY000] [Microsoft][ODBC Driver 18 for SQL Server]Connection is busy with results for another command (0) (SQLExecDirectW)
```

`bulk_create` routes through `execute_many`, which doesn't close the cursor properly in `ODBCTransactionWrapper.execute_many`. This PR tries to fix that.

## Motivation and Context

Fix failing CI in MSSQL

## How Has This Been Tested?

Run MSSQL in CI and make sure the error doesn't show up.

## Checklist:

### DIFF
--- a/tortoise/backends/mssql/client.py
+++ b/tortoise/backends/mssql/client.py
@@ -65,7 +65,7 @@ class MSSQLClient(ODBCClient):
         async with self.acquire_connection() as connection:
             self.log.debug("%s: %s", query, values)
             async with connection.cursor() as cursor:
-                await cursor.execute(f"SET NOCOUNT ON; {query}; SELECT @@IDENTITY", values)
+                await cursor.execute(f"{query}; SELECT @@IDENTITY", values)
                 return (await cursor.fetchone())[0]
 
     async def db_delete(self) -> None:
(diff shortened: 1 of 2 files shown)


seladb:
@abondar this is the fix you suggested in https://github.com/tortoise/tortoise-orm/pull/2169 and it seems to be working. Much better than my fix, thanks! 🙏

abondar:
Can you please remove `SET NOCOUNT ON;`  from https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/mssql/client.py#L68 ? I expect it is not needed anymore if we have fixed the root cause

seladb:
> Can you please remove `SET NOCOUNT ON;` from https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/backends/mssql/client.py#L68 ? I expect it is not needed anymore if we have fixed the root cause

@abondar this change doesn't work: https://github.com/tortoise/tortoise-orm/actions/runs/24065691084/job/70191195132?pr=2171

According to Claude:
```
The root cause is that without SET NOCOUNT ON, SQL Server emits a row-count result set for the INSERT before your SELECT @@IDENTITY result. So when you call cursor.fetchone(), pyodbc is sitting on that row-count result (which isn't a query), not on the @@IDENTITY result — hence "Previous SQL was not a query."
SET NOCOUNT ON suppressed that row-count message, which is why the original code worked.
```

abondar (approved):

PR #2183, opened 2026-04-21
fix: MySQL DATETIME incompatibility in migration recorder

Karmenzind:
## Description

This PR updates the migration recorder to use dialect-specific formatting for applied_at.

For MySQL, the value is now formatted as a MySQL-compatible DATETIME(6) string: `%Y-%m-%d %H:%M:%S.%f`

## Motivation and Context

Currently, record_applied uses `datetime.now(timezone.utc).isoformat()` for all dialects.
For other dialects, the existing timezone-aware ISO 8601 format is preserved (like `2026-04-21T03:44:08.811329+00:00`). This format is accepted by PostgreSQL, but it is not accepted by MySQL DATETIME / DATETIME(6) columns, causing migration recording to fail with an error like:
`Incorrect datetime value: '2026-04-21T03:44:08.811329+00:00' for column 'applied_at'`

## How Has This Been Tested?

Tested locally with MySQL.

Before this change, running migrations failed when Tortoise tried to insert the applied_at value into tortoise_migrations, due to an incompatible datetime format.

After this change, migrations completed successfully and the record was inserted correctly.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,13 @@ Changelog
 1.1
 ===
 
+1.1.8
+-----
+
+Fixed
+^^^^^
+- Fix MySQL DATETIME incompatibility in migration recorder.
+
 1.1.7
 -----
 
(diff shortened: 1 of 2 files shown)


abondar:
Thanks, fixed in https://github.com/tortoise/tortoise-orm/pull/2153 - with more general fix

Karmenzind:
Hi @abondar , do you have a timeline for releasing a version with this fix?
We've been using Tortoise in production for several months, and at the moment we're having to run migrations with a locally patched version. 
Really appreciate any update you can share. Thanks!

PR #2184, opened 2026-05-01
feat: support pathlib.Path as argument of `TortoiseConfig.from_config_file`

waketzheng:
## Description

To support Path as argument of `TortoiseConfig.from_config_file/resolve_args`, for example:
```
from pathlib import Path

filepath = Path('/path/to/config.json')
config = TortoiseConfig.from_config_file(filepath)
config2 = TortoiseConfig.resolve_args(config_file=filepath)

assert config == config2 == TortoiseConfig.from_config_file(str(filepath))
```


tests/test_config.py
```diff
@@ -0,0 +1,167 @@
...
+import pytest
+import yaml
+
+from tortoise.backends.base.config_generator import expand_db_url
+from tortoise.config import TortoiseConfig
+from tortoise.exceptions import ConfigurationError
+
+
+class TestTortoiseConfig:
+    def test_from_invalid_dict(self):
```
seladb: nit: you can use parameterized tests here

tests/test_config.py
```diff
@@ -0,0 +1,167 @@
...
+                "app2": {
+                    "models": ["app2.models"],
+                    "default_connection": "second",
+                },
+            },
+            "routers": ["path.Router"],
+            "use_tz": True,
+            "timezone": "UTC",
+        }
+        assert TortoiseConfig.from_dict(full) is not None
```
seladb: don't we want to check the actual object generate here?

tests/test_config.py
```diff
@@ -0,0 +1,167 @@
...
+        yaml_file_2 = file.with_suffix(".yaml")
+        with yaml_file_2.open("w") as f2:
+            yaml.safe_dump(simple, f2, default_flow_style=False)
+        assert (
+            TortoiseConfig.from_config_file(yaml_file)
+            == TortoiseConfig.from_config_file(str(yaml_file))
+            == TortoiseConfig.from_config_file(yaml_file_2)
+            == TortoiseConfig.from_config_file(file)
+            == TortoiseConfig.resolve_args(config_file=yaml_file)
+        )
```
seladb, line 157: I think we can move yaml to a separate test

tests/test_config.py
```diff
@@ -0,0 +1,167 @@
...
+            == TortoiseConfig.from_dict(simple)
+            == TortoiseConfig.resolve_args(config_file=file)
+        )
+
+        yaml_file = file.with_suffix(".yml")
+        with yaml_file.open("w") as f:
+            yaml.safe_dump(dict(simple), f, default_flow_style=False)
+        yaml_file_2 = file.with_suffix(".yaml")
+        with yaml_file_2.open("w") as f2:
+            yaml.safe_dump(simple, f2, default_flow_style=False)
```
seladb: What's the difference between `yaml_file` and `yaml_file_2`?
The generated yaml is exactly the same:
```shell
>>> yaml.safe_dump(dict(simple))
'apps:\n  app:\n    models:\n    - app.models\nconnections:\n  default: sqlite://db.sqlite3\n'
>>> yaml.safe_dump(simple)
'apps:\n  app:\n    models:\n    - app.models\nconnections:\n  default: sqlite://db.sqlite3\n'
```

tests/test_config.py
```diff
@@ -0,0 +1,167 @@
...
+                    "models": ["app.models"],
+                    "default_connection": "default",
+                }
+            },
+        }
+        db_url = simple["connections"]["default"]
+        modules = {"app": simple["apps"]["app"]["models"]}
+        typed_config = TortoiseConfig.from_db_url_and_modules(db_url, modules)
+        assert typed_config == TortoiseConfig.resolve_args(db_url=db_url, modules=modules)
+        assert typed_config.apps == TortoiseConfig.from_dict(simple).apps
```
seladb: nit: a nicer way to write this test:
```python
def test_from_db_url_and_modules(self):
    db_url = "sqlite://db.sqlite3"
    models = ["app.models"]
    simple = {
        "connections": {"default": db_url},
        "apps": {
            "app": {
                "models": models,
            }
        },
    }
    modules = {"app": models}

    typed_config = TortoiseConfig.from_db_url_and_modules(db_url, modules)
    assert typed_config == TortoiseConfig.resolve_args(db_url=db_url, modules=modules)
    assert typed_config.apps == TortoiseConfig.from_dict(simple).apps
```

tests/test_config.py
```diff
@@ -0,0 +1,167 @@
...
+            TortoiseConfig.resolve_args()
+        with pytest.raises(
+            ConfigurationError,
+            match="Must provide either 'config', 'config_file', or both 'db_url' and 'modules'",
+        ):
+            TortoiseConfig.resolve_args(db_url="")
+        with pytest.raises(
+            ConfigurationError, match="Cannot specify both 'config' and 'config_file'"
+        ):
+            TortoiseConfig.resolve_args(config={}, config_file="a.json")
```
seladb: I'd move it to a separate test and make it parameterized

tests/test_config.py
waketzheng: Updated to use fixture instead.

tests/test_config.py
waketzheng: We need to verify that the JSON file and YAML file produce the same result when dumped from the same dictionary, so put them in a single function.

seladb: Oh ok, I guess it's ok to leave it as is 👍

seladb (approved):
LGTM

PR #2194, opened 2026-05-15
feat: add back `as_query` method to `AwaitableQuery`

seladb:
## Description
Addresses https://github.com/tortoise/tortoise-orm/issues/2056: bring back `AwaitableQuery.as_query()`.

## Motivation and Context
 As described in https://github.com/tortoise/tortoise-orm/issues/2056 there are cases where users have complex queries and need access to the internal pypika layer. The `as_query()` method used to exist but was removed in https://github.com/tortoise/tortoise-orm/pull/1777. I don't see a reason why not to add it back.

## How Has This Been Tested?
Yes, tests added int this PR.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Added
 - ``QuerySet.contains()`` method to check if an object exists in a queryset.
 - Added comprehensive EXPLAIN support for MySQL and PostgreSQL.
 - Built-in ``DomainNameValidator``, ``URLValidator``, and ``EmailValidator`` classes for common validation patterns. (#2162)
+- `AwaitableQuery.as_query()` was added back (#2194)
 
 Fixed
 ^^^^^
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -1140,3 +1140,42 @@ async def test_union_with_annotate_raises(db):
 
     with pytest.raises(ParamsError, match="Union queries do not support annotations"):
         await qs1.union(qs2)
+
+
+@pytest.mark.asyncio
+async def test_as_query_queryset(db):
+    sql = IntFields.all().as_query().get_sql()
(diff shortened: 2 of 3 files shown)


seladb:
@abondar @waketzheng can you please review this PR?

waketzheng:
@seladb Thanks for your work.

The general direction makes sense: #2056 describes a legitimate escape-hatch use case where users need to reuse the SQL/PyPika query already built by Tortoise and then compose something that the ORM API does not currently support well.

One thing worth considering is whether `as_query()` should be added back on `AwaitableQuery` itself.

`AwaitableQuery` is a very broad base class, and not every subclass naturally maps to a single reusable `QueryBuilder`. For example, bulk operations may produce multiple statements, and raw SQL does not really have an internal PyPika query to expose in the same way. Adding the method at the base class level makes it look like this is a stable capability for every awaitable query type, which may create a wider public API contract than intended.

A slightly safer approach might be:

- Add `as_query()` only to query classes that can reliably return a single PyPika `QueryBuilder` (`QuerySet`, `ValuesQuery`, `ValuesListQuery`, `CountQuery`, `ExistsQuery`, and possibly normal `UpdateQuery` / `DeleteQuery`).
- Document it as an advanced escape hatch rather than a primary ORM API.
- Encourage use with `get_parameterized_sql()` / `execute_pypika()` where possible, rather than encouraging users to manually execute `get_sql()`.
- Add a test that covers the actual use case from #2056, e.g. composing a query with `qs.as_query().join(other_qs.as_query()).on(...)`.
- Add tests or explicit behavior for unsupported query types, so users do not assume every awaitable query can be converted into a single PyPika query.

Overall this seems like a useful capability to restore, but it may be worth keeping the public API surface narrower than adding it directly to `AwaitableQuery`.

Issue #2176, opened 2026-04-08
Repeated generation migration


panla:
**Describe the bug**

Repeated generation migration

remove id and then create id

**To Reproduce**

model

```python
class Project(Model):
    id = fields.BigIntField(primary_key=True, source_field='id')

    name = fields.CharField(max_length=64, null=False, description='名称')
    desc = fields.CharField(max_length=1024, null=False, default='', description='描述')

    class Meta:
        app = 'default_db'
        table = 'projects'
        table_description = '项目表'

        indexes = (('name',),)
```

```bash
tortoise makemigrations
```

get 0001_initial.py

```python
from tortoise import migrations
from tortoise.migrations import operations as ops
from orjson import loads
from tortoise.fields.base import OnDelete
from tortoise.fields.data import JSON_DUMPS
from tortoise import fields
from tortoise.indexes import Index

class Migration(migrations.Migration):
    initial = True

    operations = [
        
        ops.CreateModel(
            name='Project',
            fields=[
                ('name', fields.CharField(description='名称', max_length=64)),
                ('desc', fields.CharField(default='', description='描述', max_length=1024)),
            ],
            options={'table': 'projects', 'app': 'default_db', 'indexes': [Index(fields=['name'])], 'pk_attr': 'id', 'table_description': '项目表'},
            bases=['Model'],
        ),
        
    ]

```

```bash
tortoise migrate
```

create tables

no update model and run

```bash
tortoise makemigrations
```
 get 0002_auto_20260408_1716.py

```python
from tortoise import migrations
from tortoise.migrations import operations as ops
from tortoise import fields
from tortoise.indexes import Index

class Migration(migrations.Migration):
    dependencies = [('default_db', '0001_initial')]

    initial = False

    operations = [
        ops.RemoveField(model_name='Project', name='id'),
        ops.AddField(
            model_name='Project',
            name='id',
            field=fields.BigIntField(source_field='id', generated=True, primary_key=True, unique=True, db_index=True),
        ),
        ops.AddIndex(
            model_name='Project',
            index=Index(fields=['id']),
        ),
    ]

```

```text
remove pk, id and create pk, id and index
```

and.

if 

```python
id = fields.BigIntField(primary_key=True)
```

gen normal

```python
ops.CreateModel(
            name='Project',
            fields=[
                ('id', fields.BigIntField(generated=True, primary_key=True, unique=True, db_index=True)),
                ('name', fields.CharField(description='名称', max_length=64)),
[...]

PR #2195, opened 2026-05-15
fix(migrations): migration serialization bugs for PK fields

seladb:
## Description
Fixes 2 bugs that were discovered in https://github.com/tortoise/tortoise-orm/issues/2176

## Motivation and Context

Two bugs in migration serialization when a PK field is configured with
non-default attributes:
1. **`generated=False` on PK fields was lost during serialization** —
   `Field.deconstruct()` only emitted `generated` when the value was
   truthy. If a PK field explicitly set `generated=False`, the kwarg was
   omitted, causing the field to revert to its per-engine default
   (`True` for IntField) on deserialization. The fix includes the
   `generated` kwarg whenever `self.pk` is true, preserving the explicit
   value.
2. **PK fields with `source_field="<FIELD_NAME>"` were dropped from CreateModel** —
   `MigrationWriter._format_create_model()` collected all `source_field`
   values and skipped any field whose `name` appeared in that set. When a
   PK field had `source_field="<FIELD_NAME>"`, its own name matched the set entry
   and it was incorrectly excluded from the migration output. The fix
   exempts PK fields from this deduplication logic.

## How Has This Been Tested?
- Added tests for deconstructing with different options of `generated` in PK and non-PK fields
- Added test for a migration with `source_field` that is similar to the field name

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Added
 Fixed
 ^^^^^
 - ``MigrationRecorder`` now uses parameterized queries; fixes MariaDB/MySQL rejecting ISO-8601 ``applied_at`` values. (#2132)
+- Fix migration ``deconstruct()`` and ``CreateModel`` serialization for PK fields with ``generated=False`` or ``source_field``. (#2195)
 
 1.1.7
 -----
--- a/tests/fields/test_db_default.py
+++ b/tests/fields/test_db_default.py
@@ -213,6 +213,29 @@ def test_deconstruct_with_both_default_and_db_default():
     assert kwargs["db_default"] == 2
 
 
+def test_deconstruct_pk_generated_false():
+    f = fields.IntField(primary_key=True, generated=False)
+    f.model_field_name = "test_field"
+    path, args, kwargs = f.deconstruct()
+    assert kwargs.get("generated") is False
+    assert kwargs.get("primary_key") is True
+
+
+def test_deconstruct_pk_generated_default():
+    f = fields.IntField(primary_key=True)
+    f.model_field_name = "test_field"
+    path, args, kwargs = f.deconstruct()
+    assert kwargs.get("generated") is True
+    assert kwargs.get("primary_key") is True
+
+
+def test_deconstruct_non_pk_generated_false():
+    f = fields.IntField(generated=False)
+    f.model_field_name = "test_field"
+    path, args, kwargs = f.deconstruct()
+    assert "generated" not in kwargs
+
+
 # ============================================================================
 # Sentinel behavior
 # ============================================================================
--- a/tests/migrations/test_writer.py
+++ b/tests/migrations/test_writer.py
@@ -292,6 +292,37 @@ def test_writer_excludes_fk_source_field(tmp_path: Path, monkeypatch) -> None:
     _write_migration(tmp_path, monkeypatch, "0005_fk_source", operations, expected)
 
 
+def test_writer_includes_pk_with_source_field(tmp_path: Path, monkeypatch) -> None:
+    operations = [
+        CreateModel(
+            name="Widget",
+            fields=[
+                ("id", fields.IntField(primary_key=True, source_field="id")),
+                ("name", fields.CharField(max_length=100)),
+            ],
+        )
+    ]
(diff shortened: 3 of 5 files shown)


seladb:
@waketzheng @abondar I'm not sure why the coverage workflow didn't start, can you cancel and restart it (I don't have permissions to do so)?

seladb:
@abondar @waketzheng can you please review this PR?

waketzheng:
@seladb Thanks for tracking this down. The fixes make sense for the two cases discussed in #2176: preserving `generated=False` on primary-key fields, and avoiding the missing field in `CreateModel` when a PK has `source_field` equal to its own field name.

One thing worth considering is whether the `CreateModel` fix should be expressed in terms of “source fields claimed by another field” rather than special-casing PK fields.

The current change fixes the reported PK case:

```python
if name in source_fields and not getattr(field, "pk", False):
    continue
```

But the underlying issue seems a bit broader: a field should only be skipped when its name is the backing DB column for another field, such as the generated `author_id` field behind a `ForeignKeyField`. If a field has `source_field` equal to its own name, that is redundant but it probably should not cause the field to be dropped, regardless of whether it is a PK.

A slightly more general approach might be:

```python
source_fields = {
    field.source_field
    for name, field in operation.fields
    if (
        field is not None
        and getattr(field, "source_field", None)
        and field.source_field != name
    )
}
```

Then the existing `if name in source_fields: continue` logic would still skip FK backing fields, while keeping both PK and non-PK fields whose `source_field` is simply the same as the model field name.

It may also be worth adding one regression test for the non-PK version of the same pattern, for example:

```python
("name", fields.CharField(max_length=100, source_field="name"))
```

The `generated=False` deconstruction tests look useful. Since #2176 is about repeated migration generation, a higher-level regression test that exercises the writer/autodetector path for `primary_key=True, generated=False` could make the fix a bit more future-proof too.

Overall this looks like the right direction; the main question is whether the `source_field` filtering should be generalized so the same bug does not remain for non-PK fields.

seladb:
> @seladb Thanks for tracking this down. The fixes make sense for the two cases discussed in #2176: preserving `generated=False` on primary-key fields, and avoiding the missing field in `CreateModel` when a PK has `source_field` equal to its own field name.
> 
> One thing worth considering is whether the `CreateModel` fix should be expressed in terms of “source fields claimed by another field” rather than special-casing PK fields.
> 
> The current change fixes the reported PK case:
> 
> ```python
> if name in source_fields and not getattr(field, "pk", False):
>     continue
> ```
> 
> But the underlying issue seems a bit broader: a field should only be skipped when its name is the backing DB column for another field, such as the generated `author_id` field behind a `ForeignKeyField`. If a field has `source_field` equal to its own name, that is redundant but it probably should not cause the field to be dropped, regardless of whether it is a PK.
> 
> A slightly more general approach might be:
> 
> ```python
> source_fields = {
>     field.source_field
>     for name, field in operation.fields
>     if (
>         field is not None
>         and getattr(field, "source_field", None)
>         and field.source_field != name
>     )
> }
> ```
> 
> Then the existing `if name in source_fields: continue` logic would still skip FK backing fields, while keeping both PK and non-PK fields whose `source_field` is simply the same as the model field name.
> 
> It may also be worth adding one regression test for the non-PK version of the same pattern, for example:
> 
> ```python
> ("name", fields.CharField(max_length=100, source_field="name"))
> ```
> 
> The `generated=False` deconstruction tests look useful. Since #2176 is about repeated migration generation, a higher-level regression test that exercises the writer/autodetector path for `primary_key=True, generated=False` could make the fix a bit more future-proof too.
> 
> Overall this looks like the right direction; the main question is whether the `source_field` filtering should be generalized so the same bug does not remain for non-PK fields.

@waketzheng I agree with your suggestion, however when I implemented it [`test_writer_excludes_fk_source_field`](https://github.com/seladb/tortoise-orm/blob/1e5401a7d1272bf7202ece7079826cff8998b206/tests/migrations/test_writer.py#L262) fails. I think that the test is wrong, if `source_field="author_id"` checking if the field name equals to the `source_name` is not the correct check, it should check if the `source_name` equals to one of the other fields generated by the migration (which I don't know how and if we can check 🤷 ). Should I remove this test?

PR #2200, opened 2026-05-20
Fix `db_default` on FK/O2O dropped during `_init_relations`

### LINKED ISSUE #2199: FK/O2O `db_default` is dropped during `_init_relations` (CREATE TABLE omits DEFAULT on FK columns)
**Describe the bug**

When `db_default` is set on a `ForeignKeyField` / `OneToOneField`, the value is silently dropped at apply time and never reaches the generated DDL. The `CREATE TABLE` for the owning model is emitted without a `DEFAULT` clause on the FK column, even though the migration file faithfully records `db_default=...` on the relation.

Note: #2117 / PR #2129 fixed this for regular fields (the schema editor now emits `DEFAULT` for non-relation columns). This issue is the FK twin of that bug, in a different code path.

**Root cause**

`Apps._init_relations` (`tortoise/apps.py`) builds the implicit key field (`<fk>_id`) by copying the related model's PK field and then patching a small set of attributes from the FK object:

```python
key_fk_object = copy(related_field)
key_fk_object.reference = fk_object
key_fk_object.source_field = fk_object.source_field or key_field
for attr in ("index", "default", "null", "generated", "description"):
    setattr(key_fk_object, attr, getattr(fk_object, attr))
```

`db_default` is **not** in this list. Since the schema editor iterates `fields_db_projection` (which contains the key field, not the relation field) and inspects `key_field.has_db_default()`, the DEFAULT clause is never produced for FK columns.

**To Reproduce**

```python
from tortoise import fields, Model

class Dc(Model):
    id = fields.IntField(pk=True)

class App(Model):
    id = fields.IntField(pk=True)
    dc = fields.ForeignKeyField("models.Dc", db_default=2)
```

`makemigrations` produces an op that records `db_default=2` on the FK:

```python
('dc', fields.ForeignKeyField('models.Dc', source_field='dc_id',
                              db_default=2, related_name='apps',
                              on_delete=OnDelete.CASCADE)),
```

But `tortoise upgrade` emits:

```sql
CREATE TABLE "app" (
    "id" SERIAL NOT NULL PRIMARY KEY,
    "dc_id" INT NOT NULL REFERENCES "dc" ("id") ON DELETE CASCADE
);
```

— no `DEFAULT 2`.

**Expected behavior**

```sql
CREATE TABLE "app" (
    "id" SERIAL NOT NULL PRIMARY KEY,
    "dc_id" INT NOT NULL DEFAULT 2 REFERENCES "dc" ("id") ON DELETE CASCADE
);
```

i.e. `db_default` on FK/O2O propagates to the underlying column just like it does for plain fields after #2129.

**Additional context**

- Confirmed present in `1.1.7` (release) and `develop` (HEAD as of today).
- Fix is a one-line change in `_init_relations`: add `"db_default"` to the attribute-copy tuple. Tests can mirror `test_create_model_includes_db_default` from PR #2129, but with a `ForeignKeyField(..., db_default=<id>)` instead of a plain field.
- Same path covers `OneToOneField` since it shares the `init_fk_o2o_field` codepath.

mixartemev:
## Description

`_init_relations` builds the implicit `<fk>_id` key field by copying the related model's PK field and then patching a small set of attributes from the FK object:

```python
for attr in ("index", "default", "null", "generated", "description"):
    setattr(key_fk_object, attr, getattr(fk_object, attr))
```

`db_default` is missing from this list. The schema editor then iterates `fields_db_projection` (which holds the key field, not the FK relation field) and inspects `key_field.has_db_default()`. It returns `False`, so `CREATE TABLE` for the owning model is emitted without a `DEFAULT` clause on the FK column — even when `db_default=...` is set on the `ForeignKeyField` / `OneToOneField` and faithfully recorded in the generated migration.

This fix adds `"db_default"` to the attribute-copy tuple. Same path covers `OneToOneField` since both go through `init_fk_o2o_field`.

## Motivation and Context

Fixes #2199

PR #2129 fixed the same omission for non-relation fields (the schema editor now emits `DEFAULT` for plain columns). This is the FK twin of that bug, living one layer up in `Apps._init_relations` — the schema editor never gets a chance to see the default because it isn't on the key field.

## How Has This Been Tested?

Added `test_create_model_includes_db_default_on_fk` in `tests/migrations/test_schema_editor_sql.py`:

1. Declares a target model `Dc` and a model `App` with `dc = ForeignKeyField("models.Dc", db_default=2)`.
2. Runs `init_apps(Dc, App)` (which exercises `StateApps._init_relations` — the fixed path).
3. Calls `create_model(App)` via the schema editor.
4. Asserts the SQL contains `"dc_id"` and `DEFAULT 2`.

The test fails on `develop` HEAD without the fix (verified locally), and passes with it. All 298 existing migration tests continue to pass.

## Checklist:


tests/migrations/test_schema_editor_sql.py
```diff
@@ -249,3 +249,36 @@ class Meta:
...
     assert 'CREATE TABLE "widget"' in sql
     assert "DEFAULT 'active'" in sql
+
+
+@pytest.mark.asyncio
+async def test_create_model_includes_db_default_on_fk() -> None:
+    """CreateModel should include DEFAULT clause for FK columns with db_default."""
+
+    class Dc(Model):
+        id = fields.IntField(pk=True)
```
waketzheng: Use `id = fields.IntField(primary_key=True)` instead

tests/migrations/test_schema_editor_sql.py
```diff
@@ -249,3 +249,36 @@ class Meta:
...
+
+        class Meta:
+            table = "dc"
+            app = "models"
+
+    class App(Model):
+        id = fields.IntField(pk=True)
```
waketzheng: pk -> primary_key

waketzheng (changes requested):

tests/migrations/test_schema_editor_sql.py
mixartemev: ok bro, I fixed that, but in all other tests in whole this file, using `pk` notation also

waketzheng (dismissed):

waketzheng:
@mixartemev, could you fix the conflicts?

waketzheng (approved):

PR #2203, opened 2026-05-26
fix(migrations): use `primary_key` instead of deprecated `pk` in `MigrationRecorder`

mykolasolodukha:
## Description
Build the `MigrationRecorder` bookkeeping model with `primary_key=True` instead of the deprecated `pk=True`, so applying migrations no longer emits tortoise's own `pk` `DeprecationWarning`.

## Motivation and Context
`tortoise/fields/base.py` deprecated `pk=`/`index=` in favor of `primary_key=`/`db_index=`. But `MigrationRecorder._make_model` still used `IntField(pk=True)`, so any project applying migrations tripped a `DeprecationWarning` raised from *inside* tortoise — with no way to silence it from their own code. This is a pure kwarg rename (identical behavior, no schema change). A codebase-wide sweep confirmed this was the only internal use of a deprecated field kwarg.

## How Has This Been Tested?
Added a regression test in `tests/migrations/test_recorder.py` that instantiates `MigrationRecorder` with warning filters reset (to bypass the repo-wide `filterwarnings` ignore) and asserts no `pk`/`index` `DeprecationWarning` is emitted. Confirmed it fails before the rename and passes after. Full migrations suite green; `ruff` + `mypy` clean.

## Checklist:


waketzheng (dismissed):

CHANGELOG.rst
```diff
@@ -21,6 +21,7 @@ Added
 Fixed
 ^^^^^
 - ``MigrationRecorder`` now uses parameterized queries; fixes MariaDB/MySQL rejecting ISO-8601 ``applied_at`` values. (#2132)
+- ``MigrationRecorder`` no longer emits tortoise's own ``pk`` field ``DeprecationWarning`` when applying migrations; it now builds its bookkeeping model with ``primary_key=True``.
```
waketzheng: @mykolasolodukha Cloud you append `(#2203)` to the end of this line?

waketzheng (approved):

PR #2205, opened 2026-05-27
feat: rewrote the method for getting m2m objects

VladislavYar:
## Description

The `_prefetch_m2m_relation` method of the `BaseExecutor` class has been rewritten. Now objects are received directly through the specified `queryset`, which allows you to correctly use `select_related`, `annotate`, etc.
## Motivation and Context

`prefetch_related` (`m2m`) functionality becomes like in `Django`.
## How Has This Been Tested?

Added tests to `tests/test_prefetching.py` with verification of receiving data from `annotate` and `select_related`.

## Checklist:

### DIFF
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -608,41 +608,41 @@ class BaseExecutor:
         model_pk = self.model._meta.pk
         instance_pks = [model_pk.to_db_value(instance.pk, instance) for instance in instance_list]
 
-        related_objects = await queryset.filter(
-            **{f"{field_object.related_name}__in": instance_pks}
-        )
+        related_model_pk = queryset.model._meta.pk
+        model_field_name_pk = related_model_pk.model_field_name
+        fields_for_select = queryset._fields_for_select
+        if fields_for_select and model_field_name_pk not in fields_for_select:
+            queryset = queryset.only(*queryset._fields_for_select, model_field_name_pk)
 
         relation_map: dict = {}
-        if related_objects:
-            related_pk_map: dict = {obj.pk: obj for obj in related_objects}
-            related_model_pk = queryset.model._meta.pk
-            related_pks = [related_model_pk.to_db_value(pk, None) for pk in related_pk_map]
+        related_objects_by_pks = {
+            obj.pk: obj
+            for obj in await queryset.filter(**{f"{field_object.related_name}__in": instance_pks})
+        }
+        if related_objects_by_pks:
             through_table = Table(field_object.through, schema=field_object.through_schema)
             backward_field = through_table[field_object.backward_key]
             forward_field = through_table[field_object.forward_key]
 
-            _, (_, through_rows) = await asyncio.gather(
-                self.__class__(
-                    model=queryset.model, db=self.db, prefetch_map=queryset._prefetch_map
-                )._execute_prefetch_queries(related_objects),
-                self.db.execute_query(
-                    *(
-                        self.db.query_class.from_(through_table)
-                        .select(backward_field, forward_field)
-                        .where(backward_field.isin(instance_pks))
-                        .where(forward_field.isin(related_pks))
-                        .get_parameterized_sql()
-                    )
-                ),
+            _, through_rows = await self.db.execute_query(
+                *(
+                    self.db.query_class.from_(through_table)
+                    .select(backward_field, forward_field)
+                    .where(backward_field.isin(instance_pks))
+                    .where(forward_field.isin(tuple(related_objects_by_pks)))
+                    .get_parameterized_sql()
+                )
             )
 
+            reverse_map: dict = {}
             for row in through_rows:
+                forward_key_value = related_model_pk.to_python_value(row[field_object.forward_key])
                 backward_key_value = model_pk.to_python_value(row[field_object.backward_key])
-                related_object = related_pk_map.get(
-                    related_model_pk.to_python_value(row[field_object.forward_key])
-                )
-                if related_object is not None:
-                    relation_map.setdefault(backward_key_value, []).append(related_object)
+                reverse_map.setdefault(forward_key_value, []).append(backward_key_value)
+
+            for related_object in related_objects_by_pks.values():
+                for instance_pk in reverse_map.get(related_object.pk, []):
+                    relation_map.setdefault(instance_pk, []).append(related_object)
 
         for instance in instance_list:
             getattr(instance, field)._set_result_for_query(
--- a/tortoise/backends/base/executor.py
+++ b/tortoise/backends/base/executor.py
@@ -601,23 +601,29 @@ class BaseExecutor:
         field: str,
         related_query: tuple[str | None, QuerySet],
     ) -> Iterable[Model]:
-        to_attr, queryset = related_query
+        to_attr, related_queryset = related_query
 
         field_object: ManyToManyFieldInstance = self.model._meta.fields_map[field]  # type: ignore
 
-        model_pk = self.model._meta.pk
-        instance_pks = [model_pk.to_db_value(instance.pk, instance) for instance in instance_list]
+        pk_field = self.model._meta.pk
+        instance_id_set: set = {
+            pk_field.to_db_value(instance.pk, instance) for instance in instance_list
+        }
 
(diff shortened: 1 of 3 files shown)


waketzheng:
Thanks for the patch. The change looks useful for enabling `select_related` and `annotate` inside M2M prefetches, but a few behaviors seem worth confirming before merging:

- `tortoise/backends/base/executor.py:639-645` rebuilds each parent's relation list from the through-table rows. That means the final order is driven by the through-table query, not by the order returned from the custom queryset. Could you confirm whether `Prefetch("participants", Team.all().order_by("name"))` is expected to preserve queryset ordering?

- `tortoise/backends/base/executor.py:624-627` already executes the custom queryset, and the result is then passed through `_execute_prefetch_queries()` again. Could you confirm whether this is intended, or whether nested prefetches on the custom queryset may be run twice?

- `tortoise/backends/base/executor.py:611-619` assumes each related object has a usable `pk`. If the queryset uses `.only()` and omits the primary key, the mapping step may not be able to match through-table rows back to related objects. Could you confirm whether `.only()` should be supported here, and if so whether the PK needs to be forced into the select list?

Overall, the direction looks promising, but these cases seem worth verifying with focused regression tests.

VladislavYar:
@waketzheng 
You're right on all counts.

1. Supported order_by. Added a test;
2. Deleted the repeated prefetch;
3. When using only, I will force the primary key to be added. Added a test.

waketzheng:
Thanks for the update. Consider this:

The previous concerns around `order_by`, nested prefetch execution, and `.only()` requiring the related PK look addressed in the latest version.

One remaining edge case may be worth confirming: in the new through-table lookup, the parent PKs are converted with `model_pk.to_db_value(...)`, but the related PKs are passed to `isin()` as Python-level `obj.pk` values:

```py
.where(forward_field.isin(tuple(related_objects_by_pks)))
```

For `UUIDField`, `obj.pk` is a `uuid.UUID`, while `UUIDField.to_db_value()` converts it to `str(value)`. This appears to make M2M `prefetch_related()` fail for UUID primary keys on SQLite with:

```text
sqlite3.ProgrammingError: Error binding parameter 2: type 'UUID' is not supported
```

A regression test in `tests/fields/test_m2m_uuid.py` should expose this, especially because the existing `m2m_uuid_models` fixture already covers both normal UUID PKs and source-field UUID PKs:

```py
@pytest.mark.asyncio
async def test_prefetch_related(db, m2m_uuid_models):
    UUIDPkModel, UUIDM2MRelatedModel = m2m_uuid_models
    one = await UUIDPkModel.create()
    two = await UUIDM2MRelatedModel.create()
    await one.peers.add(two)

    fetched = await UUIDPkModel.get(pk=one.pk).prefetch_related("peers")

    assert list(fetched.peers) == [two]
```

It may be safer to convert the related PKs before the through-table query as well, analogous to `instance_pks`, for example by using `related_model_pk.to_db_value(obj.pk, obj)` for the objects in `related_objects_by_pks.values()`.

VladislavYar:
@waketzheng Thanks! I corrected it.

waketzheng:
The latest update appears to address the UUID primary-key case. The through-table lookup now keys `related_objects_by_pks` by `related_model_pk.to_db_value(obj.pk, obj)`, so `forward_field.isin(...)` receives DB-level values instead of Python `uuid.UUID` objects.

This also covers the source-field UUID case in the same path: the related object is still used later when mapping back via `related_object.pk`, while the through-table filter uses the converted DB value.

One small follow-up that may be worth adding is a regression test in `tests/fields/test_m2m_uuid.py`, since the existing UUID M2M tests cover direct relation access but not `prefetch_related()` for this new code path. The existing `m2m_uuid_models` fixture should make it possible to cover both normal UUID PKs and source-field UUID PKs:

```py
@pytest.mark.asyncio
async def test_prefetch_related(db, m2m_uuid_models):
    UUIDPkModel, UUIDM2MRelatedModel = m2m_uuid_models
    one = await UUIDPkModel.create()
    two = await UUIDM2MRelatedModel.create()
    await one.peers.add(two)

    fetched = await UUIDPkModel.get(pk=one.pk).prefetch_related("peers")

    assert list(fetched.peers) == [two]
```

VladislavYar:
@waketzheng I added your test

tortoise/backends/base/executor.py
```diff
@@ -602,85 +601,53 @@ async def _prefetch_m2m_relation(
         field: str,
         related_query: tuple[str | None, QuerySet],
     ) -> Iterable[Model]:
-        to_attr, related_query = related_query
-        instance_id_set: set = {
-            instance._meta.pk.to_db_value(instance.pk, instance) for instance in instance_list
-        }
+        to_attr, queryset = related_query
```
waketzheng: @VladislavYar Why do you change it from `related_query` to `queryset`?

tortoise/backends/base/executor.py
```diff
@@ -602,85 +601,53 @@ async def _prefetch_m2m_relation(
...
-                relation_map[object_id] = []
-            relation_map[object_id].append(related_object_map[related_object_id])
+            for related_object in related_objects_by_pks.values():
+                for instance_pk in reverse_map.get(related_object.pk, []):
+                    relation_map.setdefault(instance_pk, []).append(related_object)
 
         for instance in instance_list:
-            relation_container = getattr(instance, field)
-            relation_container._set_result_for_query(relation_map.get(instance.pk, []), to_attr)
+            getattr(instance, field)._set_result_for_query(
```
waketzheng: One-line style reduces readability; it would be better to roll it back.

tortoise/backends/base/executor.py
```diff
@@ -602,85 +601,53 @@ async def _prefetch_m2m_relation(
...
-        to_attr, related_query = related_query
-        instance_id_set: set = {
-            instance._meta.pk.to_db_value(instance.pk, instance) for instance in instance_list
-        }
+        to_attr, queryset = related_query
 
         field_object: ManyToManyFieldInstance = self.model._meta.fields_map[field]  # type: ignore
 
-        through_table = Table(field_object.through, schema=field_object.through_schema)
+        model_pk = self.model._meta.pk
```
waketzheng: `model._meta.pk` is a field object, not a model class or instance; therefore, the variable name can be `pk_field`.

tortoise/backends/base/executor.py
```diff
@@ -602,85 +601,53 @@ async def _prefetch_m2m_relation(
...
+        instance_pks = [model_pk.to_db_value(instance.pk, instance) for instance in instance_list]
```
waketzheng: It would be better to keep `instance_id_set` instead of changing it to a list. Keeping the change minimal makes it easier to review.

tortoise/backends/base/executor.py
VladislavYar: Because the `tuple[str | None, QuerySet]` (`related_query`) type is assigned a value with the `QuerySet` type. Therefore, it is better to create a separate variable and the name of the queryset is clearer.
I will rename it to `related_queryset`, because the `_prefetch_direct_relation` method has the same name.

tortoise/backends/base/executor.py
VladislavYar: Corrected it

tortoise/backends/base/executor.py
VladislavYar: Corrected it

tortoise/backends/base/executor.py
VladislavYar: Corrected it

waketzheng:
@VladislavYar Thanks for working on this. One concern: the original implementation queried the database only once,
  but the new implementation queries it twice. Could you refactor it so the M2M prefetch still uses a single query?

VladislavYar:
@waketzheng Are you talking about: 
```python
  _, through_rows = await self.db.execute_query(
      *(
          self.db.query_class.from_(through_table)
          .select(backward_field, forward_field)
          .where(backward_field.isin(instance_id_set))
          .where(forward_field.isin(tuple(related_objects_by_pks)))
          .get_parameterized_sql()
      )
```
?

If so, then I've tried and failed at something sane.  So I took the logic from Django. Do you have any ideas?

PR #2211, opened 2026-06-04
feat: add DISTINCT ON support for PostgreSQL.

VladislavYar:
## Description

Added `DISTINCT ON (fields)` support for PostgreSQL to `QuerySet.distinct(*fields)`.

Previously, `.distinct()` only supported plain `DISTINCT` (no arguments). Now, passing field
names generates `DISTINCT ON (fields)` on PostgreSQL, keeping one row per unique combination
of the specified fields. The feature is fully propagated to `.values()`, `.values_list()`,
and `.only()`.

Also added:
- `skipCapability` decorator to `tortoise.contrib.test` — the inverse of `requireCapability`,
  skips a test when the specified capabilities match.
- `_apply_db()` helper in `_ChooseDBMixin` that consistently sets the DB connection and
  switches the query builder to `PostgreSQLQueryBuilder` across all query classes.

## Motivation and Context

PostgreSQL's `DISTINCT ON` is a powerful feature that standard `DISTINCT` cannot replace —
it allows selecting one representative row per group without a `GROUP BY`, while retaining
full model objects. This is commonly needed for "latest per group" or "first per category"
queries.

## How Has This Been Tested?

Added `tests/test_distinct.py` covering:

- Basic `DISTINCT ON` with and without `ORDER BY`
- Multiple `DISTINCT ON` fields
- Combined with `.values_list()` — single field, multiple fields, field outside `DISTINCT ON`
- Combined with `.values()` — same variations
- Combined with `.only()`
- `ORDER BY` respects the selected row within each group
- `OperationalError` when `ORDER BY` doesn't start with `DISTINCT ON` fields
- `OperationalError` when used on a non-PostgreSQL database

## Checklist:

### DIFF
--- a/tests/test_distinct.py
+++ b/tests/test_distinct.py
@@ -1,9 +1,10 @@
 import pytest
 
-from tests.testmodels import Tournament
+from tests.testmodels import Author, Book, DefaultOrdered, SourceFieldPk, Tournament
 from tortoise.contrib import test
 from tortoise.contrib.test.condition import NotIn
 from tortoise.exceptions import OperationalError
+from tortoise.functions import Count
 
 # ---------------------------------------------------------------------------
 # Basic DISTINCT (all databases)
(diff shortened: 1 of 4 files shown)


tortoise/contrib/test/__init__.py
```diff
@@ -235,6 +236,69 @@ def skip_wrapper(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
     return decorator
 
 
+def skipCapability(
```
waketzheng: So many lines of this function. I would prefer:
```
def skip_dialect(*names: str, connection_name: str = 'models') -> Callable[[_FT], _FT]:
    return requireCapability(connection_name, dialect=NotIn(*names))
```

tests/test_distinct.py
```diff
@@ -0,0 +1,204 @@
...
+
+@test.requireCapability(dialect="postgres")
+@pytest.mark.asyncio
+async def test_distinct_on_invalid_order_by(db):
+    await Tournament.create(name="1")
+    with pytest.raises(OperationalError):
+        await Tournament.all().distinct("name").order_by("desc")
+
+
+@test.skipCapability(dialect="postgres")
```
waketzheng: How about `@test.requireCapability(dialect=NotIn("postgres"))`

VladislavYar: You're right. I removed the `skipCapability` function and changed it to `requireCapability(dialect=NotIn("postgres"))`.

waketzheng:
@VladislavYar , this PR is helpful — let's go ahead with it.

Said by AI (Claude/Codex):

### Findings

  1. **`distinct("field")` is rejected on valid PostgreSQL query chains**

     In `QuerySet.distinct()` the backend check uses:

     ```python
     isinstance(self.query, PostgreSQLQueryBuilder)

  However, many common query chains are not DB-bound yet when distinct() is called, for example:

  Tournament.filter(name="x").distinct("name")
  Tournament.annotate(...).distinct("name")

  These start from manager.get_queryset(), so self.query is still the generic placeholder builder. As a result, this
  raises OperationalError("DISTINCT ON is only supported by PostgreSQL") even when the actual model DB is PostgreSQL.

  This check should be based on the chosen/model DB capability or deferred until query construction.

  2. DISTINCT ON fields are not resolved to database columns

     The implementation passes ORM field names directly to PyPika:

     self.query.distinct_on(*self._distinct_on)

     That bypasses Tortoise’s normal field resolution. Fields with source_field, relation traversals, or fields needing
     joins will generate wrong SQL.

     Example: for a model field like:

     chars = fields.CharField(source_field="some_chars_table")

     distinct("chars") should use "some_chars_table", but this code emits "chars".

     The distinct-on fields need to be resolved the same way ordering/select fields are resolved.

  3. Default Meta.ordering is not included in the DISTINCT ON validation

     resolve_ordering() applies model default ordering when self._orderings is empty, but the DISTINCT ON validation
     only checks self._orderings.

     That means a model with default ordering not starting with the distinct-on fields can produce invalid PostgreSQL
     SQL without being caught by this validation.

  ### Test Coverage Gap

  The new tests cover Tournament.all().distinct("name"), but they do not cover:

  - filter(...).distinct("field") on PostgreSQL
  - fields with source_field
  - relation / joined fields
  - models with default Meta.ordering

  Those cases should be added before merging.

VladislavYar:
@waketzheng I fixed all the errors and wrote the tests.

VladislavYar:
@waketzheng I fixed the testing errors.

PR #2217, opened 2026-06-15
Fix invalid SQL on DELETE with backward relations/joins

sumit-raghav-06:
## Description

Refactored the `DeleteQuery._make_query` method to handle deletions when filtering across relations or implicit filters that require a table `JOIN`. 

Modern SQL backends like PostgreSQL and SQLite do not natively support standard `DELETE FROM table LEFT JOIN ...` syntax. This fix detects if joins are present (`self._joined_tables`) and refactors the statement into a clean, isolated `IN` subquery fallback.

## Motivation and Context

This fixes an issue where running a `.delete()` on queries involving backward relations or joined filters throws an operational/syntax error on PostgreSQL and SQLite

## How Has This Been Tested?

- Verified that simple deletes continue to use the optimized, direct standard execution path.
- Verified that deletes involving table joins successfully refactor into the `WHERE id IN (SELECT ...)` subquery format and execute without syntax errors

## Checklist:

### DIFF
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -1428,18 +1428,55 @@ class DeleteQuery(AwaitableQuery):
 
     def _make_query(self) -> None:
         self.query = copy(self.model._meta.basequery)
-        if self.capabilities.support_update_limit_order_by and self._limit:
-            self.query._limit = self.query._wrapper_cls(self._limit)
+        self.resolve_filters()
+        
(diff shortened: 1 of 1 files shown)


sumit-raghav-06 (commented):
guide me if anything wrong

sumit-raghav-06 (commented):
it's been more than a week i fixed the existing issue on this repo kindly have alook

PR #2219, opened 2026-06-16
feat: add libsql backend for Turso/remote SQLite support

nebularen:
## Summary

Adds a `libsql` backend to Tortoise ORM, enabling async access to Turso Cloud and self-hosted libsql servers.

### How it works

The libsql backend uses the same approach as the sqlite backend — `aiosqlite`'s thread-based async bridge — but swaps `sqlite3.connect` for `libsql.connect` as the underlying connector:

```
aiosqlite.Connection(libsql_connector) → background thread → libsql → remote Turso
```

This means **90% of the code is reused** from SqliteClient. The only differences:

1. **Custom connector**: `libsql.connect(file_path, sync_url=..., auth_token=***)` instead of `sqlite3.connect()`
2. **Parameter conversion**: libsql doesn't support `sqlite3.register_adapter`, so `datetime`/`Decimal` params are pre-converted
3. **Row conversion**: libsql returns tuples (not `sqlite3.Row`), so we manually convert to dicts using `cursor.description`
4. **Affected row count**: Uses `cursor.rowcount` instead of `connection.total_changes` (libsql lacks `total_changes`)
5. **Error translation**: `ValueError` with "UNIQUE constraint" → `IntegrityError`

### Connection URL format

```python
# Turso Cloud
db_url = "libsql://my-db.turso.io"
credentials = {"auth_token": "***"}

# Self-hosted libsql server
db_url = "libsql://localhost:8080"

# Embedded replica (local file + sync to remote)
db_url = "libsql://:memory:"
credentials = {"sync_url": "libsql://my-db.turso.io", "auth_token": "***"}
```

### Files changed

- `tortoise/backends/libsql/__init__.py` — Backend registration
- `tortoise/backends/libsql/client.py` — LibsqlClient implementation
- `tortoise/backends/base/config_generator.py` — Register `libsql://` URL scheme
- `tortoise/fields/data.py` — Add `_db_libsql` dialect for DecimalField (CAST AS NUMERIC)
- `pyproject.toml` — Add `libsql` optional dependency
- `tests/test_default.py` — Add LibsqlClient to fixture

### Test results

**1542 passed, 7 failed, 123 skipped**

The 7 failures are all **libsql limitations** (not bugs in our backend):

| Failure | Root Cause |
|---------|-----------|
| BigInt overflow | libsql uses JS numbers (max ~9e18), can't store full 64-bit int |
| zoneinfo | libsql doesn't support Python timezone objects |
| NULL bytes (`\x00`) | libsql doesn't support NUL in strings |
| two_databases | libsql single-connection mode can't have two in-memory DBs |
| POW() function | libsql doesn't include the `POW` SQL function |
| datetime strftime | Minor differences in strftime behavior |

None of these affect real-world usage with Turso Cloud.

### Dependencies

- `libsql>=0.1` (optional, via `pip install tortoise-orm[libsql]`)
- Reuses existing `aiosqlite` dependency for the async bridge

### Related

- [libsql Python SDK](https://github.com/tursodatabase/libsql/tree/main/packages/libsql-ffi/python)
- [Turso docs](https://docs.turso.tech/sdk/python/quickstart)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ aiomysql = ["aiomysql"]
 asyncmy = ["asyncmy (>=0.2.11,<1.0.0)"]
 psycopg = ["psycopg[pool,binary] (>=3.0.12,<4.0.0)"]
 asyncodbc = ["asyncodbc (>=0.1.1,<1.0.0); python_version < '4.0'"]
+libsql = ["libsql>=0.1"]
 
 [project.urls]
 documentation = "https://tortoise.github.io"
--- a/tests/test_default.py
+++ b/tests/test_default.py
@@ -10,6 +10,11 @@ from tortoise.backends.asyncpg import AsyncpgDBClient
 from tortoise.backends.mysql import MySQLClient
 from tortoise.backends.psycopg import PsycopgClient
 from tortoise.backends.sqlite import SqliteClient
+
+try:
+    from tortoise.backends.libsql import LibsqlClient
+except ImportError:
+    LibsqlClient = None  # type: ignore[misc,assignment]
 from tortoise.timezone import UTC
 
 # Optional imports for database clients that require system dependencies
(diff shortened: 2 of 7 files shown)


waketzheng:
@nebularen, thanks for your interest in TortoiseORM. Using `libsql` is not acceptable because it would increase the complexity of this library. However, if you can use `pyturso` to support the Turso database instead, I would be happy to review it.

nebularen:
Thanks for the feedback! I looked into pyturso, but it's currently designed for local/embedded use only — it requires a local database file and doesn't support remote HTTP connections to Turso Cloud. For that use case, Turso's docs recommend libsql as the only Python option (see [Remote Access docs](https://docs.turso.tech/sdk/python/quickstart#remote-access-over-the-wire)).

I'll follow up if/when pyturso adds remote connection support. Thanks again!

PR #2226, opened 2026-06-22
Honor db_constraint=False in migration schema editor

### LINKED ISSUE #2223: Regarding db_constraint, I think this is a bug
**Describe the bug**
If the purpose of db_constraint is to avoid generating constraints in the database, then it does not seem to work as intended.

**To Reproduce**
Create the following data model and run the database migration:
```
class Feedback(Model):
    user = fields.ForeignKeyField('app.User', related_name='feedbacks', db_constraint=False) 
    title = fields.CharField(default='', max_length=50)
```
Expected behavior: No foreign key constraint exists in the database.
Actual behavior: Not only does a foreign key constraint get created, it is even more standardized and reliable than when db_constraint=True 😅
My actual requirement is to skip this physical database constraint entirely, while only retaining the relational annotation at the data model layer, with full support for prefetch_related, select_related and values() queries.

youdie006:
## Description

The migration schema editor always emitted the `FOREIGN KEY` reference for `ForeignKeyField` / `ManyToManyField`, ignoring `db_constraint=False` — unlike the runtime schema generator, which respects it. `grep db_constraint tortoise/migrations/` returned zero matches, confirming the migration subsystem ignored the flag.

## Motivation and Context

Closes #2223. Models that set `db_constraint=False` (to avoid a hard DB-level FK while keeping the ORM relation) still got a constrained FK in generated migrations, defeating the flag.

## How Has This Been Tested?

- `_get_fk_field_definition`: guard the `_get_fk_reference_string(...)` append on `fk_field.db_constraint` (mirrors `backends/base/schema_generator.py:264`). Covers `create_model` and `add_field`.
- `_get_m2m_table_definition`: a no-constraint through-table template selected when `not field.db_constraint` (mirrors `schema_generator.py:328/370`); the unique index is still emitted.
- 3 new tests in `tests/migrations/test_schema_editor_sql.py` (FK in `create_model`, FK via `add_field`, M2M) assert no `REFERENCES`/`FOREIGN KEY` while the column remains — red→green. Full sqlite suite: 1912 passed, 0 failures.

## Checklist

- [x] Tests added
- [x] `CHANGELOG.rst` updated
- [x] `ruff check` / `ruff format --check` clean

### DIFF
--- a/tortoise/migrations/schema_editor/base.py
+++ b/tortoise/migrations/schema_editor/base.py
@@ -31,6 +31,12 @@ class BaseSchemaEditor(SchemaQuotingMixin):
         ' ("{forward_field}") ON DELETE CASCADE\n'
         "){extra}{comment};"
     )
+    M2M_TABLE_NO_CONSTRAINT_TEMPLATE = (
+        "CREATE TABLE {table_name} (\n"
+        '    "{backward_key}" {backward_type} NOT NULL,\n'
+        '    "{forward_key}" {forward_type} NOT NULL\n'
+        "){extra}{comment};"
+    )
     RENAME_TABLE_TEMPLATE = "ALTER TABLE {old_table} RENAME TO {new_table}"
     DELETE_TABLE_TEMPLATE = "DROP TABLE {table} CASCADE"
     ADD_FIELD_TEMPLATE = "ALTER TABLE {table} ADD COLUMN {definition}"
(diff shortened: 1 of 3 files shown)


tortoise/migrations/schema_editor/base.py
```diff
@@ -31,6 +31,12 @@ class BaseSchemaEditor(SchemaQuotingMixin):
         ' ("{forward_field}") ON DELETE CASCADE\n'
         "){extra}{comment};"
     )
+    M2M_TABLE_NO_CONSTRAINT_TEMPLATE = (
```
waketzheng, line 34: Does this template work with MySQL?

PR #2227, opened 2026-06-22
feat: q to bool

VladislavYar:
## Description

Added `__bool__` support to the `Q` class, allowing `Q` objects to be evaluated in a boolean context.

A `Q` object is considered truthy if it contains any filters or any non-empty children. An empty `Q()` with no filters and no children evaluates to `False`.

## Motivation and Context

Previously, all `Q` instances were truthy by default (standard Python object behavior), making it impossible to distinguish between an empty `Q()` and a meaningful one. This change enables natural boolean checks like `if q:` to guard against applying empty query conditions.

## How Has This Been Tested?

Added `test_q_to_bool` covering four cases:

`Q(row="data")` - truthy (has filters)
`Q()` - falsy (no filters, no children)
`Q(Q(row="data"), Q(row="data"))` - truthy (has non-empty children)
`Q(Q(), Q())` - falsy (all children are empty)

## Checklist:


tests/test_q.py
```diff
@@ -19,6 +19,17 @@ def test_q_basic():
...
+def test_q_to_bool():
+    q = Q(row="data")
+    q_empty = Q()
+    q_children = Q(Q(row="data"), Q(row="data"))
+    q_children_empty = Q(Q(), Q())
+    assert bool(q) is True
+    assert bool(q_empty) is False
+    assert bool(q_children) is True
+    assert bool(q_children_empty) is False
+
```
waketzheng, line 36: Please add more test cases, such as:

```python
q = ~q
q = Q(Q(), Q(), join_type='OR')
...
```

tortoise/expressions.py
```diff
@@ -326,6 +326,9 @@ def __eq__(self, other: object) -> bool:
             and self.filters == other.filters
         )
 
+    def __bool__(self) -> bool:
+        return any((self.filters, *(children for children in self.children)))
```
waketzheng: How about using the following code?

```python
if self.filters:
    return True
return any(self.children)

waketzheng (commented):
It would be better to also update the changelog.

tortoise/expressions.py
VladislavYar: corrected it

tests/test_q.py
VladislavYar: corrected it

waketzheng (dismissed):

waketzheng:
@VladislavYar, this PR is helpful. Could you please fix the merge conflict in the CHANGELOG?

VladislavYar:
@waketzheng done

waketzheng (approved):

PR #2229, opened 2026-06-27
feat: deprecate implicit on_delete default in ForeignKeyField, OneToOneField, ManyToManyField

### LINKED ISSUE #1801: on_delete should be required

Yad4o:
## Summary

Closes #1801.

`ForeignKeyField`, `OneToOneField`, and `ManyToManyField` silently defaulted `on_delete` to `CASCADE`. This is dangerous — a developer who forgets to set `on_delete` may inadvertently cascade-delete data they didn't intend to.

This PR adds a `DeprecationWarning` when `on_delete` is omitted, making the implicit behaviour visible and encouraging explicit intent. The default behaviour is preserved for now to avoid a breaking change.

## Changes

- Added `_UNSET` sentinel in `tortoise/fields/relational.py` to detect when `on_delete` is not passed
- `ForeignKeyFieldInstance.__init__`: emits `DeprecationWarning` when `on_delete` is not explicitly provided
- `OneToOneFieldInstance.__init__`: same (delegates to `ForeignKeyFieldInstance`)
- `ManyToManyFieldInstance.__init__`: same
- All three public factory functions (`ForeignKeyField`, `OneToOneField`, `ManyToManyField`) updated to use the sentinel as default
- `@overload` signatures are unchanged (type checkers still see `OnDelete = CASCADE`) so no type-checking regressions

## Before / After

```python
# Before — silent CASCADE, no indication this is dangerous
class MyModel(Model):
    user = fields.ForeignKeyField("models.User")  # silently cascades on delete

# After — explicit warning shown at import/class-definition time
# DeprecationWarning: Not passing `on_delete` to ForeignKeyField is deprecated and will be an
# error in a future release. Pass `on_delete` explicitly (e.g. on_delete=fields.CASCADE).

# Correct usage going forward:
class MyModel(Model):
    user = fields.ForeignKeyField("models.User", on_delete=fields.CASCADE)
```

## Notes

- Backward compatible: existing code still works, just warns
- The warning fires at model class definition time (not at query time), so it's easy to catch
- Future release can make `on_delete` truly required by removing the sentinel fallback

### DIFF
--- a/tortoise/fields/relational.py
+++ b/tortoise/fields/relational.py
@@ -16,6 +16,8 @@ if TYPE_CHECKING:  # pragma: nocoverage
 
 MODEL = TypeVar("MODEL", bound="Model")
 
+_UNSET: object = object()  # sentinel for detecting omitted on_delete
+
 
 class _NoneAwaitable:
     __slots__ = ()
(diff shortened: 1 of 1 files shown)


waketzheng:
For compatibility, I prefer using warnings rather than raising exceptions, but I don't like having too many `type: ignore` comments.

PR #2239, opened 2026-07-08
Fix filtering by FK instance to use to_field instead of pk

### LINKED ISSUE #2225: [Bug] ForeignKey ignores to_field and forces lookup on default PK field, causing type/mapping errors
**Describe the bug**
When using a ForeignKeyField with `to_field` pointing to a non-primary key field (e.g., app_id which is a UUID/CharField), Tortoise ORM fails to correctly resolve the column for lookups.

Instead of using the specified `to_field` (e.g., app_id), the ORM incorrectly defaults to the model's primary key name (e.g., id) and generates SQL using the `${fk_name}_id` pattern. This causes the generated query to filter by the wrong column or even use the wrong data type (e.g., passing a UUID string to an integer-based ID column), leading to runtime type errors or incorrect results.

**To Reproduce**
Steps to reproduce:

1. Define a model (Application) with a primary key (id) and a secondary unique field (app_id).

2. Define a second model (TableB) with a ForeignKeyField pointing to Application, setting `to_field="app_id"`.

3. Perform a filter on TableB using an Application instance.

Minimal code snippet:

```python
class Application(Model):
    id = fields.IntField(pk=True) # Default auto-increment PK
    app_id = fields.CharField(unique=True, max_length=50) # The custom field we want to link to

class TableB(Model):
    app = fields.ForeignKeyField(
        "models.Application", 
        to_field="app_id", 
        db_constraint=False
    )

# Execution
app_instance = await Application.get(app_id="550e8400-e29b-41d4-a716-446655440000")
# The query incorrectly generates: ... WHERE app_id = 1 (or tries to use the internal pk)
# instead of: ... WHERE app_app_id = '550e8400-e29b-41d4-a716-446655440000'
await TableB.filter(app=app_instance)
```

**Expected behavior**
The ORM should respect the `to_field` configuration. When filtering by a related model instance, it should resolve the link using the column mapping associated with `to_field`(app_app_id in the example above) rather than defaulting to `${fk_name}_id`.

**Additional context**
This issue effectively makes using to_field for non-primary key associations impossible for query lookups.

It appears that the query builder incorrectly assumes that the related field is always the Primary Key of the target model, ignoring the `to_field` metadata in the relationship.

Tortoise ORM version: 1.1.7

**Technical Analysis of the Bug**
My initial investigation points to the `_get_actual_filter_params` method as a likely culprit, where the logic seems to prioritize `pk` over the explicitly defined `to_field`.

Specifically, when processing a ForeignKeyField or OneToOneField, the ORM uses the following logic to resolve the filter value:

https://github.com/tortoise/tortoise-orm/blob/bc4f971bb840c90af422108b8cf07a68c04cbd65/tortoise/expressions.py#L432-L462

The logic flaw:
- The filter_value is hardcoded to retrieve `value.pk`.
- It fails to check if the ForeignKeyField has a custom `to_field` attribute defined.

When `to_field` is specified (e.g., pointing to a UUID column), the ORM ignores this configuration and incorrectly extracts the primary key (id) instead, leading to type mismatches and invalid SQL filters.

**Suggested Fix:**
The implementation should inspect the field_object for a `to_field` attribute and use that to resolve the value from the value instance, falling back to .pk only if `to_field` is not defined.

AnayGarodia:
Fixes #2225

When a `ForeignKeyField`/`OneToOneField` is declared with a `to_field` that points to a non-primary-key column, filtering by a related model instance used the instance's primary key instead of the `to_field` value.

In `Q._get_actual_filter_params`, the fk/o2o branch did `filter_value = getattr(value, "pk", value)`, hardcoding `.pk`. So for a FK with `to_field="app_id"`, `Model.filter(fk=instance)` generated SQL comparing the `<fk>_id` column (which stores the `to_field` value) against `instance.pk`, targeting the wrong value and, when the types differ (e.g. a UUID pk vs an int `to_field`), raising type errors or silently returning wrong results.

This resolves the `to_field` column via `field_object.to_field_instance.model_field_name` and reads that attribute off the instance. For the default case where `to_field` is the primary key, `to_field_instance.model_field_name` is the pk attribute, so `getattr` returns `value.pk` exactly as before; raw-value filters (`filter(fk=7)`) still fall through to the value via `getattr`'s default.

The existing test models already expose this shape: `School`'s primary key is `uuid`, while `Student.school` uses `to_field="id"` (a non-pk unique field). The added regression test in `tests/test_relations_with_unique.py` creates a `School` whose `id` differs from its pk and asserts `Student.filter(school=school)` finds the student; it fails on the current code (filters by the pk) and passes with this change. The full `test_relations` and `test_relations_with_unique` suites pass.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Added
 
 Fixed
 ^^^^^
+- Filtering by a related model instance on a ``ForeignKeyField``/``OneToOneField`` declared with a non-primary-key ``to_field`` now uses the ``to_field`` value instead of the instance's primary key, so the generated query targets the correct column. (#2225)
 - ``MigrationRecorder`` now uses parameterized queries; fixes MariaDB/MySQL rejecting ISO-8601 ``applied_at`` values. (#2132)
 - ``db_default`` on ``ForeignKeyField``/``OneToOneField`` now propagates to the underlying ``<fk>_id`` column, so ``CREATE TABLE`` emits the ``DEFAULT`` clause for FK columns. (#2199)
 - ``MigrationRecorder`` no longer emits tortoise's own ``pk`` field ``DeprecationWarning`` when applying migrations; it now builds its bookkeeping model with ``primary_key=True``. (#2203)
--- a/tests/test_relations_with_unique.py
+++ b/tests/test_relations_with_unique.py
@@ -44,3 +44,19 @@ async def test_relation_with_unique(db):
     assert fetched_principal.name == "Sang-Heon Jeon3"
     fetched_school = await School.filter(name="School1").prefetch_related("principal").first()
     assert fetched_school.name == "School1"
+
+
+@pytest.mark.asyncio
+async def test_filter_by_fk_instance_uses_to_field(db):
+    # https://github.com/tortoise/tortoise-orm/issues/2225
+    # School's primary key is `uuid`, but Student.school points to to_field="id"
+    # (a non-PK unique field). Filtering by a School instance must use the
+    # `to_field` value (school.id), not the primary key (school.uuid).
+    school = await School.create(id=42, name="School1")
+    await Student.create(name="Sang-Heon Jeon1", school=school)
+
+    assert school.pk != school.id
+
+    found = await Student.filter(school=school).first()
+    assert found is not None
+    assert found.name == "Sang-Heon Jeon1"
--- a/tortoise/expressions.py
+++ b/tortoise/expressions.py
@@ -437,9 +437,10 @@ class Q:
             key in resolve_context.model._meta.fk_fields
             or key in resolve_context.model._meta.o2o_fields
         ):
-            field_object = resolve_context.model._meta.fields_map[key]
+            field_object = cast(RelationalField, resolve_context.model._meta.fields_map[key])
             filter_key = cast(str, field_object.source_field)
-            filter_value = getattr(value, "pk", value)
+            to_field_name = field_object.to_field_instance.model_field_name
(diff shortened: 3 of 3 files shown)


CHANGELOG.rst
```diff
@@ -20,6 +20,7 @@ Added
 
 Fixed
 ^^^^^
+- Filtering by a related model instance on a ``ForeignKeyField``/``OneToOneField`` declared with a non-primary-key ``to_field`` now uses the ``to_field`` value instead of the instance's primary key, so the generated query targets the correct column. (#2225)
```
waketzheng, line 23: Move to 1.1.9, and use `#2239` instead of `#2225`

tests/test_relations_with_unique.py
```diff
@@ -44,3 +44,19 @@ async def test_relation_with_unique(db):
...
+    # (a non-PK unique field). Filtering by a School instance must use the
+    # `to_field` value (school.id), not the primary key (school.uuid).
+    school = await School.create(id=42, name="School1")
+    await Student.create(name="Sang-Heon Jeon1", school=school)
+
+    assert school.pk != school.id
+
+    found = await Student.filter(school=school).first()
+    assert found is not None
+    assert found.name == "Sang-Heon Jeon1"
```
waketzheng, line 62: Add more test case, such as:
```
student = await Student.create(name="Sang-Heon Jeon1", school=school)
assert student.school_id == school.pk
...
found2 = await Student.filter(school_id=school.pk).get()
assert found == found2
```

waketzheng (changes requested):

PR #2243, opened 2026-07-09
Update the return type hint for `IntEnumField`/`CharEnumField` from `EnumType` to `tortoise.fields.Field[EnumType]` (fixes #2155)

### LINKED ISSUE #2155: Invalid type hint for CharEnumField in generated migration script
**Describe the bug**
If I generate a migration file with `tortoise migrate` for a model that contains a `CharEnumField`, and then run mypy on my project, it produces an error. The reason seems to be that `CharEnumField` does not return an instance of `Field`.

When the migration creates a new table (`ops.CreateModel`) the following error is printed:

```
error: List item 1 has incompatible type "tuple[str, ExampleEnum]"; expected "tuple[str, Field[Any] | ManyToManyRelation[Any] | None]"  [list-item]
```

When the migration adds a `CharEnumField` to an existing table, the following error is printed:

```
error: Argument "field" to "AddField" has incompatible type "ExampleEnum"; expected "Field[Any] | ManyToManyRelation[Any] | None"  [arg-type]
```

**To Reproduce**
Below is a minimal reproducible example.

models.py:

```python
from tortoise.fields import CharEnumField

import enum


class ExampleEnum(enum.StrEnum):
    Value1 = enum.auto()
    Value2 = enum.auto()


class ExampleModel(Model):
    field: ExampleEnum = CharEnumField(ExampleEnum)
```

config.json:

```json
{
    "connections": {
        "default": "sqlite://:memory:"
    },
    "apps": {
        "monitor": {
            "models": ["models"],
            "migrations": "migrations",
            "default_connection": "default"
        }
    }
}
```

Steps to reproduce:
1. `tortoise --config-file config.json makemigrations`
2. `mypy migrations`

**Expected behavior**
No mypy errors on scripts that are generated by Tortoise.

apoorvdarshan:
## Summary

`makemigrations` generated migration files that fail `mypy` when a model uses `CharEnumField` / `IntEnumField` (issue #2155):

```
error: List item 1 has incompatible type "tuple[str, ExampleEnum]";
  expected "tuple[str, Field[Any] | ManyToManyRelation[Any] | None]"   # CreateModel
error: Argument "field" to "AddField" has incompatible type "ExampleEnum";
  expected "Field[Any] | ManyToManyRelation[Any] | None"               # AddField
```

The root cause is that the public `IntEnumField` / `CharEnumField` names are **factory functions** typed to return the *enum type* (a `TypeVar` bound to `Enum`) rather than a `Field`. That return type is intentional — it lets models write `field: MyEnum = CharEnumField(MyEnum)` and have the attribute resolve to the enum. But the migration writer rendered field expressions using those same factory names, so the generated `fields.CharEnumField(...)` call was inferred by mypy as the enum type, not a `Field`, and the surrounding migration operation rejected it.

## Fix

Render enum fields as their concrete `IntEnumFieldInstance` / `CharEnumFieldInstance` classes instead of the factory names. These are the real `Field` subclasses (`CharField` / `SmallIntField` descendants), so the generated code type-checks cleanly while constructing an **identical** field at runtime.

- `tortoise/migrations/writer.py`: exclude the two enum `*FieldInstance` classes from the `FieldInstance` → `Field` name remapping, so they render as their concrete class.
- `tortoise/fields/__init__.py`: export `CharEnumFieldInstance` and `IntEnumFieldInstance` so the generated `from tortoise import fields` alias resolves `fields.CharEnumFieldInstance(...)`.

Model-side typing is untouched: the factory functions still return the enum type, so `field: MyEnum = CharEnumField(MyEnum)` continues to type-check.

## Tests / Verification

- Reproduced the issue end-to-end with the model/config from the report: `tortoise --config-file config.json makemigrations` then `mypy migrations`. Before the fix mypy emitted both reported errors; after the fix `mypy migrations` reports **Success: no issues found** for both the `CreateModel` (initial) and `AddField` cases.
- Confirmed the generated migrations still apply at runtime (`tortoise ... migrate` → OK).
- Updated `test_writer_handles_enum_fields` and added `test_writer_enum_field_render_is_field_instance` (a #2155 regression test covering the `AddField` path). Both fail on the pristine tree and pass with the fix.
- `pytest tests/migrations/` (300 passed) and `pytest tests/fields/` (enum + field tests, all pass).
- `ruff format --check`, `ruff check`, and `mypy` are clean on the changed files (no new errors vs. the pristine tree).
- Added a `CHANGELOG.rst` entry under 1.1.8 → Fixed.

Disclosure: prepared with AI assistance; reviewed and verified locally.


waketzheng:
@apoorvdarshan Thanks for your work. It would be better to correct the return type of `IntEnumField`/`CharEnumField` rather than change the migration behavior. I've updated this PR to fix the issue.

waketzheng (approved):

Issue #359, opened 2020-04-20
[0.15.22] asyncpg error: Column "X" must appear in group by clause or be used in an aggregate function


bbedward:
Hi,

Recently encountered an issue with newer versions of tortoise-orm with the following query:

```py
Account.filter(wallet=wallet).annotate(max_index=Max("account_index")).order_by('-account_index').first()
```

The schema is as follows:

Account.py

```py
class Account(Model):
    wallet  = fields.ForeignKeyField('db.Wallet', on_delete=fields.CASCADE, related_name='accounts', index=True)
    address = fields.CharField(max_length=65)
    account_index = fields.IntField()
    work = fields.BooleanField(default=True)
    created_at = fields.DatetimeField(auto_now_add=True)

    class Meta:
        table = 'accounts'
        unique_together = ('address', 'account_index')
```

Wallet.py

```py
class Wallet(Model):
    id = fields.UUIDField(pk=True)
    seed = fields.CharField(max_length=128, unique=True)
    representative = fields.CharField(max_length=65, null=True)
    encrypted = fields.BooleanField(default=False)
    work = fields.BooleanField(default=True)
    created_at = fields.DatetimeField(auto_now_add=True)

    class Meta:
        table = 'wallets'
```

The exception is as follows:

```
ERROR:aiohttp.server:Error handling request
Traceback (most recent call last):
  File "/home/pippin/.local/lib/python3.6/site-packages/tortoise/backends/asyncpg/client.py", line 36, in translate_exceptions_
    return await func(self, *args)
  File "/home/pippin/.local/lib/python3.6/site-packages/tortoise/backends/asyncpg/client.py", line 172, in execute_query
    rows = await connection.fetch(*params)
  File "/home/pippin/.local/lib/python3.6/site-packages/asyncpg/connection.py", line 420, in fetch
    return await self._execute(query, args, 0, timeout)
  File "/home/pippin/.local/lib/python3.6/site-packages/asyncpg/connection.py", line 1403, in _execute
    query, args, limit, timeout, return_status=return_status)
  File "/home/pippin/.local/lib/python3.6/site-packages/asyncpg/connection.py", line 1411, in __execute
    return await self._do_execute(query, executor, timeout)
  File "/home/pippin/.local/lib/python3.6/site-packages/asyncpg/connection.py", line 1423, in _do_execute
    stmt = await self._get_statement(query, None)
[...]

PR #2246, opened 2026-07-15
docs: add complex query example with grouping and date filters

### LINKED ISSUE #418: How to emulate such a request through Tortoise ORM?
Hello, how I can a similar SQL request be sampled via Tortoise ORM? Read the documentation, but it doesn’t give examples of working with date and time

```sql
SELECT uid, SUM(posts) AS post, SUM(fine) AS fines, date FROM statistic
WHERE date > DATE(date) <= NOW() AND date >= DATE_SUB(NOW(), INTERVAL 31 DAY)
GROUP BY uid ORDER BY post DESC
```

pctablet505:
Fixes #418

Issue #418 asked how to write a report-style query in Tortoise that mirrors a raw SQL aggregation over a date range. The original example summed posts and fine per uid, filtered to the last 31 days, grouped by uid, and ordered by the aggregated sum. grigi noted in the thread that Tortoise didn't expose date-manipulation filters at the time, and the existing docs didn't show how to wire together annotate, filter, group_by, and values for this kind of query. They also pointed to Tortoise.get_connection('default').execute_query_dict(...) as a raw-SQL escape hatch, which we documented separately, but a first-party ORM example was still missing.

This change adds a "Complex queries" section to docs/query.rst with a concrete, copy-pasteable example. It uses date__gte and date__lte with plain date values, which avoids depending on database-specific date arithmetic and matches the DateField case from the report. The original SQL selected date while grouping only by uid; on strict databases that runs into the "column must appear in the GROUP BY clause or be used in an aggregate function" problem (see #359), so the example aggregates across the full date range and returns only uid, post, and fines. A note reminds readers to use datetime.now() instead of date.today() when the field is a DateTimeField, and that group_by() must come before values() or values_list().

I built the docs locally and confirmed the new section renders correctly.

### DIFF
--- a/docs/query.rst
+++ b/docs/query.rst
@@ -179,33 +179,40 @@ last 31 days can be written as:
 
 .. code-block:: python3
 
-    from datetime import datetime, timedelta, timezone
+    from datetime import date, timedelta
     from tortoise.functions import Sum
 
-    now = datetime.now(timezone.utc)
-    since = now - timedelta(days=31)
+    today = date.today()
+    since = today - timedelta(days=31)
 
     results = await (
-        Statistic.filter(date__gte=since, date__lte=now)
+        Statistic.filter(date__gte=since, date__lte=today)
         .annotate(post=Sum("posts"), fines=Sum("fine"))
-        .group_by("uid", "date")
+        .group_by("uid")
         .order_by("-post")
-        .values("uid", "post", "fines", "date")
+        .values("uid", "post", "fines")
     )
 
 This is roughly equivalent to the following SQL:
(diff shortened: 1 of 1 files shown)


pctablet505:
Closing this — on a closer look at #418, everything it asked about is already covered in the current docs: the raw-SQL escape hatch was documented back in 2020 (Manual SQL section in docs/examples/basic.rst), date lookups like __year/__month/__day are documented in docs/query.rst, and examples/group_by.py already shows the annotate + group_by pattern. There's nothing left for this PR to add.

PR #2248, opened 2026-07-15
Fix BlackSheep cross-task database access

### LINKED ISSUE #2244: contrib.blacksheep.register_tortoise breaks under ASGI: "No TortoiseContext is currently active" on every DB query (missing _enable_global_fallback=True)
**Describe the bug**

With Tortoise ORM 1.1.7, `tortoise.contrib.blacksheep.register_tortoise` sets up
the ORM in the app's `on_start` (ASGI lifespan) hook, but **any DB query inside a
request handler** raises:

```
RuntimeError: No TortoiseContext is currently active. Use 'async with
TortoiseContext() as ctx:' to create one, or call Tortoise.init() for global state.
```

The lifespan hook runs in a different asyncio task than request handlers. Since
1.x stores connection state in a `contextvars.ContextVar` (`_current_context`),
the context entered during startup is not visible in handler tasks, and no global
fallback is set — so `require_context()` fails.

The FastAPI integration does not have this problem because
`RegisterTortoise` defaults `_enable_global_fallback=True`
(`tortoise/contrib/fastapi/__init__.py`, `_enable_global_fallback: bool = True`).
The blacksheep integration calls `Tortoise.init(...)` **without** that flag, so
the global fallback context described in `tortoise/context.py` (lines ~48-52) is
never enabled.

**To Reproduce**

```python
import asyncio
from tortoise import Tortoise

CONFIG = {
    "apps": {"models": {"models": ["__main__"], "default_connection": "default"}},
}

class User(models.Model):
    id = fields.IntField(pk=True)

async def run(enable_fallback: bool) -> str:
    async def init_task():
        await Tortoise.init(config=CONFIG, _enable_global_fallback=enable_fallback)
    await asyncio.create_task(init_task())          # like the lifespan task

    async def query_task():
        await User.all().count()                    # like a request handler
        return "ok"
    try:
        return await asyncio.create_task(query_task())
    except RuntimeError as exc:
        return f"RuntimeError: {exc}"
    finally:

async def main():
    print("without:", await run(False))
    print("with:   ", await run(True))

asyncio.run(main())
```

Output:

```
without: RuntimeError: No TortoiseContext is currently active. ...
with:    ok
```

A full repro with an actual BlackSheep app + `register_tortoise` shows the same
`RuntimeError` on the first request that touches the DB.

**Expected behavior**

`register_tortoise` should make the ORM usable from request handlers out of the
box, the same as the FastAPI integration.

**Additional context**

Pass `_enable_global_fallback=True` in the blacksheep integration's
`Tortoise.init(...)` call (matching `contrib.fastapi.RegisterTortoise`).

Sanjays2402:
## Description
Enable the global context fallback when the BlackSheep integration initializes Tortoise.

## Motivation and Context
BlackSheep startup and request handlers run in separate asyncio tasks. Without the fallback, every request-side database query raises `No TortoiseContext is currently active`. Fixes #2244.

## How Has This Been Tested?
The regression test fails without the flag and passes with it. A BlackSheep lifecycle probe also initialized the ORM in one task and successfully queried SQLite from another.

## Checklist:
- [ ] My change requires a change to the documentation. (Not required.)
- [ ] I have updated the documentation accordingly. (Not required.)
- [ ] I have added the changelog accordingly. (Not required.)
- [ ] All new and existing tests passed. (The focused test passed; the full suite was not run.)


waketzheng:
Please fix lint issue.

Sanjays2402:
The failing job was `check` — mypy, not ruff:

```
tests/contrib/test_blacksheep.py:24: error: Call to untyped function "start" in typed context  [no-untyped-call]
tests/contrib/test_blacksheep.py:33: error: Call to untyped function "stop" in typed context  [no-untyped-call]
```

`disallow_untyped_calls = true` applies to tests too (the `tests.*` override only relaxes `check_untyped_defs`/`disallow_untyped_defs`), and blacksheep's `Application.start`/`stop` are unannotated upstream. Added `# type: ignore[no-untyped-call]` on the two call sites, matching how tests elsewhere in the repo handle third-party typing gaps.

`mypy tests/contrib/test_blacksheep.py` and `ruff format --check`/`ruff check` are clean, and `pytest tests/contrib/test_blacksheep.py` is 1 passed. Pushed 9d35ff7.

waketzheng:
Looks good. Could you update the changelog?

Sanjays2402:
Done — added a changelog entry under 1.1.8 (Fixed) describing the global connection fallback in register_tortoise. Pushed as cb5342c.

waketzheng (approved):

PR #2249, opened 2026-07-16
ci: pin third-party actions to full commit SHAs (matching the existing setup-uv pin)

kobihikri:
Hi, and thank you for Tortoise ORM.

CI supply-chain hardening that matches a practice you already follow. Every workflow already SHA-pins `astral-sh/setup-uv` (`…08807647… # v8.1.0`), but four third-party actions in secret-bearing jobs are still on mutable refs:

- `peaceiris/actions-gh-pages@v3` (×2, holds `PERSONAL_TOKEN` that pushes to `tortoise/tortoise.github.io`)
- `CodSpeedHQ/action@v3` (holds `CODSPEED_TOKEN`)
- `pypa/gh-action-pypi-publish@release/v1` (a mutable **branch**, holds `pypi_password`)

This PR pins each to the commit its current ref points at, with a version comment — same style as your setup-uv pin. I stayed on the v3 line for the two `@v3` actions (no major-version change) and used the current v1 release for pypi-publish. Dependabot's `github-actions` ecosystem can bump these going forward.

Verified SHAs: peaceiris v3.9.3 `373f7f2…`, CodSpeed v3.8.1 `76578c2…`, pypi-publish v1.14.0 `6733eb7…`.

For transparency: I used AI assistance to spot and draft this; I verified the refs and resolved the SHAs myself.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -26,7 +26,7 @@ jobs:
         run: make build
 
       - name: Run benchmarks
-        uses: CodSpeedHQ/action@v3
+        uses: CodSpeedHQ/action@76578c2a7ddd928664caa737f0e962e3085d4e7c # v3.8.1
         with:
           token: ${{ secrets.CODSPEED_TOKEN }}
           run: uv run --frozen pytest tests/benchmarks --codspeed
--- a/.github/workflows/gh-pages-release.yml
+++ b/.github/workflows/gh-pages-release.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Build docs
         run: make docs
       - name: Deploy release docs
-        uses: peaceiris/actions-gh-pages@v3
+        uses: peaceiris/actions-gh-pages@373f7f263a76c20808c831209c920827a82a2847 # v3.9.3
         with:
           personal_token: ${{ secrets.PERSONAL_TOKEN }}
           publish_dir: build/html
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Build docs
         run: make docs
       - name: Deploy latest docs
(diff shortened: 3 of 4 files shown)


waketzheng:
Thanks for this PR — pinning third-party actions to full commit SHAs is a solid security hardening step, and it's consistent with the existing setup-uv pin style. I've reviewed the changes and unfortunately found two blocking SHA mismatches, plus a few notes:

Blocking: incorrect SHA for pypa/gh-action-pypi-publish
The PR description states pypa-publish v1.14.0 6733eb7..., but the actual commit SHA for v1.14.0 is cef221092ed1bacb1cc03d23a2d87d1d172e277b. The 6733eb7 prefix doesn't correspond to any commit in that repo. Could you re-verify? If the current value is wrong, the publish job will fail — or worse, resolve to an unintended commit.

Blocking: incorrect SHA for CodSpeedHQ/action
I ran git ls-remote https://github.com/CodSpeedHQ/action.git refs/tags/v3.8.1 and got:
5322369bbf5359f5719bd337a19b4bcf4781fe3a refs/tags/v3.8.1
So the correct SHA is 5322369bbf5359f5719bd337a19b4bcf4781fe3a, not 76578c2.... Please update this one as well.

Note: consider using release/v1 SHA instead of a fixed release SHA for pypi-publish
The gh-action-pypi-publish maintainers explicitly recommend using the release/v1 branch pointer (or a v1 tag) rather than pinning to a specific release SHA, because the action releases frequently.

peaceiris SHA — confirmed correct
373f7f263a76c20808c831209c920827a82a2847 matches v3.9.3.

Happy to approve once the two SHA issues above are fixed.

waketzheng (changes requested):
Would better to add a `.github/dependabot.yml` like:
```
version: 2
updates:
  # GitHub Actions: daily/weekly/monthly
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"
```

PR #2253, opened 2026-07-25
fix(queryset): preserve select_for_update in values and values_list

### LINKED ISSUE #2250: `.values_list()` / `.values()` silently drop `select_for_update()` - rows are not locked
**Describe the bug**

Chaining `.values_list()` or `.values()` after `.select_for_update()` silently loses the `FOR UPDATE` clause: the generated SQL contains no locking clause and no error or warning is raised, so code that relies on the row locks (e.g. a periodic job selecting ids to update, guarded against a concurrent writer) is racy while looking correct.

The cause is visible in `tortoise/queryset.py`: `QuerySet.values_list()` and `QuerySet.values()` build `ValuesListQuery` / `ValuesQuery` by passing an explicit list of attributes, and the `_select_for_update*` family is not among them (the constructors do not even accept it), so the flag set by `select_for_update()` is dropped instead of being propagated the way `_clone()` propagates it. Present on `1.1.7` and on current `develop`.

**To Reproduce**

```python
import asyncio

from tortoise import Tortoise, fields


class Item(Model):
    name = fields.CharField(max_length=50)


async def main() -> None:
    await Tortoise.init(
        db_url="postgres://postgres:postgres@localhost:5432/postgres",
    )

    base = Item.filter(name="x").select_for_update()
    print("plain queryset: ", base.sql())
    print("values_list():  ", base.values_list("id", flat=True).sql())
    print("values():       ", base.values("id").sql())

asyncio.run(main())
```

Output (tortoise-orm 1.1.7, PostgreSQL 18, asyncpg):

```text
plain queryset:  SELECT "name","id" FROM "item" WHERE "name"=$1 FOR UPDATE
values_list():   SELECT "id" "0" FROM "item" WHERE "name"=$1
values():        SELECT "id" "id" FROM "item" WHERE "name"=$1
```

**Expected behavior**

Either of:

1. `values_list()` / `values()` keep the `FOR UPDATE` clause (`SELECT "id" "0" FROM "item" WHERE "name"=$1 FOR UPDATE`), matching how `limit`, `offset`, `distinct`, `order_by` and `force_index` survive the conversion; or
2. the combination raises, like `.values_list()` after `.only()` already does (`ValueError(".values_list() cannot be used with .only()")`), so the lock can never be lost silently.

Silently dropping a locking clause seems like the worst of the options - we found this only because a `pytest`-under-PostgreSQL race test failed: a cron doing `Model.filter(...).select_for_update().values_list("id", flat=True)` inside a transaction did not block a concurrent writer and clobbered its committed update. Workaround we settled on: `.select_for_update().only("id")` (a plain queryset keeps the clause).

**Additional context**

- tortoise-orm: 1.1.7 (also reproducible reading `develop` sources)
- Python: 3.14
- Database: PostgreSQL 18 (asyncpg)
- Possibly related but distinct: #1297 (subquery loses `FOR UPDATE`), #702 (`update_or_create` does not lock) - this one is specifically about the `values_list()` / `values()` conversion discarding the flag.

arimu1:
Fixes #2250

When `values()` or `values_list()` was chained after `select_for_update()`, the select_for_update parameters were not passed to `ValuesQuery` / `ValuesListQuery`, causing the SQL generator to drop `FOR UPDATE` clauses and lose row locking behavior.

This change passes and preserves `select_for_update` parameters in `ValuesQuery` and `ValuesListQuery`.

### DIFF
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -777,6 +777,19 @@ async def test_select_for_update(db, intfields_data):
 
 
 @pytest.mark.asyncio
+async def test_select_for_update_values(db, intfields_data):
+    qs = IntFields.filter(pk=1)
+    object.__setattr__(qs.capabilities, "_mutable", True)
+    qs.capabilities.support_for_update = True
+    sql_values = qs.select_for_update().values("id").sql()
+    sql_values_list = qs.select_for_update().values_list("id", flat=True).sql()
+
+    assert "FOR UPDATE" in sql_values
+    assert "FOR UPDATE" in sql_values_list
+
+
+
+@pytest.mark.asyncio
 async def test_select_related(db):
     tournament = await Tournament.create(name="1")
     reporter = await Reporter.create(name="Reporter")
--- a/tests/test_queryset.py
+++ b/tests/test_queryset.py
@@ -788,7 +788,6 @@ async def test_select_for_update_values(db, intfields_data):
     assert "FOR UPDATE" in sql_values_list
 
 
-
 @pytest.mark.asyncio
 async def test_select_related(db):
     tournament = await Tournament.create(name="1")
--- a/tortoise/queryset.py
+++ b/tortoise/queryset.py
@@ -698,6 +698,11 @@ class QuerySet(AwaitableQuery[MODEL]):
             group_bys=self._group_bys,
             force_indexes=self._force_indexes,
             use_indexes=self._use_indexes,
+            select_for_update=self._select_for_update,
+            select_for_update_nowait=self._select_for_update_nowait,
+            select_for_update_skip_locked=self._select_for_update_skip_locked,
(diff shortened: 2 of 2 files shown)


tests/test_queryset.py
```diff
@@ -776,6 +776,18 @@ async def test_select_for_update(db, intfields_data):
         assert sql5 == "SELECT `id` `id` FROM `intfields` WHERE `id`=%s FOR UPDATE"
 
 
+@pytest.mark.asyncio
+async def test_select_for_update_values(db, intfields_data):
+    qs = IntFields.filter(pk=1)
+    object.__setattr__(qs.capabilities, "_mutable", True)
```
waketzheng, line 782: Why not `setattr`?

waketzheng: Or just `qs.capabilities._mutable = True`

Issue #2034, opened 2025-12-12
AWS IAM based authentication (RDS / Aurora) support


andreleoni:
**Is your feature request related to a problem? Please describe.**

Currently, Tortoise-ORM does not support AWS IAM–based authentication for database connections (e.g. Amazon RDS or Aurora).
This forces applications running on AWS (ECS, EKS, Lambda, EC2) to rely on static database passwords stored in environment variables or secret managers.

This is problematic for security-sensitive environments where:

- Password rotation is required
- Long-lived credentials are discouraged
- IAM-based authentication is the recommended AWS best practice
- Other ORMs/drivers in the ecosystem already support IAM auth via token-based connections, but Tortoise-ORM currently lacks a native or first-class solution.

-----

**Describe the solution you'd like**

Add native support for AWS IAM authentication in Tortoise-ORM database backends (PostgreSQL and MySQL), allowing connections using short-lived IAM tokens instead of static passwords.

Ideally, this would include:

- An optional iam_auth or similar flag in the DB config
- Automatic generation and refresh of IAM auth tokens using boto3
- Compatibility with async drivers (asyncpg, aiomysql)
- Support for common AWS environments (ECS, EKS, Lambda, EC2 with IAM roles)

-----

**Describe alternatives you've considered**

Generating IAM tokens manually with boto3 and injecting them as passwords, which is fragile and hard to maintain

- Wrapping the database driver directly, bypassing Tortoise-ORM abstractions
- Falling back to static credentials stored in Secrets Manager or environment variables
- Using other ORMs or lower-level drivers that already support IAM auth
- All of these either add operational complexity or defeat the purpose of using Tortoise-ORM as a high-level async ORM.

-----

**Additional context**

AWS strongly recommends IAM authentication for RDS/Aurora in production environments

- IAM tokens are valid for only 15 minutes, so transparent refresh is essential
- This feature would significantly improve Tortoise-ORM adoption in AWS-native architectures
- Similar functionality exists in other ecosystems (e.g. SQLAlchemy with custom auth, direct asyncpg usage)
- If helpful, I’m willing to assist with design discussion, testing, or documentation.

PR #2261, opened 2026-08-11
feat(postgres): allow password to be a sync/async callable

get-bowen:
## Description

Allow the PostgreSQL `password` credential to be a **sync or async callable**, resolved once per new connection.

Short-lived database credentials — AWS RDS/Aurora IAM tokens, Azure Entra ID tokens, Vault leases — expire long before a connection pool does. Today the only way to use them with Tortoise is to tear down and re-init the client on a timer, which kills in-flight queries when the pool is closed underneath them.

```python
async def get_token() -> str:
    return await mint_short_lived_token()   # cached until shortly before expiry

await Tortoise.init(
    config={
        "connections": {
            "default": {
                "engine": "tortoise.backends.asyncpg",
                "credentials": {
                    "host": "db.host",
                    "port": 5432,
                    "user": "someuser",
                    "password": get_token,
                    "database": "somedb",
                },
            }
        },
        "apps": {...},
    }
)
```

Refs #2034.

### Scope

This deliberately adds the **primitive** rather than native IAM support. #2034 asks for an `iam_auth` flag with boto3 token generation built in; that would put a cloud-vendor SDK inside Tortoise and only solve one provider. A callable password is what `asyncpg` and SQLAlchemy both expose, and it covers AWS IAM, Azure Entra ID, GCP, Vault and anything else in a few lines of user code.

PostgreSQL only. MySQL has the same `create_pool(password=...)` shape, but neither `aiomysql` nor `asyncmy` resolves callables, so supporting it there needs its own design — happy to follow up if you'd like it.

## Motivation and Context

We run Tortoise against Azure Postgres with Entra ID (passwordless) auth, where tokens live ~60 minutes. Our workaround was a background task calling `Tortoise.close_connections()` + re-init every 5 minutes. On the asyncpg backend `_close()` gives in-flight work a 10-second grace period before `pool.terminate()`, so a long-running batch job gets its connections pulled mid-statement and fails with `InterfaceError: pool is closing`. Resolving the password per connection removes the need to recycle the pool at all.

## How Has This Been Tested?

- New `tests/backends/test_password_factory.py`: `resolve_password` over string/`None`/sync/async callables, asyncpg forwarding the callable untouched, psycopg keeping the password out of the conninfo, the generated psycopg connection class minting a fresh password per connect, and `star_password` not choking on a callable.
- `make check` — ruff format + ruff check + mypy (134 files) + bandit all clean.
- Full sqlite suite: 1914 passed, 148 skipped, 2 xfailed.
- Verified end to end against a live Azure Postgres instance with a rotating Entra ID token on the asyncpg backend.

## Implementation notes

- **asyncpg** already resolves callable passwords itself (`connect_utils.py`, awaits the result if awaitable), and `BasePostgresClient` forwarded `password` untouched — so this worked by accident but was typed `str | None` and undocumented. It is now typed, tested and documented so it doesn't get "fixed" away.
- **psycopg** bakes credentials into an immutable conninfo string, so the password is instead injected at connect time through a generated connection class. `psycopg_pool` does support a callable `conninfo`, but only from 3.3, while `pyproject.toml` allows `psycopg[pool] >=3.0.12` — this keeps that floor intact.
- **`Tortoise.star_password()`** was a latent crash: it does `password[0 : len(password) // 3]` and `str.replace(password, ...)`, both `TypeError` on a non-string. Now skips non-strings.

## Checklist:

- [x] I have added an entry to the CHANGELOG.


waketzheng:
@get-bowen Nice job! I merged `upstream/develop` and fixed the lint issues. Before the new version is released, you can use this feature from the develop branch:
```
uv pip install -U "tortoise-orm @git+https://github.com/tortoise/tortoise-orm"
```

waketzheng (approved):

PR #2263, opened 2026-08-13
fix: replace bare except with except Exception in sphinx autodoc typehints script

TrueFurina:
Replace 3 bare `except:` clauses in `docs/sphinx_autodoc_typehints.py` with `except Exception:` — same intended behavior (forward-reference evaluation fallbacks and warning logging), but KeyboardInterrupt/SystemExit are no longer intercepted first. Verified: `python -m py_compile` passes.

### DIFF
--- a/docs/sphinx_autodoc_typehints.py
+++ b/docs/sphinx_autodoc_typehints.py
@@ -53,10 +53,10 @@ def format_annotation(annotation, fully_qualified=False):
                 if 'ForwardRef(' in str(bound):
                     try:
                         bound = bound._evaluate(sys.modules[annotation.__module__].__dict__, None)
-                    except:
+                    except Exception:
                         try:
                             bound = bound._evaluate(type_globals.__dict__, None)
-                        except:
+                        except Exception:
                             bound = bound.__forward_arg__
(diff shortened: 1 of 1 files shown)


waketzheng:
Thanks for this — replacing bare `except:` with `except Exception:` is a good change, and the reasoning in the PR description is correct: `KeyboardInterrupt` and `SystemExit` should propagate rather than be silently swallowed by a fallback handler.

The diff looks correct. Both the nested try/except in `format_annotation` and the fallback in `get_all_type_hints` are intended to catch ordinary evaluation errors, not process-level signals, so `except Exception:` preserves the original behavior while being more explicit.

One minor thought: `except Exception:` is still fairly broad. If you want to be more precise, you could narrow the first one in `format_annotation` to something like `except (NameError, AttributeError, TypeError):`, since that's what `_evaluate` is likely to raise on a forward-reference failure. But I don't consider this blocking — the current form is a clear improvement over the bare `except:` and is fine as-is if you prefer to keep the fallback logic simple.

CI shows no performance impact and `py_compile` passes, so I'm happy to approve this.

waketzheng (approved):

Issue #384, opened 2020-05-06
Change restrictions on Charfield, Textfield


gdmoore:
`CharField` _requires_ a `max_length` but not all databases mandate this. In Postgresql, the max_length can be optional. It would be nice to have support for the same functionality. Discussing it with @grigi brought up the idea of allowing `max_length` to be 0, in which case it will be unbounded *if the database supports it*.

`TextField` doesn't support any indexes currently. I think it should allow b-tree indexing. There are many use cases where b-tree index is useful, such as needing to perform prefix matches (`textfield__startswith="..."`). This is a separate use case from full-text search. 

Wherever there are DB-specific implementation differences for Fields, the documentation should make it clear what restriction(s) apply across databases.

davidpurser:
This is a problem for me as well.  It's unclear to me whether the "index" or "unique" properties of a field actually affect any logic within Tortoise, other than the schema generation.  I've just disabled 'index' in Tortoise but I still have the index created manually in Postgres.

whisust:
I would kind of label it as a bug, as the following is valid PG syntax

```postgresql
CREATE TABLE IF NOT EXISTS roles
(
    id         UUID      NOT NULL PRIMARY KEY DEFAULT uuid_generate_v4(),
    name       TEXT   NOT NULL UNIQUE,
    realm      TEXT,
    created_at TIMESTAMPTZ NOT NULL             DEFAULT NOW(),
    updated_at TIMESTAMPTZ
);

CREATE INDEX roles_realm_idx ON roles (realm);
```

and the matching code 

```python
class Role(Model):
    id = fields.UUIDField(pk=True)
    name = fields.TextField(unique=True, null=False)
    real = fields.TextField(index=True),
    created_at = fields.DatetimeField(auto_now=True, null=False)
    updated_at = fields.DatetimeField()
```

fails on `name` due to the `TextField` init code either on `unique` or `index`
```python
class TextField(Field[str], str):  # type: ignore
    """
    Large Text field.
    """

    indexable = False
    SQL_TYPE = "TEXT"

    def __init__(
        self, pk: bool = False, unique: bool = False, index: bool = False, **kwargs: Any
    ) -> None:
        if pk:
            ...
        if unique:
            # PG TextField DO support unique indexes
            raise ConfigurationError(
                "TextField doesn't support unique indexes, consider CharField or another strategy"
            )
        if index:
            # PG TextField DO support other indexes
[...]

PR #2264, opened 2026-08-17
Allow TextField indexes on PostgreSQL

evstratbg:
## Description

Allow `TextField(unique=True)` and `TextField(db_index=True)` when the configured database dialect is PostgreSQL.

This adds dialect-aware indexability validation and applies it consistently to schema generation and the migration schema editor. Other backends continue to reject indexed `TextField` declarations when generating or migrating schema. `CharField` behavior is unchanged.

## Motivation and Context

PostgreSQL supports B-tree indexes and unique constraints on `TEXT`, but Tortoise currently rejects these declarations before the database dialect is known.

This addresses the `TextField` indexing portion of #384. The separate `CharField` request in that issue is intentionally out of scope.

## How Has This Been Tested?

- Added asyncpg and psycopg schema-generation tests for `TEXT UNIQUE` and `CREATE INDEX ... (text_column)`.
- Added SQLite rejection coverage to verify that the behavior remains PostgreSQL-only.
- Added migration schema editor tests for PostgreSQL support and non-PostgreSQL rejection.
- Ran the full SQLite test suite: `1913 passed, 148 skipped, 2 xfailed`.
- Ran the affected test files: `76 passed`.
- Ran Ruff format/check, Bandit, package build, Twine checks, and the documentation build.

## Checklist:

### DIFF
--- a/tortoise/fields/base.py
+++ b/tortoise/fields/base.py
@@ -244,7 +244,7 @@ class Field(Generic[VALUE], metaclass=_FieldMeta):
                 raise ConfigurationError(
                     f"{self.__class__.__name__} can't set both db_index and index"
                 )
-        if not self.indexable and (unique or db_index):
+        if not self._is_indexable() and (unique or db_index):
             raise ConfigurationError(f"{self.__class__.__name__} can't be indexed")
         if (pk := kwargs.pop("pk", None)) is not None:
             if primary_key is None:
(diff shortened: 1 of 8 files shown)


tortoise/fields/base.py
```diff
@@ -244,7 +244,7 @@ def __init__(
                 raise ConfigurationError(
                     f"{self.__class__.__name__} can't set both db_index and index"
                 )
-        if not self.indexable and (unique or db_index):
+        if not self._is_indexable() and (unique or db_index):
```
waketzheng, line 247: It would be better to write it as:

```python
if (unique or db_index) and not self._is_indexable():
```

tortoise/fields/base.py
```diff
@@ -290,6 +290,21 @@ def __init__(
         self.model: type[Model] = model  # type: ignore
         self.reference: Field | None = None
 
+    def _is_indexable(self, dialect: str | None = None) -> bool:
+        if dialect is not None:
+            return bool(self.get_for_dialect(dialect, "indexable"))
+        if self.indexable:
+            return True
+        return any(
```
waketzheng, line 298: Using `getattr` twice makes the code harder to understand. How about the following?

```python
return any(
    bool(getattr(sub_cls, "indexable", False))
    for name in dir(self.__class__)
    if name.startswith("_db_") and isinstance(sub_cls := getattr(self, name), type)
)
```

waketzheng (changes requested):

PR #2266, opened 2026-08-24
Add db_collation argument for fields

### LINKED ISSUE #686: Allow setting collate on fields
In SQLite, I can make a text field have a collate of `NOCASE`, but I don't know how to do that in Tortoise. I tried searching the docs, but there seems to be no way to make a field have a collate of `NOCASE`, only do case-insensitive matches on `SELECT`, which is not what I want.

Collates to be supported.

ChrisJr404:
## Description

Adds a `db_collation` argument to fields so you can set a column collation without hand writing SQL. When it is set, the generated schema gets a `COLLATE` clause, and it defaults to `None` so nothing changes for existing models.

The example from the issue works now:

```python
name = fields.CharField(max_length=50, db_collation="NOCASE")
```

produces `"name" VARCHAR(50) COLLATE NOCASE NOT NULL` on SQLite. The collation name is passed through as given so you use whatever the target database knows. Postgres wraps it in double quotes since its collation names are identifiers, the other backends emit it as a bare name.

## Motivation and Context

Closes #686. There was no way to ask a text column for a specific collation (like `NOCASE` on SQLite), you had to drop down to raw SQL. This lets you declare it on the field where it belongs.

## How Has This Been Tested?

Added `test_schema_db_collation` that generates the schema for a model using `db_collation` on a `CharField` and a `TextField` and checks the `COLLATE` clause shows up. Ran the sqlite test suite locally and everything passes.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Changelog
 Added
 ^^^^^
 - ``Q.__bool__()`` so ``Q`` objects with no filters/children (including nested empty ``Q`` children) are falsy.
+- ``db_collation`` argument on fields to set a column collation, emitted as a ``COLLATE`` clause in the generated schema. (#686)
 
 1.1.8
 -----
--- a/tests/schema/models_collation.py
+++ b/tests/schema/models_collation.py
@@ -1,0 +1,12 @@
+"""
+This example demonstrates SQL Schema generation for fields that set a db_collation.
+"""
+
+from tortoise import fields
+from tortoise.models import Model
+
+
+class Account(Model):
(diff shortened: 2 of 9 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ Changelog
 Added
 ^^^^^
 - ``Q.__bool__()`` so ``Q`` objects with no filters/children (including nested empty ``Q`` children) are falsy.
+- ``db_collation`` argument on fields to set a column collation, emitted as a ``COLLATE`` clause in the generated schema. (#686)
```
waketzheng, line 17: Use `#2266` instead of `#686`

Issue #1549, opened 2024-01-23
Decimal field represented in scientific notation when using pydantic_model_creator


Yalchin403:
**Describe the bug**
I use pydantic_model_creator for creating schema from model which I use for serialization in the response. I have a decimal field, which is shown in scientific notation in the response instead of being Decimal.

**To Reproduce**
- Create a model containing a decimal field
- Add a schema to with `pydantic_model_creator` 
- Add a view to create a record of this model and return this model using the schema you created
- You will see the decimal field in scientific notation

**Expected behavior**
I normally expect to get the defined value as it is, for example, if I created this record's decimal field as 200.00 i expect it as 200.00 not as 2E+2

**Additional context**

<img width="1010" alt="image" src="https://github.com/tortoise/tortoise-orm/assets/54992849/778f76d1-1715-4b26-b43d-989129f96bd9">

sloppycoder:
i have the same problem...

markus-96:
This is because of how the decimal field is saved in the database. You can use a custom serializer, see the following example:

```py
from decimal import Decimal
from typing import Callable, Coroutine

from pydantic import BaseModel, field_serializer
from tortoise import models, fields, run_async
from tortoise.contrib.pydantic import pydantic_model_creator
from tortoise.contrib.test import init_memory_sqlite


class Item(models.Model):
    dec = fields.DecimalField(max_digits=100, decimal_places=2)


class Item2(BaseModel):
    id: int
    dec: Decimal


def run_in_memory_sqlite(func: Callable[..., Coroutine]) -> None:
    run_async(init_memory_sqlite(func)())


async def main():
    Item_Pydantic = pydantic_model_creator(Item)

    class Item_Pydantic2(Item_Pydantic):  # subclassing to add custom serializer, `json_encoders` is deprecated!
        @field_serializer('dec')  # custom serializer
        def dec_serializer(self, dec: Decimal):
            return "{0:f}".format(dec)  # as far as I know this does not affect precision

    acc = await Item.create(dec='5E+2')  # just for testing, lets input a scientifically notated number.
    print((await Item_Pydantic2.from_tortoise_orm(acc)).model_dump_json())
    # result: {"id":1,"dec":"500"}

    acc2 = Item2(dec="5E+2", id=1)  # normal pydantic will return the scientific notation!
    print(acc2.model_dump_json())
    # result: {"id":1,"dec":"5E+2"}


if __name__ == '__main__':
    run_in_memory_sqlite(main)
```

PR #2271, opened 2026-09-02
fix: DecimalField no longer strips the scale it just quantized to

dylanpulver:
`DecimalField.to_python_value` does `Decimal(value).quantize(self.quant).normalize()`. `quantize`
sets the scale to `decimal_places`; `normalize` on the same expression strips it again. So a
`DecimalField(max_digits=12, decimal_places=2)` holding `100.00` becomes `Decimal('1E+2')`:

```
to_python_value('100.00')   -> Decimal('1E+2')      str -> 1E+2      f'${...}' -> $1E+2
to_python_value('10')       -> Decimal('1E+1')      str -> 1E+1
to_python_value('1000000')  -> Decimal('1E+6')      str -> 1E+6
to_python_value('12.34')    -> Decimal('12.34')
```

`Model.__init__` runs `to_python_value` on user input too, so it is written that way as well — the
SQLite column literally holds `'1E+2'` — and `pydantic_model_creator` then emits
`{"id":1,"price":"1E+2"}`.

**I believe this is the cause of #1549** (open since 2024-01-23, reporter's example: created as
`200.00`, got `2E+2`). The workaround on that thread attributes it to "how the decimal field is
saved in the database"; it reproduces with no database at all, from `to_python_value` alone — the
DB row is `1E+2` *because* tortoise wrote it that way.

**Why the existing tests can't see it:** `tests/fields/test_decimal.py`
asserts through `obj.decimal == Decimal("…")`, and `Decimal.__eq__` compares numerically —
`Decimal('1E+2') == Decimal('100.00')` is `True`. The whole file passes with the defect live *and*
with it fixed. `test_aggregate_sum_with_f_expression` expects `Decimal("4E+1")` for forty, which is
the shape of a value transcribed from a run rather than written by hand. The new tests assert on
`str()` and on `as_tuple().exponent` instead.

**Mutants.** Reverting `tortoise/fields/data.py` against `origin/develop` and keeping the new tests
fails both of them. Dropping the `quantize` as well (the "stop transforming it" reading) fails 14
tests in `test_decimal.py` and 16 suite-wide — the quantize is load-bearing, only the `normalize` is
the defect.

**Run:** under the repo's own `make deps`, `pytest` → `1914 passed, 148 skipped, 2 xfailed, 0
failed` (2064 collected). An earlier version of this description reported `2 failed`; those were
`test_recursive` and `test_version` failing only because the package was not pip-installed, and they
are gone under `make deps`. `ruff format --check`, `ruff check` (0.15.4, the version in `uv.lock`)
and `mypy` are clean on the changed files. Verified end to end on **SQLite and PostgreSQL 17.9**.
MySQL, MSSQL and Oracle are untested here, though `to_python_value` is backend-independent.

Worth noting for Postgres specifically: `NUMERIC(8,2)` re-imposes the scale on read, so a refetch
looks correct on `develop` while `obj.price` straight after `create()` is still `Decimal('1E+4')`.
The same row therefore reports two different scales in one request, and aggregates are wrong on both
backends.

One self-correction worth stating: my first version of the new test expected `12.345` → `12.35`.
`quantize` uses the context rounding, `ROUND_HALF_EVEN`, so it is `12.34`. My expectation was wrong,
not the code; the test now uses `12.3456` so it does not depend on the rounding mode.

AI assistance: found and drafted with Claude Code (Claude Opus 5, `claude-opus-5`), from a sweep for
field round trips that do not close, not from production use.

### DIFF
--- a/tortoise/fields/data.py
+++ b/tortoise/fields/data.py
@@ -382,7 +382,7 @@ class DecimalField(Field[T_DECIMAL], Decimal):  # type: ignore
 
     def to_python_value(self, value: Any) -> Decimal | None:
         if value is not None:
-            value = Decimal(value).quantize(self.quant).normalize()
+            value = Decimal(value).quantize(self.quant)
         return value
 
     @property
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Added
 Fixed
 ^^^^^
 - ``QuerySet.distinct().count()`` no longer counts rows duplicated by a join (e.g. filtering on a m2m relation); it now counts distinct primary keys and matches the number of rows the query returns. (#2255)
+- ``DecimalField`` no longer strips the scale it just applied: ``to_python_value()`` called ``.normalize()`` immediately after ``.quantize()``, so a ``DecimalField(max_digits=12, decimal_places=2)`` holding ``100.00`` came back as ``Decimal('1E+2')`` and rendered as ``1E+2`` in strings, f-strings and JSON. (#2271)
 
 1.1.8
 -----
--- a/tests/fields/test_decimal.py
+++ b/tests/fields/test_decimal.py
@@ -309,3 +309,34 @@ async def test_aggregate_max_with_f_expression(db):
         .values("max_decimal")
     )
     assert values[0] == {"max_decimal": None}
+
+
+@pytest.mark.asyncio
+async def test_scale_is_preserved(db):
+    # Decimal.__eq__ compares numerically, so `== Decimal("27.27")` cannot see
+    # the scale. Assert on the string form, which is what reaches a template,
+    # an f-string or a JSON serializer.
+    obj0 = await testmodels.DecimalFields.create(decimal=Decimal("100.00"), decimal_nodec=10)
+    obj = await testmodels.DecimalFields.get(id=obj0.id)
+    assert str(obj.decimal) == "100.0000"
+    assert str(obj.decimal_nodec) == "10"
+    assert obj.decimal.as_tuple().exponent == -4
+    assert obj.decimal_nodec.as_tuple().exponent == 0
+
+
+def test_to_python_value_scale():
+    field = fields.DecimalField(max_digits=12, decimal_places=2)
+    for value, expected in (
+        ("100.00", "100.00"),
+        ("100", "100.00"),
+        ("10", "10.00"),
+        ("1000000", "1000000.00"),
+        ("0", "0.00"),
+        ("1234.5", "1234.50"),
+        ("12.3456", "12.35"),
+    ):
+        assert str(field.to_python_value(value)) == expected
+
+    field0 = fields.DecimalField(max_digits=12, decimal_places=0)
+    assert str(field0.to_python_value("10")) == "10"
+    assert str(field0.to_python_value("100")) == "100"


waketzheng:
SQLite handles decimal values correctly:
```


class Device(Model):
    price = fields.DecimalField(8, 2)


async def main() -> None:
    await Tortoise.init(db_url="sqlite://db.sqlite3", modules={"models": ["__main__"]})
    price = 10000
    obj = await Device.create(price=price)
    saved = await Device.get(pk=obj.pk)
    assert saved.price == price
    print(dict(saved))
    # {'price': Decimal('1E+4'), 'id': 2}


if __name__ == "__main__":
    run_async(main())
```
Therefore, instead of modifying `tortoise/fields/data.py`, modify `tortoise/contrib/pydantic/creator.py` instead.
```diff
--- a/tortoise/contrib/pydantic/creator.py
+++ b/tortoise/contrib/pydantic/creator.py
@@ -5,11 +5,12 @@ import inspect
 from base64 import b32encode
 from collections.abc import Iterator, MutableMapping
 from copy import copy
+from decimal import Decimal
 from enum import Enum, IntEnum
 from hashlib import sha3_224
-from typing import TYPE_CHECKING, Any, TypeAlias, cast
+from typing import TYPE_CHECKING, Annotated, Any, TypeAlias, cast

-from pydantic import ConfigDict, computed_field, create_model
+from pydantic import ConfigDict, PlainSerializer, computed_field, create_model
 from pydantic import Field as PydanticField
 from pydantic.fields import ComputedFieldInfo

@@ -28,10 +29,12 @@ from tortoise.contrib.pydantic.descriptions import (
 )
 from tortoise.contrib.pydantic.utils import get_annotations
 from tortoise.exceptions import NoValuesFetched
-from tortoise.fields import Field, JSONField
+from tortoise.fields import DecimalField, Field, JSONField
 from tortoise.fields.data import CharEnumFieldInstance, IntEnumFieldInstance

 if TYPE_CHECKING:  # pragma: nocoverage
+    from typing_extensions import TypeForm
+
     from tortoise.models import Model

 # Type alias for a single entry in the recursion stack: (model_class, field_name, max_recursion)
@@ -486,12 +489,14 @@ class PydanticModelCreator:
             json_schema_extra["readOnly"] = constraints["readOnly"]
             del constraints["readOnly"]
         fconfig.update(constraints)
-        python_type: type[Enum] | type[IntEnum] | type
+        python_type: type[Enum] | type[IntEnum] | type | TypeForm[Decimal]
         if isinstance(field, (IntEnumFieldInstance, CharEnumFieldInstance)):
             python_type = field.enum_type
+        elif isinstance(field, DecimalField):
+            python_type = Annotated[Decimal, PlainSerializer(lambda x: f"{x:f}", return_type=str)]
         else:
             python_type = getattr(field, "related_model", field.field_type)
-        ptype = python_type
+        ptype: Any = python_type
```

tortoise/fields/data.py
```diff
@@ -382,7 +382,7 @@ def __init__(self, max_digits: int, decimal_places: int, **kwargs: Any) -> None:
 
     def to_python_value(self, value: Any) -> Decimal | None:
         if value is not None:
-            value = Decimal(value).quantize(self.quant).normalize()
+            value = Decimal(value).quantize(self.quant)
```
waketzheng, line 385: `normalize` is useful — it changes `Decimal('9.100000')` to `Decimal('9.1')`, which is much more human‑readable. Please roll back the change and modify the Pydantic creator instead.

waketzheng (changes requested):

waketzheng (approved):

PR #2276, opened 2026-09-09
fix(models): apply defaults after concurrent inserts

### LINKED ISSUE #2060: `update_or_create` loses defaults under concurrent access
## Describe the bug

`update_or_create` has a race condition when creating new rows. Two concurrent calls on the same non-existent key both see "no row exists" and both proceed to INSERT. The losing caller's INSERT fails with `IntegrityError`, which is caught and retried via GET, but GET just fetches the row without applying `defaults`. The caller neither created nor updated, despite the method name.

This happens because the transaction ends before `_create_or_get` is called. Both transactions complete their `SELECT FOR UPDATE` (finding nothing), then race on the subsequent INSERT.

```python
# Current implementation, trimmed for clarity
# https://github.com/tortoise/tortoise-orm/blob/a7259172a034208884f077a8ffcc5d49d599f3d9/tortoise/models.py#L1181
async with in_transaction(connection_name=db.connection_name) as connection:
    instance = await cls.select_for_update().using_db(connection).get_or_none(**kwargs)
    if instance:
        await instance.update_from_dict(defaults).save(using_db=connection)
        return instance, False
return await cls._create_or_get(db, defaults, **kwargs)
```

## To Reproduce

This is what would trigger the race in a real application, but the timing window is narrow and hard to hit in practice:

```python
import asyncio
from tortoise import Tortoise, fields

class Item(Model):
    name = fields.CharField(max_length=100, unique=True)
    value = fields.CharField(max_length=100)

async def main():
    await Tortoise.init(db_url="postgres://...", modules={"models": [__name__]})

    # Concurrent update_or_create on same key
    results = await asyncio.gather(
        Item.update_or_create(name="test", defaults={"value": "a"}),
        Item.update_or_create(name="test", defaults={"value": "b"}),
        return_exceptions=True
    )
    print(results)

asyncio.run(main())
```

The reproducer below uses the same logic as `update_or_create` but adds an `asyncio.Barrier` to force both transactions to overlap, making the race 100% reproducible:

<details>
<summary>100% reproducible test case using asyncio.Barrier</summary>

```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.11"
# dependencies = ["tortoise-orm", "asyncpg"]
# ///
"""
Reproducer for update_or_create race condition in Tortoise ORM.

Requires PostgreSQL:
    docker run --rm -d --name pg -e POSTGRES_PASSWORD=test -p 5432:5432 postgres:16

Run: chmod +x reproducer.py && ./reproducer.py
"""
import asyncio
from tortoise import Tortoise, fields
from tortoise.transactions import in_transaction
from tortoise.exceptions import IntegrityError, DoesNotExist

DB_URL = "postgres://postgres:test@localhost:5432/postgres"
barrier = asyncio.Barrier(2)


class Item(Model):
    name = fields.CharField(max_length=100, unique=True)
    value = fields.CharField(max_length=100)


async def _create_or_get(defaults, **kwargs):
    """Upstream _create_or_get logic."""
    merged = {**kwargs, **defaults}
    try:
        async with in_transaction() as connection:
            return await Item.create(using_db=connection, **merged), True
    except IntegrityError as exc:
        try:
            return await Item.filter(**kwargs).get(), False
        except DoesNotExist:
            pass
        raise exc


async def update_or_create_with_barrier(defaults, **kwargs):
    """Upstream update_or_create logic with barrier to force race condition."""
    async with in_transaction() as connection:
        instance = await Item.select_for_update().using_db(connection).get_or_none(**kwargs)
        await barrier.wait()  # Both transactions reach here before either continues
        if instance:
            await instance.update_from_dict(defaults).save(using_db=connection)
            return instance, False
    return await _create_or_get(defaults, **kwargs)


async def main():
    await Tortoise.init(db_url=DB_URL, modules={"models": [__name__]})
    await Item.all().delete()

    print("Running two concurrent update_or_create(name='test')...\n")

    results = await asyncio.gather(
        update_or_create_with_barrier({"value": "A"}, name="test"),
        update_or_create_with_barrier({"value": "B"}, name="test"),
        return_exceptions=True,
    )

    for i, r in enumerate(results, 1):
        if isinstance(r, Exception):
            print(f"Call {i}: {type(r).__name__}")
        else:
            print(f"Call {i}: created={r[1]}, value='{r[0].value}'")

if __name__ == "__main__":
    asyncio.run(main())
```

</details>

Output:

```
Running two concurrent update_or_create(name='test')...

Call 1: created=False, value='B'
Call 2: created=True, value='B'
```

Both calls returned the same value. One caller's `defaults` were silently ignored. If the losing caller had created the row, the values would differ. Instead, `created=False` confirms it just fetched the winner's row without applying its own `defaults`.

## Expected behavior

- Both callers should either create or update.
- The losing caller's `defaults` should be applied via `UPDATE`, not discarded.
- The returned value should always reflect the provided `defaults` for a call.

## Additional context

The retry in `_create_or_get` masks the race by catching `IntegrityError` and falling back to GET, but this means:
- The losing caller's `defaults` are silently dropped
- Extra round-trips (failed INSERT + GET)

I haven't found another issue which perfectly describes this, but it seems tangentially related to #1530.

## Suggested fix

Use `INSERT ... ON CONFLICT DO UPDATE` for true atomic upsert, similar to how `bulk_create` already handles its `on_conflict`/`update_fields` parameters. This would require lookup fields to have a unique constraint, which is arguably correct semantics here to avoid ambiguity anyway.

fzlzjerry:
## Description

Retry the existing transactional update path when another caller wins the initial insert. The losing `update_or_create()` call now applies its own defaults to a freshly read row rather than returning the other caller's values unchanged.

The creation transaction boundary and `get_or_create()` behavior stay unchanged. An explicit `None` check also keeps falsy model instances on the update path.

## Motivation and Context

Closes #2060.

Re-reading the row matters: the unlocked conflict lookup may already be stale or its row may have been deleted. The retry uses the normal update path and its existing backend locking behavior, without adding backend-specific upsert SQL or changing the public API.

## How Has This Been Tested?

- The [unchanged upstream CI workflow on the fork](https://github.com/fzlzjerry/tortoise-orm/actions/runs/34341676313) passed **all 23 jobs** on `45676b9`: Python 3.10–3.14 across SQLite, PostgreSQL (asyncpg/psycopg), MySQL (InnoDB/MyISAM and asyncmy), and SQL Server, plus checks and coverage.
- Nine regression/control cases cover concurrent defaults, disjoint fields, a stale conflict lookup, a deleted winner, caller transaction rollback, falsy instances, unrelated integrity errors, and unchanged `get_or_create()` semantics. Final fixtures yield six failures and three passing controls against the original implementation.
- The race tests call the real creation helper and synchronize before its INSERT; no success/failure result is fabricated. Their unique name is non-nullable so SQL Server enforces the constraint as well.
- Independent runs without method instrumentation passed 78 concurrent API calls on each of SQLite, PostgreSQL, and MySQL.

## Checklist:

- [ ] My change requires a change to the documentation. N/A — behavior correction, no API change.
- [ ] I have updated the documentation accordingly. N/A.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,10 @@ Added
 ^^^^^
 - ``Q.__bool__()`` so ``Q`` objects with no filters/children (including nested empty ``Q`` children) are falsy.
 
+Fixed
+^^^^^
+- ``Model.update_or_create()`` now applies the caller's defaults after losing a concurrent insert, re-reading the current row through the existing transactional update path. (#2060)
+
 1.1.8
 -----
 
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -3,7 +3,7 @@ import sys
 
 import pytest
 
-from tests.testmodels import Tournament, UniqueName
+from tests.testmodels import Tournament, UniqueName, UniqueNameRequired
 from tortoise import connections
 from tortoise.contrib.test import requireCapability
 from tortoise.contrib.test.condition import NotEQ
@@ -54,7 +54,7 @@ async def test_concurrent_create_applies_defaults(
     db_isolated, monkeypatch, method_name, defaults_pair
 ):
     """Both callers read a missing row before either attempts the real INSERT."""
-    original_create_or_get = UniqueName._create_or_get
+    original_create_or_get = UniqueNameRequired._create_or_get
     both_missing = asyncio.Event()
     create_attempts = 0
 
@@ -66,8 +66,10 @@ async def test_concurrent_create_applies_defaults(
         await both_missing.wait()
         return await original_create_or_get(db, defaults, **kwargs)
 
-    monkeypatch.setattr(UniqueName, "_create_or_get", classmethod(synchronized_create_or_get))
-    method = getattr(UniqueName, method_name)
+    monkeypatch.setattr(
+        UniqueNameRequired, "_create_or_get", classmethod(synchronized_create_or_get)
+    )
+    method = getattr(UniqueNameRequired, method_name)
     tasks = [
         asyncio.create_task(method(name="race", defaults=defaults)) for defaults in defaults_pair
     ]
@@ -81,8 +83,8 @@ async def test_concurrent_create_applies_defaults(
 
     assert create_attempts == 2
     assert sum(created for _, created in results) == 1
-    assert await UniqueName.filter(name="race").count() == 1
-    stored = await UniqueName.get(name="race")
+    assert await UniqueNameRequired.filter(name="race").count() == 1
+    stored = await UniqueNameRequired.get(name="race")
     if method_name == "update_or_create":
         for defaults, (instance, _) in zip(defaults_pair, results):
             for field, value in defaults.items():
@@ -104,7 +106,7 @@ async def test_concurrent_create_applies_defaults(
 @pytest.mark.asyncio
 async def test_update_or_create_refreshes_after_create_race(db_isolated, monkeypatch):
     """The unlocked conflict lookup must not overwrite another writer's later changes."""
-    original_create_or_get = UniqueName._create_or_get
+    original_create_or_get = UniqueNameRequired._create_or_get
 
     async def racing_create_or_get(cls, db, defaults, **kwargs):
         await cls.create(name=kwargs["name"], optional="winner", other_optional="old")
@@ -113,14 +115,14 @@ async def test_update_or_create_refreshes_after_create_race(db_isolated, monkeyp
         await cls.filter(pk=instance.pk).update(other_optional="concurrent")
         return instance, created
 
-    monkeypatch.setattr(UniqueName, "_create_or_get", classmethod(racing_create_or_get))
-    instance, created = await UniqueName.update_or_create(
+    monkeypatch.setattr(UniqueNameRequired, "_create_or_get", classmethod(racing_create_or_get))
+    instance, created = await UniqueNameRequired.update_or_create(
         name="race", defaults={"optional": "loser"}
     )
     assert created is False
     assert instance.optional == "loser"
     assert instance.other_optional == "concurrent"
-    stored = await UniqueName.get(pk=instance.pk)
+    stored = await UniqueNameRequired.get(pk=instance.pk)
     assert stored.optional == "loser"
     assert stored.other_optional == "concurrent"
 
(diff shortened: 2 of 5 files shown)


waketzheng:
Thanks for the update — the fix logic is sound and the test coverage remains excellent. I have one design concern about the retry loop that I'd like to discuss before merging.

**Concern: unbounded `while True` retry loop**

The new implementation wraps the entire "locked read → update / create" flow in `while True` (line 210). The logic is correct: if `_create_or_get()` returns `created=False` (meaning a concurrent caller won the insert race), the loop restarts and retries the update path with a fresh read.

However, `while True` has no retry limit and no backoff. In theory, if `_create_or_get()` keeps returning `created=False` — e.g., because concurrent callers keep winning the insert race before this caller can enter the update path — the loop could spin indefinitely. In practice this is extremely unlikely, but an unbounded loop in a core ORM method is still a hazard worth addressing.

**Suggested alternatives:**

1. **Bounded retry with backoff:** Replace `while True` with `for attempt in range(max_retries)` and add a small backoff (e.g., `asyncio.sleep(backoff * (2 ** attempt))`). Raise a clear exception after exhausting retries.
2. **Single retry is probably enough:** The conflict-recovery path itself is already a "retry." In most cases, one retry should be sufficient to enter the update path. Consider capping at 1–2 attempts instead of an unbounded loop.
3. **If you believe infinite looping is mathematically impossible,** please add a comment explaining why — e.g., "each iteration either updates, creates, or raises, so the loop cannot spin." Right now there's no such justification, and the loop looks like it could run forever.

Everything else looks good:
- `instance is not None` correctly handles falsy model instances.
- The `UniqueNameRequired` model for SQL Server is a nice detail.
- CI is green across all backends and CodSpeed shows no performance impact.

I'm happy to approve once the retry loop is bounded (or justified with a comment). If you'd rather keep `while True`, at minimum add a retry counter and a log line so we can diagnose abnormal spins in production.

fzlzjerry:
@waketzheng Addressed in `cd99f08`. Recovery is bounded to two creation attempts, with a fresh transactional read after every INSERT conflict, including the last one. A surviving row is updated from that fresh read; only if the final read also finds no row do we raise the existing `OperationalError`.

This preserves the existing case where a conflict winner is deleted and the next creation succeeds. It also handles a second INSERT conflict whose winner survives, instead of exhausting before that row can be updated. Ordinary conflicts add no sleep, logging, or configurable retry API; get_or_create is unchanged. The method docstring and changelog describe exhaustion.

New deterministic regressions cover the final surviving winner and repeated deletion with caller-transaction rollback. On the old implementation, the latter reached the 30-second watchdog rather than terminating; the fixed implementation raises OperationalError and rolls back both rows. This bounds recovery attempts, not database lock-wait duration.

Validation on macOS arm64 / Python 3.13.5:
- Full SQLite suite: 1,941 passed, 148 skipped, 2 expected failures.
- Focused concurrency/model/transaction suites: 132 passed.
- Standalone public-API smoke: 64 concurrent update_or_create calls applied their respective defaults to one shared row.
- Ruff 0.15.4 format/check passed.

I have not rerun PostgreSQL, MySQL, or SQL Server here; their backend CI remains separate.

waketzheng (approved):

PR #2280, opened 2026-09-12
Clear removed model options in AlterModelOptions

### LINKED ISSUE #2279: makemigrations keeps regenerating the same migration after a model docstring is removed
**Describe the bug**

If I delete a model's docstring, makemigrations writes a migration for it, and then writes the same migration again every time I run it after that. Each one applies without error, but the next run creates another one just like it. I can never get back to `No changes detected`.

I have several models like this, so now every migration I generate comes with a block of extra operations that have nothing to do with what I changed. The only way to stop it is to put the docstrings back.

**To Reproduce**

```python
class Widget(Model):
    """A widget."""

    name = fields.CharField(max_length=50)

    class Meta:
        table = "widget"
```

Run makemigrations and migrate. `0001` records `'table_description': 'A widget.'`.

Now delete the docstring, change nothing else, and run makemigrations and migrate three times. Each run writes a new migration with the same operation in it:

```python
ops.AlterModelOptions(options={'table': 'widget', 'app': 'models', 'pk_attr': 'id'})
```

**Expected behavior**

The first run writes one AlterModelOptions. After that's applied, the second run should say `No changes detected`.

Reproduced on 1.1.7 and 1.1.8, Python 3.13, SQLite and PostgreSQL 16.

**What I think is going on**

`StateModelDiff.generate_operations` compares options with `table`, `app`, `indexes`, `unique_together` and `constraints` filtered out, so it does look at `table_description` and notices it's gone. But `AlterModelOptions.state_forward` applies the change with `model_state.options.update(self.options)`. `dict.update` can't delete a key that isn't in the new dict, so the old description stays in the state and the same diff shows up on the next run. Django's version of this operation keeps an `ALTER_OPTION_KEYS` list and pops the keys that are missing.

One other thing that's probably the same root cause: `database_forward` returns `None`, and table comments only get written by `CreateModel`. So on PostgreSQL, editing or deleting a docstring never updates `COMMENT ON TABLE` either.

edu2004eu:
## Description

`AlterModelOptions.state_forward` applied its options with `model_state.options.update(...)`, which can only add or change keys. When an option is removed from a model the generator emits an `AlterModelOptions` whose options simply don't contain it, so the old value stayed in the state and the same operation was emitted again on the next run.

This clears the keys that belong to this operation and aren't in the new options. The five keys carried by other operations (`table`, `app`, `indexes`, `unique_together`, `constraints`) are left alone — that list already existed as a literal in `_model_options_for_compare`, so I moved it to `AlterModelOptions.UNMANAGED_OPTION_KEYS` and had the diff use it. The bug was the two sides disagreeing about which keys this operation owns, so it seemed worth having one definition.

## Motivation and Context

Fixes #2279.

Deleting a model's docstring clears `table_description`, and from then on makemigrations writes the same migration on every run and never gets back to `No changes detected`. In a project with a few such models, every later migration carries a block of no-op operations unrelated to the change being made.

In practice `table_description` is the only key this affects today, since `pk_attr` is always present.

Not covered here: `database_forward` returns `None` and table comments are only written by `CreateModel`, so on PostgreSQL a changed or deleted docstring still never reaches `COMMENT ON TABLE`. Same root cause, but it needs SQL emission per backend, so I left it out of this PR.

## How Has This Been Tested?

Three tests, in the files already covering these units:

- `test_alter_options_removes_dropped_key` — a removed key is gone from the state after the operation applies.
- `test_alter_options_keeps_unmanaged_keys` — `unique_together` and `table` survive, guarding against the clearing being too broad.
- `test_removing_table_description_settles_after_one_migration` — generates the operation, applies it, and asserts the next diff is empty. This is the reported symptom.

The first and third fail on `develop` and pass with the change.

Also reproduced end to end before and after: a one-model SQLite project, docstring removed, `makemigrations` + `migrate` three times. Before, three identical migrations; after, one, then `No changes detected`.

`make check` passes. `pytest tests/migrations/` is 302 passed, 6 skipped; the full SQLite suite is 1917 passed, 148 skipped, 2 xfailed. One unrelated failure locally, `tests/fields/test_time.py::test_zoneinfo`, is a missing tzdata on my machine and fails the same way without this change.

## Checklist:

### DIFF
--- a/tortoise/migrations/operations.py
+++ b/tortoise/migrations/operations.py
@@ -363,6 +363,9 @@ class DeleteModel(TortoiseOperation):
 
 
 class AlterModelOptions(TortoiseOperation):
+    # Options carried by other operations; the rest belong to this one.
+    UNMANAGED_OPTION_KEYS = frozenset({"table", "app", "indexes", "unique_together", "constraints"})
+
     def __init__(self, name: str, options: dict[str, Any]):
         self.name = name
         self.options = options
@@ -374,6 +377,11 @@ class AlterModelOptions(TortoiseOperation):
         model_state = self.get_model_state(state, app_label, self.name)
 
         model_state.options.update(self.options)
+        # update() only adds or changes keys, so an option removed from the model would
+        # otherwise survive here and be re-detected on every later makemigrations run.
+        for key in list(model_state.options):
+            if key not in self.UNMANAGED_OPTION_KEYS and key not in self.options:
+                del model_state.options[key]
         state.reload_model(app_label, self.name)
 
     async def database_forward(
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Fixed
 - ``QuerySet.distinct().count()`` no longer counts rows duplicated by a join (e.g. filtering on a m2m relation); it now counts distinct primary keys and matches the number of rows the query returns. (#2255)
 - BlackSheep: ``register_tortoise`` now enables the global connection fallback, so database access works when BlackSheep runs handlers in tasks other than the one that initialized the ORM. (#2248)
 - Fix naive DatetimeField timezone drift. (#2277)
+- ``AlterModelOptions`` now clears options that were removed from the model, instead of only merging in the ones that remain; removing a model docstring no longer makes ``makemigrations`` regenerate the same migration on every run. (#2279)
 
 1.1.8
 -----
--- a/tests/migrations/test_operations_state.py
+++ b/tests/migrations/test_operations_state.py
@@ -341,6 +341,36 @@ def test_alter_options(state_with_model: State):
     assert model_state.options["ordering"] == ["-id"]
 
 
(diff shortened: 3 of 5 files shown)


tortoise/migrations/operations.py
```diff
@@ -374,6 +377,11 @@ def state_forward(self, app_label: str, state: State) -> None:
         model_state = self.get_model_state(state, app_label, self.name)
 
         model_state.options.update(self.options)
+        # update() only adds or changes keys, so an option removed from the model would
+        # otherwise survive here and be re-detected on every later makemigrations run.
+        for key in list(model_state.options):
```
waketzheng: Can be two lines:
```
for key in (model_state.options.keys() - self.UNMANAGED_OPTION_KEYS - self.options.keys()):
    del model_state.options[key]
```

edi-monefy: Resolved in 11c4ab2fe408a9147b4dd521b7894065f0f5fc50

waketzheng (approved):

